PR #56, opened 2020-06-08
[newrelic-infrastructure] Changed nri-rabbitmq to nri-redis to avoid confusion

kidk:
- Typo in values.yaml


douglascamata:
@kidk you have to bump up the chart version in `Chart.yaml`, because the chart is being changed. A patch version bump should be enough. 👍

kidk:
Thanks @douglascamata 

I think I did that now 😄

PR #58, opened 2020-06-09
[nri-prometheus] Add support for the New Relic staging environment

erabug:
#### Is this a new chart?

Nope!

This extends what @alejandrodnm added in #53 and adds NR staging environment support to the prometheus exporter.

#### Checklist
- [x] Chart Version bumped


erabug:
I had the wrong email associated with the commit, hence the force-push. It should be correct now.

charts/nri-prometheus/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: 1.3.0
+appVersion: 1.4.0
```
alejandrodnm, line 2: You don't need to change this one, the appVersion is related to the version of the application in this case is the same as the docker image tag so `1.3.0` we only have to update this when updating to a new version of the image.

erabug: Ah, right, good catch. Fixed.

PR #85, opened 2020-07-07
[nri-bundle] Bump subcharts versions

AdrienKuhn:
#### Is this a new chart

No

Upgrade subcharts versions
* nri-prometheus
* nri-metadata-injection
* nri-kube-events
* newrelic-logging

N/A

#### Checklist
- [X] Chart Version bumped


AdrienKuhn:
Subcharts tarballs are supposed to be ignored according to https://github.com/newrelic/helm-charts/blob/master/.gitignore#L36

Do you want me to clear the `charts` subdirectory ?

douglascamata:
@AdrienKuhn it would be great, is that still works 😅 

Thanks again.

AdrienKuhn:
@douglascamata Actually it doesn't seems to work as we are referencing the local charts in the [requirements.yaml file](https://github.com/newrelic/helm-charts/blob/master/charts/nri-bundle/requirements.yaml#L3)

douglascamata:
@AdrienKuhn ah yes. Feel free to update to the remote reference instead then. I think it will be even better like so.

AdrienKuhn:
@douglascamata 

I'm not able to make it work...
I will try to investigate in the next days and I will open a new PR.

PR #123, opened 2020-09-15
[nri-metadata-injection] Changes to support running newrelic metadata injector in Kubernetes clusters with restrictive Pod Security Policy.

### LINKED ISSUE #122: [nri-metadata-injection] Pod and Job can't run in restrictive environment

When running in a cluster with restrictive pod security policies (read only rootfs set to true, RunAsUser set to MustRunAsNonRoot) the metadata injector and it's associated job fail to run.

Update the nri-metadata-injection chart so Pod Security Context is configurable and add a volume/volume mount config for the k8s-webhook-cert-manager pod so that tmp can be mounted as an emptyDir volume.

You could add a Pod Security Policy to the chart, but as the pod doesn't need to run as root (as far as I can tell) and the readonly root issues can be resolved with a emptyDir mount there is no reason to do this.

Alternative changing the related container build for the injector to set a user. The certificate job script could be changed to use /dev/shm as the folder for creating the certs to get around the read only filesystem. Didn't really like this idea.

avenging:
The changes to deployment.yaml allow you to configure a Pod Security Context for the main metadata injector pod so that it doesn't run as root and get denied scheduling
The changes to the job.yaml allow you to configure mounts in the Pod so that /tmp can be mounted as an emptydir when policy enforces a read-only root filesystem.
Additional Values added to values.yaml
README.md updated with new configuration values
Chart version bumped.

Fixes #122

#### Is this a new chart
No

So that the metadata injector can run in more restrictive environments

  - fixes #122

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/nri-metadata-injection/Chart.yaml
+++ b/charts/nri-metadata-injection/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic metadata injection webhook.
 name: nri-metadata-injection
-version: 1.1.0
+version: 1.1.1
 appVersion: 1.3.0
 home: https://hub.docker.com/r/newrelic/k8s-metadata-injection
 source:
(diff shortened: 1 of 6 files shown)


ardias:
Thank you @avenging for the PR

ardias (approved):

PR #126, opened 2020-09-15
[nri-kube-events] Configure entire Pod to run as user 1000 (nri-kube-events)

### LINKED ISSUE #125: [nri-kube-events] Ensure pods start when Pod Security Policies enforce MustRunAsNonRoot

Unfortunately when you try and deploy this chart on a Kubernetes cluster where the default Pod Security Policy enforces MustRunAsNonRoot the nri-kube-events pod fails to start.

Interestingly the chart configures for the newrelic infra agent via a SecurityContext so that it runs as user 1000, but then goes on to set `runAsNonRoot: false` which then fall foul of the Pod Security Policy's MustRunAsNonRoot configuration.

Once this is changed to true the infra agent starts but the events container still fails as it has named user that Kubernetes can't work out if the user is root or not as it doesn't have a numeric id.

Really both these containers should probably work with a MustRunAsNonRoot configuration.

The deployment of the nri-kube-events should be configured (if it can) to run as non-root for both containers. So fixing `runAsNonRoot: false` so it is `runAsNonRoot: true` for the infra agent and setting a security context for the events container would likely fix these issues.

None

avenging:
#### Is this a new chart

No

The nri-kube-events pod can't run on a cluster with a restrictive Pod Security Policy. This seems to be due to some omissions/errors in deployment.yaml in the chart. 

This PR configures the entire Pod to run as user 1000, changes `runAsNonRoot: false` to `runAsNonRoot: true` which must've been an accident and makes the security contexts of both containers equivalent so that it can run with a restrictive Pod Security Policy.

  - fixes #125

#### Checklist
- [x] Chart Version bumped


ardias:
Thank you @avenging for the PR. Looks good. Can you please rebase to the latest master? thank you again

paologallinaharbur (approved):
LGTM

@avenging
Thanks for the contribution, I checked the behaviour in staging and in production and everything looks good.

I will proceed with the merge, then once it is release we will be able to include the new chart in nri-bundle

PR #129, opened 2020-09-18
[nri-metadata-injection] Allow support to modify the webhook timeout value

jeenx:
#### Is this a new chart
No
Webhooks add to API request latency, they should evaluate as quickly as possible.
Admission webhooks created using `admissionregistration.k8s.io/v1beta1` default timeouts to `30` seconds. This can cause potential problems with the scheduler and block new pods from being created. This is fixed to default to `10` seconds in `admissionregistration.k8s.io/v1`. However, it would be nice to have it configurable to allow for a shorter time or longer if necessary. 
  - fixes #

Reference: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts
#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/nri-metadata-injection/Chart.yaml
+++ b/charts/nri-metadata-injection/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic metadata injection webhook.
 name: nri-metadata-injection
-version: 1.1.0
+version: 1.1.1
 appVersion: 1.3.0
 home: https://hub.docker.com/r/newrelic/k8s-metadata-injection
 source:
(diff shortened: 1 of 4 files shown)


ardias:
Hi @jeenx , thank you for this PR. We are aware that in some cases this causes issues. 
If the timeout is by default 30s, could you then change the timeout value to be the default value, instead of 10? We would prefer not to have the default behaviour changed. And since you made it configurable, you can change the value to suit your needs, correct?

jeenx:
> Hi @jeenx , thank you for this PR. We are aware that in some cases this causes issues.
> If the timeout is by default 30s, could you then change the timeout value to be the default value, instead of 10? We would prefer not to have the default behaviour changed. And since you made it configurable, you can change the value to suit your needs, correct?

Sure thing, I updated it to default to 30 seconds. Just keep in mind the default value will change once apiVersion is updated
From the [documentation](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#:~:text=Note%3A%20Default%20timeout%20for%20a,a%20small%20timeout%20for%20webhooks.):
```
Note: Default timeout for a webhook call is 10 seconds for webhooks registered created using admissionregistration.k8s.io/v1, and 30 seconds for webhooks created using admissionregistration.k8s.io/v1beta1.
```

ardias:
Thanks @jeenx . We are trying to get rid of this chart altogether, but in the meantime, thank you for the PR

ardias (approved):

PR #130, opened 2020-09-22
[newrelic-infrastructure] support serviceaccount annotations

MarcusNoble:
#### Is this a new chart
No

Adds support for users to provide annotations to use on the created service account. We have an operator that lets us inject secrets into pods from external secrets store (AWS SecretsManager) rather than using Kubernetes Secrets. To make use of this we need the pod to have the permission to pull the secrets, in EKS we can do this by assigning an IAM role with the service account via an annotation.

N/a

#### Checklist
- [x] Chart Version bumped


ardias (approved):

ardias:
Hi @MarcusNoble sorry, we've been changing some of the charts and you have a conflict here.

PR #131, opened 2020-09-22
[nri-kube-events] support serviceaccount annotations

MarcusNoble:
#### Is this a new chart
No

Adds support for users to provide annotations to use on the created service account. We have an operator that lets us inject secrets into pods from external secrets store (AWS SecretsManager) rather than using Kubernetes Secrets. To make use of this we need the pod to have the permission to pull the secrets, in EKS we can do this by assigning an IAM role with the service account via an annotation.

N/a

#### Checklist
- [x] Chart Version bumped


charts/nri-kube-events/README.md
```diff
@@ -23,6 +23,7 @@ This chart will deploy the New Relic Events Routers as a Deployment.
...
 | `serviceAccount.create`                                    | If true, a service account would be created and assigned to the deployment                                                                                                                                                                     | true                            |
 | `serviceAccount.name`                                      | The service account to assign to the deployment. If `serviceAccount.create` is true then this name will be used when creating the service account                                                                                              |                                 |
+| `serviceAccount.annotations`                               | The annotations to add to the service account if `serviveAccount.create` is set to true.                                                                                                                                                       |                                 |
```
ardias, line 26: small typo here `servi**v**eAccount.create

ardias (approved):
Just needs a small typo fix

charts/nri-kube-events/README.md
MarcusNoble:
```suggestion
| `serviceAccount.annotations`                               | The annotations to add to the service account if `serviceAccount.create` is set to true.                                                                                                                                                       |                                 |
```

PR #135, opened 2020-09-28
fix: ensure that the cluster name is a bit more distinct

ardias:
#### Is this a new chart
No

Adds the clustername to the display name env variable so that the Infra agents generates a distinct fingerprint when nodes in different cluster have the same (under certain conditions)


ardias:
I still need to reproduce the environment that surfaced this issue

gsanchezgavier (approved):
lgtm

paologallinaharbur (approved):

PR #151, opened 2020-10-09
[nri-prometheus] adias/add worker threads config

ardias:
#### Is this a new chart
No

adds missing config options

#### Checklist
- [x] Chart Version bumped


davidbrota (approved):

charts/nri-prometheus/values.yaml
```diff
@@ -68,6 +68,14 @@ config:
   # scrape_enabled_label: "prometheus.io/scrape"
   # Wether k8s nodes needs to be labelled to be scraped or not. Defaults to false.
   require_scrape_enabled_label_for_nodes: true
+  # Wether the integraton sends metrics directly to New Relic (true) or to the Infrastructure agent (false)
+  # standalone: false
+  # List of emitters to use. Valid values include: 'stdout', 'telemetry',' infra-sdk'
+  # Each emitter outputs data in a different format. Use 'telemetry' if using 'standalone = true', 'infra-sdk' if 'standalone = false'. 'stdout' is for debug purposes
+  # emmiters:
+  #   - "infra-sdk"
```
gsanchezgavier, line 76: is it not pomi running standalone with the telemetry sdk always in k8s?

gsanchezgavier: if so i would remove this comments

gsanchezgavier (approved):
looks good just small question

paologallinaharbur (approved):

charts/nri-prometheus/values.yaml
ardias: I'm not sure. Did you change that to set standalone automatically?

ardias: ok, removed the emitters config. The code sets them based on standalone

PR #155, opened 2020-10-16
[nri-bundle] Showing how to configure dependencies in nri-bundle

### LINKED ISSUE #154: [nri-bundle] We should document how to modify configs in chart dependencies
When deploying nri-bundle chart it is possible to modify values for the dependencies charts:
 - Helm documentation: https://helm.sh/docs/chart_template_guide/subcharts_and_globals

This is possible already but I believe we should add at least a hint with some common configs to show how to do so
Es:
```yaml
#values.yaml of nri-bundle
infrastructure:
  enabled: true
# <removed the rest for brevity>
# configure the k8s integration dependency
newrelic-infrastructure:
#  enableLinux: true
# enableWindows: false
#  verboseLog: false
#  image:
#    repository: newrelic/infrastructure-k8s
#    tag: ""
#    windowsTag: 1.21.0-windows-1809-alpha
#    pullPolicy: IfNotPresent
# configure another dependency
nri-kube-events:
  key: value
```

paologallinaharbur:
Improved documentation and values.yaml in order to show how to pass variables to chart dependencies.

Fix #154 

#### Checklist
- [x] Chart Version bumped 


charts/nri-bundle/values.yaml
```diff
@@ -2,24 +2,71 @@
...
+# enables nri-kube-events
 kubeEvents:
   enabled: false
 
+# enables newrelic-logging
 logging:
   enabled: false
 
+# It is possible to configure dependencies values from this chart.
+# For example you can configure verboseLog variable of the newrelic-infrastructure chart installed ad a dependency as follows.
```
ardias, line 30: small typo:  **ad** -> **as**
also maybe add some emphasis on the variable `verboseLog`  to make it easier to see, like bold characters

ardias (approved):

charts/nri-bundle/README.md
```diff
@@ -21,6 +21,57 @@ This chart bundles multiple New Relic products helm-charts.
 | `kubeEvents.enabled`             | Install the [`nri-kube-events` chart][6] | false |
 | `logging.enabled`                | Install the [`newrelic-logging` chart][7] | false |
 
+## Configure Dependencies
+
+It is possible to configure dependencies values from this chart.  
+For example you can configure verboseLog variable of the newrelic-infrastructure chart installed ad a dependency as follows.
```
davidbrota, line 27:
```suggestion
For example you can configure verboseLog variable of the newrelic-infrastructure chart installed as a dependency as follows.
```

charts/nri-bundle/README.md
```diff
@@ -21,6 +21,57 @@ This chart bundles multiple New Relic products helm-charts.
...
+            # https://github.com/newrelic/nri-discovery-kubernetes
+            exec: /var/db/newrelic-infra/nri-discovery-kubernetes
+            match:
+              label.app: redis
+        integrations:
+          - name: nri-redis
+            env:
+              # using the discovered IP as the hostname address
+              HOSTNAME: ${discovery.ip}
+              PORT: 6379
```
davidbrota, line 71: shouldn't we use ${discovery.port} here?

charts/nri-bundle/values.yaml
```diff
@@ -2,24 +2,71 @@
...
+#             # https://github.com/newrelic/nri-discovery-kubernetes
+#             exec: /var/db/newrelic-infra/nri-discovery-kubernetes
+#             match:
+#               label.app: redis
+#         integrations:
+#           - name: nri-redis
+#             env:
+#               # using the discovered IP as the hostname address
+#               HOSTNAME: ${discovery.ip}
+#               PORT: 6379
```
davidbrota, line 65: same as above: ${discovery.port}?

davidbrota (commented):
I have a question about the port configuration when using auto discovery

charts/nri-bundle/values.yaml
paologallinaharbur: same as above

charts/nri-bundle/README.md
paologallinaharbur: I took the example from here:
 - https://docs.newrelic.com/docs/integrations/kubernetes-integration/link-apps-services/tutorial-monitor-redis-running-kubernetes
That is the same reported here:
 - https://github.com/newrelic/helm-charts/blob/fcd1f52c1e931997bec3ca4e5874760d9c33f0e5/charts/newrelic-infrastructure/values.yaml#L137

I believe that since we are running in k8s and there are services in front of the endpoints same ports can be reused and therefore there is no need to change then. They common case is to have standard ports

davidbrota: ok, makes sense 👍

davidbrota (approved):

PR #173, opened 2020-11-03
[nri-metadata-injection] cert-manager support for webhook

velothump:
#### Is this a new chart
No
Add cert-manager support for webhook. We use cert-manager for all our other validating/mutating webhooks, so it makes sense for newrelic to also support this.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/nri-metadata-injection/Chart.yaml
+++ b/charts/nri-metadata-injection/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic metadata injection webhook.
 name: nri-metadata-injection
-version: 1.1.2
+version: 1.2.0
 appVersion: 1.3.0
 home: https://hub.docker.com/r/newrelic/k8s-metadata-injection
 source:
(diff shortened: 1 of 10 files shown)


mitchellmaler:
Since cert-manager is at a v1 version now would it make sense to use the stable api version in the custom resources?
"cert-manager.io/v1"

velothump:
> Since cert-manager is at a v1 version now would it make sense to use the stable api version in the custom resources?

Updated to v1 now. And rebased to resolve conflicts.

douglascamata:
@velothump can you please split this PR into an update in the `nri-metadata-injection` chart and then later follow up with the update to the `nri-bundle` chart?

velothump:
@douglascamata Thanks for the advice comments, I've updated the PR to reflect that.

paologallinaharbur (approved):

douglascamata (approved):
This looks good to me. I requested someone from the code owners to have a look at it too, as I believe it's a great feature to have. 

Thanks a lot, @velothumb!

PR #186, opened 2020-11-10
[nri-kube-events] - Infra-agent / Kube forwarder connects to New Relic One endpoint over the

lenfree:
#### Is this a new chart: 
No

Infra-agent / Kube forwarder connects to New Relic One endpoint over the
internet. Hence, some of us require proxy to be set for this to work

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/nri-kube-events/Chart.yaml
+++ b/charts/nri-kube-events/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic Kube Events
 name: nri-kube-events
-version: 1.3.3
+version: 1.3.4
 appVersion: 1.3.0
 engine: gotpl
 home: https://docs.newrelic.com/docs/integrations/kubernetes-integration/kubernetes-events/install-kubernetes-events-integration
@@ -11,6 +11,6 @@ keywords:
   - newrelic
   - monitoring
 maintainers:
-- name: douglascamata
-- name: jorik
-- name: bpschmitt
+  - name: douglascamata
+  - name: jorik
+  - name: bpschmitt
--- a/charts/nri-kube-events/values.yaml
+++ b/charts/nri-kube-events/values.yaml
@@ -24,6 +24,10 @@
 
 agentHTTPTimeout: "30s"
 
+# Kube-events container will spin up a infra-agent(event-forwarder) as a sidecar container to forward
+# events to New Relic One endpoint. However, due to security restrictions, some users might require
+# to use an http proxy to route traffic over the internet. If this is the case for you, set this value
+# to your http proxy endpoint.
 proxy: {}
 
 sinks:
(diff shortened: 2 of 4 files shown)


paologallinaharbur:
`nri-kube-events` assumes to send the data through an agent in the same cluster, therefore a proxy should not be necessary, isn't this the case?

Where is located your agent?

lenfree:
> `nri-kube-events` assumes to send the data through an agent in the same cluster (running as a sidecar), therefore a proxy should not be necessary, isn't this the case?
> 
> Where is located your agent?

Hi @paologallinaharbur, sorry I wasn't quite sure on how to configure this since I am new to this. 😄

Firstly, I have a infrastructure agent running as a daemon set (https://github.com/newrelic/helm-charts/tree/master/charts/newrelic-infrastructure) in the cluster.  However, when I installed nri-kube-events using helm chart, according to docs, it  will start an events-forwarder and a kube-events where it will send data to events-forwarder? If I am correct, the events-forwarder is another infra-agent running as a sidecar? Because when this deployment started, events-forwarder container keeps crashing and when I checked the logs, it says something about dial tcp error to New Relic One endpoint so the first thing came into my mind is proxy. Could you advise what's the best way to configure this? 😄

paologallinaharbur:
Hello @lenfree, my bad, I misread the PR and I thought you were adding the proxy to connect to the agent 😅 

And yes you are right it uses an agent running as a sidecar, that agent might need a proxy in order to connect to New Relic

lenfree:
> Hello @lenfree, my bad, I misread the PR and I thought you were adding the proxy to connect to the agent 😅
> 
> And yes you are right it uses an agent running as a sidecar, that agent might need a proxy in order to connect to New Relic

Would you like me to update PR description? 😄

charts/nri-kube-events/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic Kube Events
 name: nri-kube-events
-version: 1.3.2
+version: 1.3.3
```
paologallinaharbur, line 4: In the meanwhile a chart with 1.3.3 was released, can you bump it to 1.3.4?

charts/nri-kube-events/values.yaml
```diff
@@ -24,6 +24,8 @@
 
 agentHTTPTimeout: "30s"
 
+proxy: {}
```
paologallinaharbur, line 27: Can you add some comments commented out?
Only emplaining by which container it is used and why it can be useful

paologallinaharbur (changes requested):

paologallinaharbur:
No need to update the description, the comment it is enough to clarify. 👍 
Can you also update branch resolving the conflics? 

And thanks for the contribution!

charts/nri-kube-events/Chart.yaml
lenfree: sure! let me fix this now

charts/nri-kube-events/values.yaml
lenfree: sure! let me fix this

lenfree:
> No need to update the description, the comment it is enough to clarify. 👍
> Can you also update branch resolving the conflics?
> 
> And thanks for the contribution!

Rebase off master and address PR comments.

Thanks,

paologallinaharbur (approved):

paologallinaharbur:
I merged the code @lenfree, thanks again for the contribution

This version of the chart will be available shortly and it will be included in the next nri-bundle release

PR #211, opened 2020-11-24
[charts/newrelic-logging] Add namespace to all namespaced resources

douglascamata:
#### Is this a new chart
No.

Uses `.Release.Namespace` to explicitly set the namespace on every namespaced resource.

This should allow us to properly deploy this chart to a specific namespace.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-logging/Chart.yaml
+++ b/charts/newrelic-logging/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to deploy New Relic Kubernetes Logging as a DaemonSet
 name: newrelic-logging
-version: 1.4.0
+version: 1.4.1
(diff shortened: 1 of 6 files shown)


arvdias:
@douglascamata does this do the same thing as #185 ?

douglascamata:
@arvdias no. In fact, that PR puts all the Logging chart resources in the `newrelic` namespace. Here I make it respect the namespace that can be sent through the Helm CLI.

In #185 we also have the unification of the global cluster name variable: this chart currently uses `clusterName`, where all the other charts for infra products use `cluster`. There's still value in this and we should contact the author to update the PR there.

douglascamata:
@arvdias let's use that PR as it does other valuable things.

PR #233, opened 2020-12-09
[newrelic-infrastructure] bumping app version

paologallinaharbur:
- The base image of newrelic/infrastructure-k8s has been updated to 2.0.0.
That base image is bundling the integration nri-nginx 3.0.1 that contains a breaking change.
More info regarding all the integration upgraded can be found in the [release notes of the base image](https://github.com/newrelic/infrastructure-bundle/releases/tag/2.0.0).

Since this contains a breaking change for an integration bundled also the chart version has been updated

#### Checklist
- [x] Chart Version bumped


charts/newrelic-infrastructure/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic Infrastructure Agent as a DaemonSet
 name: newrelic-infrastructure
-version: 1.6.5
```
roobre, line 4: Do we need to bump the chart version here? If we're not pulling new deps or breaking the format of the values I'd leave it as it is.

roobre: NVM, seems that version is the importat one to bump.

roobre (approved):
lgtm

charts/newrelic-infrastructure/Chart.yaml
paologallinaharbur: Discussed offline

Issue #210, opened 2020-11-20
newrelic-infrastructure: needs affinity values


cweidinger:
add .Values.affinity to the Daemonset's spec.template.spec.affinity

**Is your feature request related to a problem? Please describe.**

We need to prevent this from getting deployed to fargate on eks and the most sane way is an affinity rule to avoid fargate's lavel

**Describe the solution you'd like**
Add

{{- if .Values.affinity }}
      affinity:
{{ toYaml .Values.affinity | indent 8 }}
{{- end }}
to the Daemonset's template like
https://github.com/helm/charts/blob/90590531b829dbc1142833a5d124cef7cc17b893/stable/prometheus-node-exporter/templates/daemonset.yaml#L80

**Describe alternatives you've considered**

We considered adding a standard label to all our nodes which wouldn't be applied to the fargate instance which we could use with the newrelic-infrastructure's nodeSelector but that would require a redeploy of all of our clusters and doesn't make much sense to have a node label that applies to all nodes (except fargate)

freehck-codefresh:
We need this functionality too. We use newrelic to monitor only specific pods, and we do know their location. It's kind of overhead to deploy the daemonset on all the nodes.

DragonStuff:
We at [TableCheck Japan](https://tablecheck.com/) use the newrelic-infrastructure chart extensively, and want to expand our usage to our mixed AWS EKS Managed Node Groups & Fargate cluster.

Please let me know if you want me to open a new MR and sign the CLA (as I see this is unsigned @freehck-codefresh).

roobre:
H there!
Thanks for raising your concerns. Jut wanted to let you know that we're working on EKS fargate support now. As a part of this effort, we're also adding the ability to add custom `nodeAffinity` items to the daemonSet :)

roobre:
This feature was implemented in #348. Definiing `global.fargate` will now automatically add affinity rules to exclude the DaemonSet from fargate nodes.

in the future, this label will also enable other features for instrumenting EKS fargate, in which we're still working.

Additionally, this PR also allwos to manually define node affinity rules. However please keep in mind that if these rules are manually overriden, `global.fargate` will no longer add the fargate exclusion automatically.

DragonStuff:
@roobre Thank you so much, we’re going to use this immediately. Keep up the great work and we 💜 New Relic.

PR #249, opened 2020-12-28
Affinity values

freehck-codefresh:
This PR adds affinity configuration for #210

### DIFF
--- a/charts/newrelic-infrastructure/templates/daemonset.yaml
+++ b/charts/newrelic-infrastructure/templates/daemonset.yaml
@@ -253,10 +253,11 @@ spec:
       {{- if $.Values.priorityClassName }}
       priorityClassName: {{ $.Values.priorityClassName }}
       {{- end }}
-{{- if .Values.affinity }}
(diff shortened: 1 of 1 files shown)


gsanchezgavier:
Thanks @freehck-codefresh for the contribution! 
We have to look this in more detail to check if this is compatible with the architecture of the integration. For instance KSM metrics are scraped from the infra pod running on the same node otherwise those metrics will be missing. There might be other issues that could appear just applying the affinity config to the Daemonset.

freehck-codefresh:
@gsanchezgavier 
On the one hand, I understand that your service is supposed to be deployed on each node of the cluster.
But in my specific case I just wanted to monitor some specific services and there were no need to deploy the full daemonset, I needed only several of them.

It's just a use-case when a user has sixty nodes, but need to send metric to NR only from couple of services.

paologallinaharbur:
Stale, closing it for now since not aligned with the architecture and would possibly create issues in the future 😕

PR #250, opened 2020-12-30
nri-prometheus: add advanced config flags to values

roobre:
#### Is this a new chart

Nope

Adds advanced configuration flags to `nri-prometheus`'s `values.yml`. Please note that changing this values can affect CPU and Memory consumption in unforeseen ways. We recommend to exercise care when changing `worker_threads`, and not to modify other flags unless instructed by the New Relic support team, as the default values cover the vast majority of use cases.

#### Checklist
- [x] Chart Version bumped
- (N/A) Variables are documented in the README.md

---

Related https://github.com/newrelic/nri-prometheus/issues/130

### DIFF
--- a/charts/nri-prometheus/values.yaml
+++ b/charts/nri-prometheus/values.yaml
@@ -105,7 +105,7 @@ config:
   require_scrape_enabled_label_for_nodes: true
 
   # Number of worker threads used for scraping targets.
-  # For large clusters with many (>100) endpoints, slowly increase until scrape
+  # For large clusters with many (>400) endpoints, slowly increase until scrape
   # time falls between the desired `scrape_duration`.
   # Increasing this value too much will result in huge memory consumption if too
   # many metrics are being scraped.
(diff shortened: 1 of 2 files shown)


charts/nri-prometheus/values.yaml
```diff
@@ -105,10 +105,21 @@ config:
   require_scrape_enabled_label_for_nodes: true
 
   # Number of worker threads used for scraping targets.
-  # For large clusters with many endpoints, increase to 10.
+  # For large clusters with many (>100) endpoints, slowly increase until scrape
```
paologallinaharbur, line 108: In the documentation we specify 400 as a safe value, we should be consistent

paologallinaharbur: https://docs.newrelic.com/docs/integrations/prometheus-integrations/install-configure-openmetrics/configure-prometheus-openmetrics-integrations-large-kubernetes-environments

paologallinaharbur (approved):

Issue #252, opened 2021-01-08
[nri-kube-events] Allow to define collector_url for the agent container


roobre:
**Bug description**

`values.yml` does not expose a field which allows to override the `collector_url` used by the infra agent to report metrics, which is breaking for customers that use non-standard endpoints.

**Version of Helm and Kubernetes**

Any

**Which chart?**

`nri-kube-events 1.4.0`

PR #255, opened 2021-01-11
[nri-kube-events]  config file for agent sidecar

paologallinaharbur:
Fix https://github.com/newrelic/helm-charts/issues/252

This follows the approach of nri-kubernetes, we add an object that if configured creates a cm that is mounted inside the sidecar
https://github.com/newrelic/helm-charts/blob/fb7be596931a13701596fa382e536b73293bac36/charts/newrelic-infrastructure/values.yaml#L124

#### Checklist
- [x] Chart Version bumped


charts/nri-kube-events/templates/configmap.yaml
```diff
@@ -18,3 +18,16 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
 {{ include "nri-kube-events.labels" . | indent 4 }}
+{{ if .Values.config }}
+---
+apiVersion: v1
+kind: ConfigMap
+data:
+    newrelic-infra.yml: |
+{{ toYaml .Values.config | indent 6 }}
```
roobre, line 27:
```suggestion
  newrelic-infra.yml: |
{{ toYaml .Values.config | indent 4 }}
```

charts/nri-kube-events/templates/configmap.yaml
```diff
@@ -18,3 +18,16 @@ metadata:
...
+metadata:
+  name: {{ template "nri-kube-events.fullname" . }}-agent-config
+  namespace: {{ .Release.Namespace }}
+  labels: 
+{{ include "nri-kube-events.labels" . | indent 4 }}
+{{ end }}
```
roobre, line 33: If you forget about the `\n` at the end of the file yo go to POSIX hell, where people build tools that do many things and do them wrong :)

charts/nri-kube-events/values.yaml
```diff
@@ -36,6 +36,29 @@ sinks:
...
+  # Infrastructure agent. If options have command line equivalents, New Relic
+  # will use the command line option to override any value set in this file.
+  #
+  #
+  # You can use this section in order to add any agent configuration.
+  # For example to add FedRAMP-compliant endpoints(https://docs.newrelic.com/docs/security/security-privacy/compliance/fedramp-compliant-endpoints#infrastructure-endpoints),
+  # to set verbose logs, etc
+  # verbose: 1
+  # FedRAMP
+  # collector_url: https://gov-infra-api.newrelic.com
```
roobre, line 57: I think we can avoid using this particular example.

charts/nri-kube-events/values.yaml
```diff
@@ -36,6 +36,29 @@ sinks:
   # The newRelicInfra sink sends all events to New relic.
   newRelicInfra: true
 
+# If you wish to provide your own newrelic.yml file include it under config:
+# the sample config file is included here as an example. Note that some option
+# overlap with environment variables: for example proxy
+# config:
```
roobre, line 42: How would this play with `nrStaging`?
https://github.com/newrelic/helm-charts/blob/c876e9fbdb253b9a5ea180c3531c4e2bb4802c41/charts/nri-kube-events/values.yaml#L122

I think we can keep both (for compatibility) but perhaps we could add a comment on which one will take priority.

roobre (changes requested):
I'd change the examples and the extra indent on the yaml, but otherwise it's looking good.

charts/nri-kube-events/templates/configmap.yaml
paologallinaharbur: I am a bit confused by the comment, I guess you want a `\n`, right? :)

charts/nri-kube-events/templates/configmap.yaml
paologallinaharbur: I fixed that, it was coming from the nri-kubernetes chart, but you are right no need for that

charts/nri-kube-events/values.yaml
paologallinaharbur: it is like the proxy, environment variable take precedence, i'll add a note for that

charts/nri-kube-events/values.yaml
paologallinaharbur: I added them since they are not listed in the standard documentation page, but only in this doc:
https://docs.newrelic.com/docs/security/security-privacy/compliance/fedramp-compliant-endpoints#infrastructure-endpoints

But you are right it might be too specific

roobre (approved):
lgtm

PR #262, opened 2021-01-14
[apm-php-agent] Add a chart for PHP agent APM

### LINKED ISSUE #127: [php-daemon]: There is no php-daemon Helm Chart
As a developer, I'd like to be able to deloy the New Relic PHP-Daemon as a chart with vendor-specific infrastructure configuration handled by New Relic.

I don't really want to maintain my own custom deployment of the PHP-Daemon.

I'd much prefer a simple chart install.

Obviously, running my own Helm Chart, or creating my own deployment, but I'd rather not.

nitneuk:
#### Is this a new chart

Yes

This PR adds a new chart that install the newrelic php agent to a kubenetes cluster. I guess it should be part of the `nri-bundle` to easily install one or more APMs along with infrastructure installation. This could be done using the `apm.php.enabled=true` configuration on the `nri-bundle` chart or installed in a "standalone" way.

  - fixes #127 

- Should I make another PR for the nri-bundle updates? 
- CI lint seems ok, but failed on building `nri-bundle` dependencies. It seems ok locally with `helm dependency build charts/nri-bundle`, what's wrong to it?  

#### Checklist
- [x] Chart Version bumped


janavenkat:
@nitneuk can you please follow

roobre:
Hi there,

Sorry for the delay answering this. At the time being, our team manages the helm charts but we don't have much context on the PHP agent. Specifically, I am not sure whether running a single instance of a PHP daemon is something recommended, as opposed to, for instance, embedding the daemon in your containers, run it as a sidecar, or as a pod in the same namespace.

I'll get in touch with the team in charge of the PHP APM solution to try to figure this out. In the meantime, could you elaborate the motivation for this @nitneuk? We'd be curious to hear about what made you lean towards this solution, as opposed to others (e.g. the two I mentioned before).

nitneuk:
Hi @roobre,

Thank you for the reply. 

By reading this documentation https://docs.newrelic.com/docs/agents/php-agent/advanced-installation/docker-other-container-environments-install-php-agent I understood that it's better to no setup the deamon inside the same container as the application. So, I decided to set the daemon in another container. Without recommendation (maybe I didn't found it?) on how to set up NR APM in kubernetes, I thought it was better to have a single daemon for one or multiple applications (but maybe it's not). 

Other motivations of having the apm daemon set as this, was to not duplicate the configuration and pods concerning the apm daemon over all our project applications in the cluster. The goals behind this was to not consume to much resources on the cluster by having a single apm daemon on each node and a service with a topolgy key. That way all applications pods can reach the deamon located in the same node. 

I would be glad to know what's the PHP APM team would recommand. Maybe there's drawbacks that I didn't see to have setup the APM like this.

roobre:
Hello,

Just wanted to give a quick heads up, we're still in the process of figuring out what's the preferred way to tackle this with the PHP APM team.

In the meantime, I want to try to align in the different options we have here and ask you again why the DaemonSet was your preferred choince. In my view, there are three different ways in which the PHP Daemon can be deployed in a K8S cluster:

1. As a sidecar with the application (each replica of the application pod will have the agent close)
2. As a Deployment
3. As a DaemonSet.

1 vs either 2 or 3 have their own tradeoffs which are complex in nature (and what we're in process of figuring out), but I'd like to explore a bit more the reasons you had for picking 3 over 2. I see in your chart that you specify a Service for the DaemonSet. As you might be already aware, the [Kuberentes Documentation](https://github.com/newrelic/helm-charts/pull/262) states:

> * Service: Create a service with the same Pod selector, and use the service to reach a daemon on a random node. (No way to reach specific node.)

If I'm reading this correctly, that would mean that the instrumented PHP workloads will not preferrably talk to the Agents they have "closer", but at random. While this stays true with the Deployment option, the latter has the plus of being able to scale as the application demands. With the DaemonSet, this is not the case, and might be an inconvenience.
Furthermore, if locality is not exploited, I see no reason to make this a DaemonSet and enforce unnecessary replicas on clusters with many nodes but not so many PHP workloads.

As I said, we're still figuring this out internally (specially the sidecar vs. non-sidecar options), but I want to take the chance to hear your thoughts on the points  above since, in my view, the DaemonSet option is perhaps the least fitting of the three.

Thanks in advance for your thoughts!

nitneuk:
Hi @roobre ,

Sorry for the late response, I was on holiday last week. 

My team and I got in touch with the NR support with that ticket id : 446274. You could see further information on our configuration in this ticket. 

As you understood, we decided to use the 3rd approach, to make the agent located in our applications container the closest possible of  the daemon. Indeed, by default, the Kubernetes Service would round robin over the pods that match the described selector. That way the agent would NOT access to the closest daemon by default. 

As you said, the daemonset documentation says : 
> Service: Create a service with the same Pod selector, and use the service to reach a daemon on a random node. **(No way to reach specific node.)**

That's why I've added a topology key on the service (https://github.com/newrelic/helm-charts/pull/262/files#diff-3c95061e0aaaee9bc95f9a36c03ecb78742aefcd27fdbc8c126b44d341c9e301R17-R19). From what I understood from the documentation (https://kubernetes.io/docs/concepts/services-networking/service-topology/#using-service-topology), the service topology would use the daemon located on the node that would match de same node's label (here "kubernetes.io/hostname" that correspond to the node name) as the agent's node. 

> If your cluster has Service Topology enabled, you can control Service traffic routing by specifying the topologyKeys field on the Service spec. This field is a preference-order list of Node labels which will be used to sort endpoints when accessing this Service. Traffic will be directed to a Node whose value for the first label matches the originating Node's value for that label. If there is no backend for the Service on a matching Node, then the second label will be considered, and so forth, until no labels remain.

As we has multiple replicas on our php application (located on our several nodes) we decided to use a DeamonSet over a Deployment as it appears to fit better for monitoring purposes. But you're wright concerning the deployment replicas, if we deploy more applications on our cluster, we may need additional daemons (more that 1 by node) and that way, maybe a Deployment with a service (with a topology key) and a minimum replicas (that match the number of nodes) would fit better. 

From the NR support ticket, I was not aware that : 
>  an agent needs to continue talking to the SAME daemon it previously was talking to.
So that would mean that an agent would have to always reach the same daemon as previous requests. But do you know if it's ok for application's agents replicas to reach multiple daemons? Even if an agent pod would always reach the same daemon ?

Thanks for you're feedback, it's really appreciated on our side.  I hope we would  successfully find the better solution at the end.

adri:
I'm also interested in how this can proceed. 

If I understand the [docs](https://docs.newrelic.com/docs/agents/php-agent/advanced-installation/docker-other-container-environments-install-php-agent/#overview) correctly, it is preferred that the agent and daemon are installed on different containers. 

These are the options for Kubernetes:

> 1. As a sidecar with the application (each replica of the application pod will have the agent close)
> 2. As a Deployment
> 3. As a DaemonSet.

With the information from the other comments here, I think 1. is the only real option for deploying the NewRelic daemon in Kubernetes. Did I get this correctly?

adri:
Hey :) Just checking in because I'm still wondering how to deploy the New Relic PHP agent best in Kubernetes. If there are any updates or investigations going on it would be great to know.

roobre:
Hey there,

Coming back to this PR as it seems it has become stale:

While I think the deployment mode this PR suggest is perfectly valid, I am not sure it is a model that can suit a significant majority of the use cases, as in some scenarios it might be better to deploy the agent as a sidecar or a Deployment due to resources or security concerns.

While it is a high quality contribution, which NR truly appreciates, we are afraid that incorporating this chart into our repository could easily be interpreted as a DaemonSet being the official suggested way of deploying the PHP agent, which is not the case as I outlined above.
For this reason I'm afraid we cannot accept this contribution at the moment.

Nevertheless, we are again thankful for your efforts on this :)

PR #230, opened 2020-12-04
[nri-kube-events] Fix 'kube-events' container restarts due to '/tmp' …

rbabyuk:
…RO fylesystem

#### Is this a new chart
no
#### What this PR does / why we need it:
kube-events container is restarted every 8-10m(?) due to 'cannot write log file to /tmp/newrelic...log fylesyste, is read-only'

#### Which issue this PR fixes
  - fixes #72

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version bumped
- [ ] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[mychartname]`)

PR #271, opened 2021-02-09
[nri-kube-events] bump version to 1.4.0

### LINKED ISSUE #72: [nri-kube-events] attempting to write warnings to RO filesystem
`nri-kube-events`'s `kube-events` container keeps crashing every few minutes as a watch warning causes the service to try to write to `/tmp`, which is a readonly filesystem. I am not sure if this is necessarily a bug in `kube-events` or the container should just be updated to allow it to write to a tmpfs like the side-car `infra-agent` container already can.

**Version of Helm and Kubernetes**

```
$ kubectl version
...
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.4", GitCommit:"8d8aa39598534325ad77120c120a22b3a990b5ea", GitTreeState:"clean", BuildDate:"2020-03-12T20:55:23Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
$ helm version
version.BuildInfo{Version:"v3.2.3", GitCommit:"8f832046e258e2cb800894579b1b3b50c2d83492", GitTreeState:"clean", GoVersion:"go1.13.12"}
```

**Which chart?**

`nri-kube-events` / `kube-events`

**What happened?**

When any of the watchers emit a warning, it attempts to log it to `/tmp`. This is a read-only filesystem which causes the container to crash and restart.

```
W0623 14:41:44.507228       1 reflector.go:302] pkg/mod/k8s.io/client-go@v0.0.0-20190620085101-78d2af792bab/tools/cache/reflector.go:98: watch of *v1.Event ended with: The resourceVersion for the provided watch is too old.
log: exiting because of error: log: cannot create log: open /tmp/nri-kube-events.newrelic-events-nri-kube-events-c6bfbd6ff-6bm5s.nri-kube-events.log.WARNING.20200623-144144.1: read-only file system
```

**What you expected to happen?**

`kube-events` warnings shouldn't cause the container to restart. This should either be made into a writable directory or `kube-events` should only send warnings to stdout.

**How to reproduce it?**

This is just a generic Rancher deployed K8s cluster on v1.17.4

roobre:
#### Is this a new chart

Nope

Update to the latest version of `nri-kube-events`

Fixes #72
Closes #230

#### Checklist
- [x] Chart Version bumped


charts/nri-bundle/requirements.lock
```diff
@@ -4,7 +4,7 @@ dependencies:
   version: 2.0.1
 - name: nri-prometheus
   repository: file://../nri-prometheus
-  version: 1.6.0
+  version: 1.6.2
```
roobre, line 7: I am not entirely sure why `helm dependency update` changed this, the `requirements.yaml` file still points to 1.6.0.

paologallinaharbur: The local version you had is 1.6.2 if you are in `origin/main` 

https://github.com/newrelic/helm-charts/releases/tag/nri-prometheus-1.6.2

Even if 
```
  - name: nri-prometheus
    repository: file://../nri-prometheus
    condition: prometheus.enabled
    version: 1.6.0
```
is referenced, since you are checking it locally it gets the 1.6.2 and in the lock the inconcruence is shown

paologallinaharbur: I have never understood why we check for the chart locally and not in the helm repo.

roobre: Interesting, I hadn't notice the `file://` URI. Should I uncommit the change just in case then?

paologallinaharbur: I would bump nri-prometheus as well and rerun helm dependency update.

Now I believe that we are packaging the 1.6.2 version calling it 1.6.0

charts/nri-kube-events/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic Kube Events
 name: nri-kube-events
-version: 1.5.0
-appVersion: 1.3.0
```
paologallinaharbur, line 5: I am not 100% sure that the CI is capable to bump both dependency and bundle in one PR.
We can either check or split those in two

paologallinaharbur (approved):

Issue #282, opened 2021-02-19
[newrelic-infrastructure] Do not grant permission to "secret" if control plane monitoring is not configured


roobre:
The default RBAC role for the cluster grants access to read secrets in other namespaces. This is done to allow the integration to fetch the mTLS secrets for etcd, which are in turn required to gather metrics from it.

However this permission can be viewed as unnecessary, specially if control plane monitoring is not configured.

**Describe the solution you'd like**

It would be good if the helm chart skipped this permission if control plane monitoring is not configured for the cluster (e.g. `etcdTlsSecretName` is not defined).

**Describe alternatives you've considered**

The default rbac config can be ommited completely by setting `rbac.create` to false. However this requires the user to manually create an rbac with all the other requirements for the integration to work.

roobre:
Additionally, it would be good if the current role could be limited to access just the namespace where the control plane resides. I think we could specify something in the values.yml for this, like `etcdNamespace` defaulting to blank, in which case the role would be unbound as it is now.

paologallinaharbur:
the idea is to create a rolebinding for the secrets instead of the clusterrolebinding granting access to the newrelic namespace and the etcd one

PR #290, opened 2021-02-23
[newrelic-infrastructure] reduce scope of permissions

paologallinaharbur:
#### Is this a new chart
no
Reducing the scope of the cluster role binding that is currently too broad

  - fixes https://github.com/newrelic/helm-charts/issues/282

#### Checklist
- [x] Chart Version bumped


charts/newrelic-infrastructure/templates/rolebinding.yaml
```diff
@@ -0,0 +1,35 @@
...
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "newrelic.fullname" . }}-secrets
+subjects:
+- kind: ServiceAccount
+  name: {{ template "newrelic.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace }}
+{{- if .Values.etcdTlsSecretNamespace}}
```
roobre, line 17: I don't get why these are two different cases. How are we going to scope the secret (`limitSecretAccess == true`) if `etcdTlsSecretNamespace` is not defined?

Couldn't we just replace the whole `limitSecretAccess` flag by a `etcdTlsSecretNamespace != ""` check?

paologallinaharbur: You are right, the change is safe enough to be done without introducing a new flag.

By the way we should change as well the static manifests in the nri-kubernetes repo

charts/newrelic-infrastructure/templates/clusterrole.yaml
```diff
@@ -27,4 +26,17 @@ rules:
...
+metadata:
+  labels: {{ include "newrelic.labels" . | indent 4 }}
+  name: {{ template "newrelic.fullname" . }}-secrets
+rules:
+  - apiGroups: [""]
+    resources:
+      - "secrets"
+    verbs: ["get", "list"]
 {{- end -}}
+{{- end -}}
```
roobre, line 42: Newline before `EOF`

paologallinaharbur: 🙅

roobre (approved):
If manual testing passes, lgtm

PR #293, opened 2021-02-25
[nri-kube-events] update events-forwarder version

arvdias:
#### Is this a new chart
no

updates events-forwarder version

#### Checklist
- [x] Chart Version bumped


charts/nri-kube-events/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic Kube Events
 name: nri-kube-events
-version: 1.6.1
+version: 1.6.2
```
paologallinaharbur, line 4: should we change the minor? 
(Not that important since only the forwarder changed and not the main app)

paologallinaharbur (approved):

charts/nri-kube-events/Chart.yaml
roobre: +1 for changing the minor, I prefer to bump minor in charts whenever the minor appVersion of any component is bumped as well (and same goes for patch or major).

PR #306, opened 2021-03-19
[all] Support additional podLabels being provided

MarcusNoble:
#### Is this a new chart
No

Adds a helper to all charts to handle passing in `podLabels` and `global.podLabels` (for when using nri-bundle) and applies them to all created pods.

  - fixes #

`nri-bundle` will need updating following this change to uppdate the dependency chart versions.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure/Chart.yaml
+++ b/charts/newrelic-infrastructure/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic Infrastructure Agent as a DaemonSet
 name: newrelic-infrastructure
-version: 2.3.1
+version: 2.4.0
 appVersion: 2.3.0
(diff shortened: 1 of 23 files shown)


roobre:
Hi, many thanks for taking the time to write this PR!

I was wondering if you could elaborate a bit more on the use case for this. We already allow to specify user-defined pod annotations, which are the ones that one typically needs to change. Where does the need to also specify pod labels come from?

Thanks!

MarcusNoble:
We're using New Relic as part of a platform we're building. As part of the platform we apply stadard labels to all created pods for reporting use and to identify what manages them. We also leverage labels with some custom operators that filter based on label at the MutationWebhook level.

roobre:
This feature has been incorporated into `newrelic-infrastructure` chart version 3 here: https://github.com/newrelic/helm-charts/pull/624

The same feature will be incorporated in other chart as a part of efforts scheduled for this quarter.

Issue #295, opened 2021-03-01
[nri-bundle] Update KSM dependency


roobre:
**Is your feature request related to a problem? Please describe.**

`nri-bundle` pulls `kube-state-metrics` as a dependency. However, it pulls `1.9.7` from the "stable" Helm charts repo, which is deprecated.

**Describe the solution you'd like**

The dependency should be updated to `kube-state-metrics:1.9.8` from [the ksm repo](https://kubernetes.github.io/kube-state-metrics/)

PR #307, opened 2021-03-23
[nri-bundle] Update KSM repository and fix committed dependencies 

douglascamata:
#### Is this a new chart

No.

We should stop using the now long time deprecated official Helm charts repository to avoid problems. 

I also took the freedom to fix how dependencies are handled in the `nri-bundle` chart: only KSM should be committed to avoid problems when its repository could be down. All the other charts are pulled from the same repository as `nri-bundle`, so they don't need to be committed together.

This also changes slightly the configuration of the release chart action. I'll be watching the release build to confirm that it works properly.

#### Checklist
- [x] Chart Version bumped


gsanchezgavier (approved):
thanks! lgtm please double check on merging

paologallinaharbur:
Please after the merge remember to update release notes!

paologallinaharbur:
Fix https://github.com/newrelic/helm-charts/issues/295

PR #334, opened 2021-04-19
[nri-metadata-injection] Migrate cert generation to kube-webhook-certgen

invidian:
#### Is this a new chart
image

This PR migrates cert generation manifests to use kube-webhook-certgen, which should be more robust than existing k8s-webhook-cert-manager solution, as:
- It does not rely on kube-controller-manager configuration, as it uses
  self-signed certificates.
- Certificates are valid for 100 years so they will never expire BUT
  they are rotated on every Helm upgrade via pre-upgrade hook.
- It is maintained externally.

There are some drawbacks though:
- On serverless platforms like AWS Fargate, installing and upgrading
  chart takes ~3 minutes, as hook must converge before actual
  manifests are applied.

Manifests below has been copied from [1], I stripped unused fields and
added extra volumes fields, which are actually supported by this chart
for Job Pod.

[1] https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/templates/prometheus-operator/admission-webhooks/job-patch

Perhaps some extra testing should be done to ensure this works as expected.

Signed-off-by: Mateusz Gozdek <mateusz@kinvolk.io>

#### Checklist
- [x] Chart Version bumped


paologallinaharbur:
Hello, the ci/cd is currently failing with the error message:
```
{"level":"error","ts":"2021-04-19T20:03:19.724Z","caller":"server/main.go:50","msg":"failed to load key pair","err":"open /etc/tls-key-cert-pair/tls.crt: no such file or directory","stacktrace":"main.main\n\t/home/runner/work/k8s-metadata-injection/k8s-metadata-injection/cmd/server/main.go:50\nruntime.main\n\t/opt/hostedtoolcache/go/1.16.3/x64/src/runtime/proc.go:225"}
{"level":"info","ts":"2021-04-19T20:03:19.724Z","caller":"server/main.go:82","msg":"starting the TLS readiness server"}
{"level":"info","ts":"2021-04-19T20:03:19.725Z","caller":"server/main.go:89","msg":"starting the webhook server"}
2021/04/19 20:03:22 http: TLS handshake error from 10.244.0.1:5017: tls: certificate private key (<nil>) does not implement crypto.Signer
```

roobre:
> Manifests below has been copied from [1], I stripped unused fields and added extra volumes fields, which are actually supported by this chart for Job Pod.

What's the motivation for doing this vs just adding a helm dependency?

invidian:
>What's the motivation for doing this vs just adding a helm dependency?

There is no independent chart for it (though maybe there could be one).

invidian:
>Hello, the ci/cd is currently failing with the error message:

The path to the files in the secret changes and it wasn't represented in Pod configuration. Fixed that now.

charts/nri-metadata-injection/templates/admission-webhooks/job-patch/job-createSecret.yaml
```diff
@@ -0,0 +1,45 @@
...
+          imagePullPolicy: {{ .Values.jobImage.pullPolicy }}
+          args:
+            - create
+            - --host={{ template "nri-metadata-injection.fullname" . }},{{ template "nri-metadata-injection.fullname" . }}.{{ .Release.Namespace }}.svc
+            - --namespace={{ .Release.Namespace }}
+            - --secret-name={{ template "nri-metadata-injection.fullname" . }}
+	        {{- if .Values.jobImage.volumeMounts }}
+	        volumeMounts:
+	        {{- include "tplvalues.render" ( dict "value" .Values.jobImage.volumeMounts "context" $ ) | nindent 10 }}
+	        {{- end }}
```
roobre, line 33: Tabs here

roobre:
I'd prefer if we could clarify that these charts were borrowed from somewhere, and quote that source. WDYT about adding a `README.md` in the `job-patch` dir containing this?

charts/nri-metadata-injection/templates/admission-webhooks/job-patch/job-createSecret.yaml
invidian: Thanks, fixed. I need to fix my vim config finally..

invidian:
>I'd prefer if we could clarify that these charts were borrowed from somewhere, and quote that source. WDYT about adding a README.md in the job-patch dir containing this?

Sure.

roobre (approved):
LGTM!

douglascamata:
I didn't test this yet, but is does compatible with [cert-manager](https://cert-manager.io/)?

douglascamata:
Great PR, @invidian, thanks a lot. I didn't know about Helm hooks before and this made me discover them!

invidian:
>I didn't test this yet, does it keep compatibility with cert-manager?

Ah. Good point. In case of `cert-manager` integration, keys in the certificate Secret will be different. That should be conditional.

douglascamata:
I see already a bunch of templates checking the value in `.Values.certManager.enabled`, so I thought it was handled. But double checking doesn't hurt 😄

invidian:
>Ah. Good point. In case of cert-manager integration, keys in the certificate Secret will be different. That should be conditional.

Actually, generated secret should have the right keys, fortunately `kube-webhook-certgen` allows to customize that, so I use standard keys for `kubernetes.io/tls` secret type. I'm actually surprised it does not generate this secret type by default with those keys... Now also extra env variables on deployment are not needed.

charts/nri-metadata-injection/templates/deployment.yaml
```diff
@@ -20,9 +20,6 @@ spec:
       labels:
         {{- include "nri-metadata-injection.labels" . | nindent 8 }}
     spec:
-      {{- if (and (not .Values.customTLSCertificate) (not .Values.certManager.enabled)) }}
```
paologallinaharbur, line 23: Why did we need the service account?

paologallinaharbur: I did not find any reason why we were adding such (broad) permissions to it

paologallinaharbur: @douglascamata do you remember?

invidian: I don't know. It was there since the beginning of git history. Maybe in the past certificates were generated as part of this deployment? I went trough the k8s-metadata-injection code and I don't think it needs the service account. I can remove it in separate commit though.

douglascamata: I'm not sure as well about the need for the service account in this deployment. It's completely stateless, it doesn't do any call to the api server... 🤔

paologallinaharbur (commented):
The upgrade path seems to be failing:

```
$ helm install test newrelic/nri-metadata-injection -f values.yaml
$ helm upgrade test ./charts/nri-metadata-injection -f values.yaml
[...] waits endlessy
$k logs test-nri-metadata-injection-admission-create-vwm7x
W0422 14:54:51.124992       1 client_config.go:608] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
{"level":"fatal","msg":"got secret, but it did not contain a 'ca' key","source":"k8s/k8s.go:114","time":"2021-04-22T14:54:51Z"}
```

invidian:
>The upgrade path from the last version seems to be failing confused :

I changed the secret name to force it to be re-created, which solves this issue.

I also had issues with removing ServiceAccount, so `default` service account must be specified explicitly for time being to be able to do the update.

paologallinaharbur (approved):
Thanks, tomorrow we can merge and release! if the secret name changes, we should take care of updating the documentation as well on the docs website

charts/nri-metadata-injection/templates/deployment.yaml
```diff
@@ -56,7 +57,7 @@ spec:
       volumes:
       - name: tls-key-cert-pair
         secret:
-          secretName: {{ template "nri-metadata-injection.fullname" . }}
+          secretName: {{ template "nri-metadata-injection.fullname" . }}-admission
```
roobre, line 60: This change breaks the cert-manager setup, which is creating the old name

roobre, line 60: This change breaks the cert-manager setup, which is creating the old name

https://github.com/newrelic/helm-charts/blob/b4d3421a21d2291f664547f1f190495fbd40c2ef/charts/nri-metadata-injection/templates/cert-manager.yaml#L41

roobre (changes requested):

charts/nri-metadata-injection/templates/deployment.yaml
invidian: I fixed the name to be consistent. Hopefully that won't break cert-manager.

roobre (approved):
Seems to be working great with `cert-manager` and the latest changes!
LGTM!

roobre:
@invidian Please rebase latest `master` and I'll merge as soon as I can :)

invidian:
>@invidian Please rebase latest master and I'll merge as soon as I can :)

Rebased.

roobre:
Big thank you and congrats for this @invidian 🎉

PR #336, opened 2021-04-20
[newrelic-pixie] Add chart for newrelic-pixie integration

douglascamata:
#### Is this a new chart

Yes.

Creates a new chart for the newrelic-pixie integration.

#### Checklist
- [x] Chart Version bumped


charts/newrelic-pixie/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Configuration
+
+| Parameter                     | Description                                                  | Default                    |
+| ----------------------------- | ------------------------------------------------------------ | -------------------------- |
+| `global.cluster` - `cluster`  | The cluster name for the Kubernetes cluster. Required.       |                            |
+| `global.licenseKey` - `licenseKey` | The New Relic license key (stored in a secret). Required.    |                            |
+| `apiKey`                      | The Pixie API key (stored in a secret). Required.            |                            |
+| `verbose`                     | Whether the integration should run in verbose mode or not.   | false                      |
+| `global.customSecretName` - `customSecretName` | Name of the Secret object where the New Relic license and Pixie API key are stored                                                                                                                                                                         |                                 |
```
roobre, line 18: From the readme I did not get clearly that these were _alternatives_ to provide the values directly. Perhaps we can clarify it a bit?

charts/newrelic-pixie/values.yaml
```diff
@@ -0,0 +1,28 @@
...
+# customSecretName:
+# The key in the custom secret that contains the New Relic license key
+# customSecretLicenseKey:
+# The key in the custom secret that contains the Pixie API key
+# customSecretApiKeyKey:
+
+image:
+  repository: newrelic/newrelic-pixie-integration
+  tag: latest
+  pullPolicy: Always
```
roobre, line 21: I'd wait to merge this until we have a production ready image with a versioned tag, even if it's beta.

charts/newrelic-pixie/values.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+resources:
+  limits:
+    memory: 80M
+  requests:
+    cpu: 100m
+    memory: 30M
```
roobre, line 28: We usually don't have default resources in the charts (specially not `requests`), as they might prevent installation on some environments.

douglascamata: We have them defined for the `newrelic-infrastructure` chart. It's better that it's not installed than not running properly due to lack of resources.

charts/newrelic-pixie/templates/secret.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if (include "newrelic-pixie.areValuesValid" .) }}
+{{- $licenseKey := include "newrelic-pixie.licenseKey" . -}}
+{{- $apiKey := include "newrelic-pixie.apiKey" . -}}
+apiVersion: v1
```
roobre, line 4: I'd add an if above so we don't generate an empty secret if neither `$licenseKey` nor `$apiKey` are set.

douglascamata: `{{- if (include "newrelic-pixie.areValuesValid" .) }}` is checking for license key and api key. 👍

alejandrodnm (commented):
@douglascamata Do we need to add an option for sending data to staging like we don with the other charts?

charts/newrelic-pixie/README.md
douglascamata: I'll add a section for this 👍

charts/newrelic-pixie/ci/test-values.yaml
```diff
@@ -0,0 +1,7 @@
+global:
+  licenseKey: 1234567890abcdef1234567890abcdef12345678
+  apiKey: 1234567890abcdef
+  cluster: test-cluster
+pixieClusterId: foobar
+image:
+  tag: 0.1.0-pre
```
roobre, line 7: Is there any reason for overriding the image here? I think I'd forget to update this all the time.

douglascamata: Good point. I think now that I added the `appVersion` to the chart in the most recent commits I don't need it. Let me try to remove it and see if the CI passes.

charts/newrelic-pixie/templates/_helpers.tpl
```diff
@@ -0,0 +1,142 @@
...
+  {{- else -}}
+      {{- .Values.apiKey | default "" -}}
+  {{- end -}}
+{{- else -}}
+    {{- .Values.apiKey | default "" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return the customSecretName where the New Relic license and Pixie API key will be stored
```
roobre, line 86: "Will be" or "are currently being"?

douglascamata: @roobre this `newrelic-pixie.customSecretName` is meant to allow a customer that already has a secret created in the cluster and want to reuse it. It's not meant to create a secret with such name.

douglascamata: I'll clarify this in the README

charts/newrelic-pixie/templates/secret.yaml
roobre: But that will evaluate to `true` if `customSecretName` is used, right? In that case I don't see any point to render this secret.

charts/newrelic-pixie/README.md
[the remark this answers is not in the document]
douglascamata: But they are decoupled. Customers can provide the secret name and then the name of each key below.

charts/newrelic-pixie/templates/secret.yaml
douglascamata: Oh yes! I missed this. Fixing it.

charts/newrelic-pixie/values.yaml
roobre: You are right. I think we do no have them on nri-prometheus but I'm not sure about the reason. Let's go with them :)

charts/newrelic-pixie/templates/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+            secretKeyRef:
+              {{- if (include "newrelic-pixie.apiKey" .) }}
+              name: {{ template "newrelic-pixie.fullname" . }}-secrets
+              key: pixieApiKey
+              {{- else }}
+              name: {{ include "newrelic-pixie.customSecretName" . }}
+              key: {{ include "newrelic-pixie.customSecretApiKeyKey" . }}
+          {{- end }}
+        - name: PIXIE_CLUSTER_ID
+          {{- if .Values.pixieClusterId }}
```
alejandrodnm, line 49: Where does this value come from?

douglascamata: It's intentionally not documented. We should load this from a secret that already exists in the cluster. But in the CI it doesn't exist, so I need another way to send it there.

charts/newrelic-pixie/README.md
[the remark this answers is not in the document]
roobre: From what I can tell, I cannot specify a different secret name for the two keys, right?

The secret name is the same for both according to: https://github.com/newrelic/helm-charts/pull/336/files#diff-6165171244721798d8a1814595418a53d9338d7d660d67299ebd28ebf75a7349R35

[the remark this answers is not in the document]
douglascamata: Fixed in 361c3e5

roobre (approved):
lgtm

charts/newrelic-pixie/Chart.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
```
paologallinaharbur, line 1: does it make sense to bump this to v2 or we want to support Helm 2 as well?

Updating this is in the future might be more painful

charts/newrelic-pixie/templates/deployment.yaml
roobre:
```suggestion
          {{- if .Values.pixieClusterId }}
          {{- /* .Values.pixieClusterId is only used for CI tests, for regular operation it's mounted from pixie secrets. */ }}
```

charts/newrelic-pixie/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+Then, to install this chart, run the following command:
+
+```sh
+helm install newrelic/newrelic-pixie \
+  --set cluster=<Kubernetes cluster name> \
+  --set licenseKey=<Your New Relic license key> \
+  --set apiKey=<Your Pixie API key> \ 
+  --namespace pl \
+  --generate-name
```
paologallinaharbur, line 39: Isn't better to specify an example name instead of using `--generate-name`?

charts/newrelic-pixie/values.yaml
```diff
@@ -0,0 +1,30 @@
+# IMPORTANT: The Kubernetes cluster name
+# https://docs.newrelic.com/docs/kubernetes-monitoring-integration
+# cluster: ""
```
paologallinaharbur, line 3: in the readme these values are documented with the `global.`. I would try do document both, or provide an example for both in the values.yaml

For example in https://github.com/newrelic/helm-charts/tree/master/charts/nri-kube-events it is done in that way

charts/newrelic-pixie/README.md
```diff
@@ -0,0 +1,71 @@
...
+| `verbose`                     | Whether the integration should run in verbose mode or not.   | false                      |
+| `global.customSecretName` - `customSecretName` | Name of an existing Secret object, not created by this chart, where the New Relic license is stored                                                                                                                                                                         |                                 |
+| `global.customSecretLicenseKey` - `customSecretLicenseKey` | Key in the existing Secret object, indicated by `customSecretName`, where the New Relic license key is stored.                                                                                                                                                             |                                 |
+| `customSecretApiKeyName` | Name of an existing Secret object, not created by this chart, where the Pixie API key is stored.    |                                 |
+| `customSecretApiKeyKey` | Key in the existing Secret object, indicated by `customSecretApiKeyName`, where the Pixie API key is stored.   |                                 |
```
paologallinaharbur, line 25: I do not like that much this name 😅 KeyKey is quite confusing

paologallinaharbur: but I totally understand that since it is the key of the key it is normal to have it in this way..

But I would still prefer
```
customSecretApiKey:
  key: asd
  name: asd
```

paologallinaharbur (approved):
Just really small things, if you want to check the comments

charts/newrelic-pixie/Chart.yaml
douglascamata: This is the api version of the chart itself. I can update it. 👍

charts/newrelic-pixie/README.md
douglascamata: Me neither, but that's what it is: a key where the api key is stored. Do you have suggestions? I couldn't find any better name.

charts/newrelic-pixie/README.md
douglascamata: No. In fact, it's dangerous for us to specify a fixed name here. Imagine the scenario where a cluster is used by two different teams, on different namespaces. One team installs this chart and configures it for them. Then our instructions tell the other team to also install it with the same name, completing breaking the chart for the previous team.

If the default behavior is generate name, this problem doesn't happen.

charts/newrelic-pixie/Chart.yaml
paologallinaharbur: but v2 works with helm3 only I believe

charts/newrelic-pixie/README.md
paologallinaharbur: not really unfortunately 😕

charts/newrelic-pixie/Chart.yaml
douglascamata: @paologallinaharbur in fact, it's better to keep api version 1 here. We still need to support Helm 2 and Helm 3 is compatible with with this api version. We should only move when we know that Helm 2 isn't supported anymore or that there's a Helm 4 that isn't compatible with api version 1.

charts/newrelic-pixie/templates/deployment.yaml
ulm0: Hi, i'm deploying this new integration, and with no value for pixieClusterId, where is this value supposed to come from? because the pod newrelic-bundle-newrelic-pixie-* is not being able to run due to the following errors:

- At first pl-cluster-secrets is not present
- then, pl-cluster-secrets does not contain cluster-id (Error: couldn't find key cluster-id in Secret newrelic/pl-cluster-secrets)

douglascamata: Hey @ulm0! This key should appear in the secret after the Pixie cloud connector pod correctly authenticates and exchange some data with the Pixie Cloud.

PR #338, opened 2021-04-21
[newrelic-infrastructure] better detection for enableProcessMetrics

roobre:
#### Is this a new chart

No

Previously, if the user defined `enableProcessMetrics` to `false`, it wouldn't be printed into the template. This could be workarounded by definining it as `"false"` in the values file, but this is not possible from the command line with `--set`.

This PR implements an alternative (and perhaps rather hacky) approach to detect an undefined `.Values.enableProcessMetrics`.

Any less hacky solution is welcome.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure/templates/daemonset.yaml
+++ b/charts/newrelic-infrastructure/templates/daemonset.yaml
@@ -136,6 +136,7 @@ spec:
             - name: "DISCOVERY_CACHE_TTL"
               value: {{ .Values.discoveryCacheTTL | quote }}
             {{- end }}
+            {{- /* Ugly hack to check if .Values.enableProcessMetrics is defined: */}}
             {{- if ne (print .Values.enableProcessMetrics) "<nil>" }}
             - name: "NRIA_ENABLE_PROCESS_METRICS"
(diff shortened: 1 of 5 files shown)


charts/newrelic-infrastructure/templates/daemonset.yaml
```diff
@@ -136,7 +136,7 @@ spec:
             - name: "DISCOVERY_CACHE_TTL"
               value: {{ .Values.discoveryCacheTTL | quote }}
             {{- end }}
-            {{- if .Values.enableProcessMetrics }}
+            {{- if ne (print .Values.enableProcessMetrics) "" }}
```
paologallinaharbur, line 139: ohh.. smart 😅

paologallinaharbur (approved):

paologallinaharbur:
PLease remember to add changelog :)

charts/newrelic-infrastructure/templates/daemonset.yaml
roobre: Smart as an euphemism for ugly hahahaha

paologallinaharbur (approved):

PR #340, opened 2021-04-23
[WIP] Add nri-k8s-operator chart

invidian:
Signed-off-by: Mateusz Gozdek <mateusz@kinvolk.io>

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [ ] Chart Version bumped
- [ ] Variables are documented in the README.md
- [ ] Title of the PR starts with chart name (e.g. `[mychartname]`)

PR #347, opened 2021-04-27
[newrelic-infra-operator] Add chart for the New Relic Infra Operator

roobre:
Helm chart for the New New Relic K8S Operator, which allows automatic injection of the infrastructure sidecar on EKS Fargate clusters.

Please note the name is still a work in progress.

### Pending tasks

- [x] Configuration yaml to be passed to the opeartor
- [x] ClusterRoleBinding
- [x] Specify a second (higher) set of resources for the sidecar that monitors KSM

---
Follow-up of https://github.com/newrelic/helm-charts/pull/340


.gitignore
```diff
@@ -36,4 +36,4 @@ Session.vim
 
 .history
 
-values.yaml
+/values.yaml
```
invidian, line 39: This commit can be dropped now after rebasing on latest master, it was only temporarily.

roobre: rebased

charts/nri-k8s-operator/values.yaml
```diff
@@ -0,0 +1,87 @@
...
+  requests:
+    cpu: 100m
+    memory: 30M
+
+serviceAccount:
+  # Specifies whether a ServiceAccount should be created for the job and
+  # the deployment.
+  #
+  # When `customTLSCertificate` is set to true, the service account is not
+  # necessary and it's not created in the case of `create=true`.
```
invidian, line 44: This is not correct in this chart.

charts/nri-k8s-operator/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 description: A Helm chart to deploy the New Relic Kubernetes Operator.
 name: nri-k8s-operator
 version: 0.1.0
-appVersion: 0.1.0
+appVersion: 1.4.0
```
invidian, line 5: Why that?

invidian: Shouldn't we start releasing from 0.1.0 following semver?

charts/nri-k8s-operator/Chart.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+description: A Helm chart to deploy the New Relic Kubernetes Operator.
+name: nri-k8s-operator
```
paologallinaharbur, line 3: Remember to update this

paologallinaharbur: :trollface:

charts/nri-k8s-operator/values.yaml
```diff
@@ -0,0 +1,98 @@
...
+# Ref: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts
+timeoutSeconds: 30
+
+# Operator configuration
+# The following are the default values for the operator
+config:
+  # sidecarInjection: {}  # Sidecar injection can be disabled by setting it to an empty object
+
+  sidecarInjection:  # List of sidecar injection policies, indexed by namespace
+    "*":  # This policy applies to all namespaces that do not have a specific entry in this map
```
paologallinaharbur, line 83: Would make sense to have it as a regex?

charts/nri-k8s-operator/Chart.yaml
roobre: Definitely, this is just the same placeholder for the image we're using right now. Will update it to 0.1.0 when we have the real one.

charts/nri-k8s-operator/Chart.yaml
roobre: As well as the name of the folder 😅

charts/nri-k8s-operator/values.yaml
roobre: If we add regex support, we'd need to convert this into a list to guarantee order of precedence. Might be a useful feature, but not sure if there will be demand for it. Might also be confusing for users, where `"custom"` would match `custom-namespace`.

roobre: Another possible approach would be to put regex matchers between slashes, like `/[Rr]egex/`. We can definitely support this if we think it's valuable.

charts/nri-k8s-operator/Chart.yaml
roobre: Done!

charts/newrelic-infra-operator/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,31 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "newrelic-infra-operator.fullname" . }}-agent-mock
+  labels:
+    {{- include "newrelic-infra-operator.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "newrelic-infra-operator.fullname" . }}-agent-mock
```
paologallinaharbur, line 26: creating a comment not to forget to update this :)

charts/nri-k8s-operator/values.yaml
roobre: Adding regexes increases complexity and makes following policies a bit less straightforward, so we decided to left it out for the time being.

charts/newrelic-infra-operator/templates/clusterrolebinding.yaml
roobre: omw

roobre: Done!

charts/newrelic-infra-operator/templates/clusterrole.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "newrelic-infra-operator.fullname" . }}
+  labels:
+    {{- include "newrelic-infra-operator.labels" . | nindent 4 }}
+rules:
+  - apiGroups: ["rbac.authorization.k8s.io"]
+    resources: ["clusterrolebindings"]
```
paologallinaharbur, line 9: In case we want to create automatically secrets in user namespace, we would need as well get, update and delete (for garbage collect).

Otherwise the user should create it manually or we should mount the license is a plain env var

invidian: I think it's reasonable to add a secret permissions right now.

roobre: In the past we've had users quite reticent to grant `get` permissions to secretsa cross the cluster. Would this be doable with just `create` and `update` and `delete`?

charts/newrelic-infra-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,75 @@
...
+      {{- if .Values.image.pullSecrets }}
+      imagePullSecrets:
+        {{- toYaml .Values.image.pullSecrets | nindent 8 }}
+      {{- end }}
+      containers:
+      - name: {{ template "newrelic-infra-operator.name" . }}
+        image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+        imagePullPolicy: "{{ .Values.image.pullPolicy }}"
+        env:
+        - name: clusterName
```
paologallinaharbur, line 37: This should be CLUSTER_NAME

charts/newrelic-infra-operator/Chart.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+description: A Helm chart to deploy the New Relic Kubernetes Operator.
+name: newrelic-infra-operators
```
invidian, line 3:
```suggestion
name: newrelic-infra-operator
```

roobre: Nice catch!

invidian (commented):
I think we still miss:
- `PodDisruptionBudget`
- `post-install` and `post-upgrade` hooks with `kubectl rollout` to restart the webhook.

charts/newrelic-infra-operator/README.md
```diff
@@ -0,0 +1,61 @@
+# newrelic-metadata-injection
```
invidian, line 1: Needs updating.

charts/newrelic-infra-operator/Chart.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+description: A Helm chart to deploy the New Relic Kubernetes Operator.
```
invidian, line 2:
```suggestion
description: A Helm chart to deploy the New Relic Infrastructure Kubernetes Operator.
```

charts/newrelic-infra-operator/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Chart Details
+
+This chart will deploy the [New Relic Infrastructure metadata injection webhook][1].
```
invidian, line 5: Needs updating.

charts/newrelic-infra-operator/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Configuration
+
+| Parameter                     | Description                                                  | Default                    |
+| ----------------------------- | ------------------------------------------------------------ | -------------------------- |
+| `cluster`                     | The cluster name for the Kubernetes cluster.                 |                            |
+| `injectOnlyLabeledNamespaces` | Limit the injection of metadata only to specific namespaces that match the label `newrelic-metadata-injection: enabled`. | false |
+| `image.repository`            | The container to pull.                                       | `newrelic/k8s-metadata-injection`   |
```
invidian, line 13: I suggest we align table using http://markdowntable.com/.

charts/newrelic-infra-operator/README.md
```diff
@@ -0,0 +1,61 @@
...
+| `affinity`                    | Node affinity to use for scheduling                          | `{}`                                |
+
+## Example
+
+Make sure you have [added the New Relic chart repository.](../../README.md#installing-charts)
+
+Then, to install this chart, run the following command:
+
+```sh
+helm install newrelic/newrelic-mutation-webhook --set cluster=my_cluster_name
```
invidian, line 46: Needs updating.

charts/newrelic-infra-operator/README.md
```diff
@@ -0,0 +1,61 @@
...
+The default set of resources assigned to the pods is shown below:
+
+    resources:
+      limits:
+        memory: 80M
+      requests:
+        cpu: 100m
+        memory: 30M
+
+[1]: https://docs.newrelic.com/docs/integrations/kubernetes-integration/link-your-applications/link-your-applications-kubernetes#configure-injection
```
invidian, line 60: This can be removed I think?

charts/newrelic-infra-operator/templates/NOTES.txt
```diff
@@ -0,0 +1,23 @@
+Your deployment of the New Relic metadata injection webhook is complete. You can check on the progress of this by running the following command:
```
invidian, line 1: Needs updating.

charts/newrelic-infra-operator/templates/admission-webhooks/job-patch/clusterrole.yaml
```diff
@@ -0,0 +1,30 @@
+{{- if (and (not .Values.customTLSCertificate) (not .Values.certManager.enabled)) }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "newrelic-infra-operator.fullname" . }}-admission
```
invidian, line 5: I wonder if we should again apply `truncate 63` when we append to `fullname` or `name` to avoid issues with too long names.

charts/newrelic-infra-operator/templates/admission-webhooks/job-patch/clusterrole.yaml
```diff
@@ -0,0 +1,30 @@
...
+rules:
+  - apiGroups:
+      - admissionregistration.k8s.io
+    resources:
+      - mutatingwebhookconfigurations
+    verbs:
+      - get
+      - update
+{{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
+{{- if semverCompare "> 1.15.0-0" $kubeTargetVersion }}
```
invidian, line 21: Given that we want to target Kubernetes 1.16+, I think we can remove this and simplify.

charts/newrelic-infra-operator/templates/admission-webhooks/mutatingWebhookConfiguration.yaml
```diff
@@ -0,0 +1,36 @@
...
+  rules:
+  - operations: ["CREATE"]
+    apiGroups: [""]
+    apiVersions: ["v1"]
+    resources: ["pods"]
+{{- if .Values.injectOnlyLabeledNamespaces }}
+  namespaceSelector:
+    matchLabels:
+      newrelic-metadata-injection: enabled
+{{- end }}
```
invidian, line 31: I guess this needs updating?

charts/newrelic-infra-operator/templates/clusterrole.yaml
```diff
@@ -0,0 +1,43 @@
...
+rules:
+  - apiGroups: [""]
+    resources:
+      - "secrets"
+    verbs: ["get", "update", "patch"]
+    resourceNames: [{{ template "newrelic-infra-operator.fullname" . }}-config]
+  - apiGroups: [""]
+    resources:
+      - "secrets"
+    verbs: ["create"]
```
invidian, line 16: Can we use `resourceNames` as well?

charts/newrelic-infra-operator/templates/clusterrole.yaml
```diff
@@ -0,0 +1,43 @@
...
+    resources:
+      - "nodes"
+      - "nodes/metrics"
+      - "nodes/stats"
+      - "nodes/proxy"
+      - "pods"
+      - "services"
+    verbs: ["get", "list"]
+  - nonResourceURLs: ["/metrics"]
+    verbs: ["get"]
```
invidian, line 43: Isn't this used for API monitoring which we are not doing right now?

charts/newrelic-infra-operator/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,31 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "newrelic-infra-operator.fullname" . }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "newrelic-infra-operator.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
+---
+{{/* infra-agent is the ClusterRoleBinding to be used by the ServiceAccoutns of the injected agents */}}
```
invidian, line 16:
```suggestion
{{/* infra-agent is the ClusterRoleBinding to be used by the ServiceAccounts of the injected agents */}}
```

charts/newrelic-infra-operator/values.yaml
```diff
@@ -0,0 +1,102 @@
...
+        # Sidecar will be injected if label value is truthy, will not if it's falsy.
+        # Upon conflict (one label is truthy but other is falsy), the falsy takes preference.
+        # If none of the labels in this list exists, `unlabeled` will be honored.
+        labels:
+          - newrelic.com/inject-sidecar
+#      # Monitor everything in example-namespace unless it has `newrelic.com/inject: false`
+#      example-namespace:
+#        unlabeled: true
+#        labels:
+#          - newrelic.com/inject-sidecar
```
invidian, line 97: To be honest, I'd go with more standard configuration, which would look like following:
```yaml
config:
  infraSidecarInjection:
    policies:
    # Inject to all pods in namespace "foo"
    - namespaceName: "foo"
    # Inject to all pods in namespace with label "foo=bar".
    - namespaceSelector:
        matchLabels:
          foo: bar
    # Inject to pods in all namespaces with label "foo=baz".
    - podSelector:
        matchLabels:
          foo: baz
```

Then `namespaceSelector` and `podSelector` fields will use [LabelSelector](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector), which allows to do any kind of selection (match, not match etc).

`namespaceName` is special for <1.21 clusters for matching the namespaces, which has no labels.

It will both simplify the implementation and will provide consistent API for users, similar to other Kubernetes resources.

charts/newrelic-infra-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,87 @@
...
+        - name: RELEASE_NAME
+          value: {{ template "newrelic-infra-operator.fullname" . }}
+        - name: NRIA_LICENSE_KEY
+          valueFrom:
+            secretKeyRef:
+              {{- if (include "newrelic-infra-operator.licenseKey" .) }}
+              name: {{ template "newrelic-infra-operator.fullname" . }}-config
+              key: license
+              {{- else }}
+              name: {{ include "newrelic-infra-operator.customSecretName" . }}
```
invidian, line 48: If both `customSecretName` and `licenseKey` are empty, we should throw an error while rendering.

charts/newrelic-infra-operator/templates/admission-webhooks/mutatingWebhookConfiguration.yaml
roobre: Removed for now, I don't think we want to support this use case in the short term.

charts/newrelic-infra-operator/templates/clusterrole.yaml
roobre: Sadly it doesn't seem to be possible for the `create` verb, as we discussed earlier :(

charts/newrelic-infra-operator/values.yaml
roobre: Agreed to go for the k8s standard, I didn't know this mechanism was as flexible. Does this miechanism support negative matches? I like the idea of being able to define a policy that achieves "monitor everything in this namespace EXCEPT this pods".

Not super critical, since you always can label everything else, but just wondering.

charts/newrelic-infra-operator/templates/clusterrole.yaml
roobre: I believe this allows us to fetch `/metrics/summary` from Kubelet, which we are still doing (although through the API server proxy). I'm not sure if this is still required if we do it from there, maybe @paologallinaharbur can provide some light on this.

charts/newrelic-infra-operator/templates/deployment.yaml
roobre: Previously, the approach that the charts followed were to have a helper checking for this, and surround every template with an `if` against that helper's output.

I'm not much of a fan of that approach but can't think about anything better right away, do you have any suggestion?

invidian: You can use `required` in conditional in the template, perhaps that would work? I'm not sure exactly what is the issue here.

charts/newrelic-infra-operator/values.yaml
invidian: You can use `notIn` operator for everything except as far as I understand: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements.

charts/newrelic-infra-operator/templates/deployment.yaml
roobre: I think it's a bit ugly since we're not using "one value or the other" in the same place.  What are your thoughts about:

```diff
diff --git a/charts/newrelic-infra-operator/templates/deployment.yaml b/charts/newrelic-infra-operator/templates/deployment.yaml
index 7c696ed..8f5f03a 100644
--- a/charts/newrelic-infra-operator/templates/deployment.yaml
+++ b/charts/newrelic-infra-operator/templates/deployment.yaml
@@ -44,9 +44,11 @@ spec:
               {{- if (include "newrelic-infra-operator.licenseKey" .) }}
               name: {{ template "newrelic-infra-operator.fullname" . }}-license
               key: license
-              {{- else }}
+              {{- else if include "newrelic-infra-operator.customSecretName" . }}
               name: {{ include "newrelic-infra-operator.customSecretName" . }}
               key: {{ include "newrelic-infra-operator.customSecretLicenseKey" . }}
+              {{- else }}
+              {{- nil | required "Cannot find License Key, either licenseKey or customSecretName must be defined" }}
               {{- end }}
         volumeMounts:
         - name: config
```

charts/newrelic-infra-operator/values.yaml
```diff
@@ -0,0 +1,96 @@
...
+#
+# Configure how long the API server should wait for a webhook to respond before treating the call as a failure
+#
+# Ref: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts
+timeoutSeconds: 30
+
+# Operator configuration
+# The following are the default values for the operator
+config:
+  # infraSidecarInjection: {}  # Sidecar injection can be disabled by setting this to an empty object
```
invidian, line 80: I'd rather set it to `nil`, so it is easy to disable the injection. Or actually, do we want to support disabling it? Shouldn't we disable whole operator deployment? Also` enabled: true/false` would be more standard than `infraSidecarInjection: null`.
```suggestion
  # infraSidecarInjection: null  # Sidecar injection can be disabled by setting this to an empty object
```

charts/newrelic-infra-operator/templates/deployment.yaml
invidian: Looks good to me. We could even put the whole license part into the template on it's own to make the `deployment.yaml` a bit simpler.

charts/newrelic-infra-operator/values.yaml
```diff
@@ -0,0 +1,96 @@
...
+# Webhook timeout
+#
+# Configure how long the API server should wait for a webhook to respond before treating the call as a failure
+#
+# Ref: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts
+timeoutSeconds: 30
+
+# Operator configuration
+# The following are the default values for the operator
+config:
```
invidian, line 79: The config for operator would look like this:
```yaml
# Operator configuration
# The following are the default values for the operator
config:
  # infraAgentInjection: ~  # Sidecar injection can be disabled by setting this to an empty object
  infraAgentInjection:
    resourcesPrefix: "foo"
    licenseSecret:
      name: "bar"
      key: "baz"
      value: "doh"
    infraAgentConfig:
      extraEnvVars:
        FOO: bar
      resources:
        requests:
          memory: "64Mi"
          cpu: "250m"
        limits:
          memory: "128Mi"
          cpu: "500m"
      image:
        repository: image
        tag: v0.1.0
        pullPolicy: Always
      podSecurityContext:
        runAsUser: 1000
        runAsGroup: 1000
    policies:
      # Inject to all pods in namespace "foo"
      - namespaceName: "foo"
      # Inject to all pods in namespace with label "foo=bar".
      - namespaceSelector:
          matchLabels:
            foo: bar
      # Inject to pods in all namespaces with label "foo=baz".
      - podSelector:
          matchLabels:
            foo: baz
```

charts/newrelic-infra-operator/values.yaml
roobre: I'm fine with the `enabled`, but since the config is passed "as is", it would be operator after unmarshalling the yaml the one deciding to not run.

invidian: I'd drop it for now, until we actually have some other functionality in the operator.

paologallinaharbur:
Before merging test and validate the upgrade path. I.e. make sure that the subject list is not wiped out

charts/newrelic-infra-operator/README.md
```diff
@@ -0,0 +1,37 @@
...
+## Resources
+
+The default set of resources assigned to the pods is shown below:
+
+    resources:
+      limits:
+        memory: 80M
+      requests:
+        cpu: 100m
+        memory: 30M
```
invidian, line 34:
```suggestion
    ```yaml
    resources:
      limits:
        memory: 80M
      requests:
        cpu: 100m
        memory: 30M
    ```
```

charts/newrelic-infra-operator/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Chart Details
+
+This chart will deploy the [New Relic Infrastructure Operator][1], which injects the New Relic Infrastructure solution as a sidecar to specific pods. This is typically used in environments where DaemonSets are not available, such as EKS Fargate.
+
+## Configuration
+
+| Parameter                     | Description                                                  | Default                    |
+| ----------------------------- | ------------------------------------------------------------ | -------------------------- |
+| `cluster`                     | The cluster name for the Kubernetes cluster.                 |                            |
```
invidian, line 11: We should document other parameters later.

invidian (approved):
Overall LGTM, we can fix the remaining stuff later.

charts/newrelic-infra-operator/Chart.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+description: A Helm chart to deploy the New Relic Infrastructure Kubernetes Operator.
+name: newrelic-infra-operator
+version: 0.1.0
+appVersion: 1.4.0
```
invidian, line 5: This still needs to be fixed.

roobre: I'll repoint to 0.1.0-pre since we have that released (although outdated)

charts/newrelic-infra-operator/templates/admission-webhooks/mutatingWebhookConfiguration.yaml
```diff
@@ -0,0 +1,31 @@
...
+    {{- include "newrelic-infra-operator.labels" . | nindent 4 }}
+webhooks:
+- name: newrelic-infra-operator.newrelic.com
+  clientConfig:
+    service:
+      name: {{ template "newrelic-infra-operator.fullname" . }}
+      namespace: {{ .Release.Namespace }}
+      path: "/mutate-v1-pod"
+{{- if not .Values.certManager.enabled }}
+    caBundle: ""
```
paologallinaharbur, line 20: @invidian  will we need at the end "reinvocationPolicy: IfNeeded"?

paologallinaharbur: I just saw you mentioned "we can fix the remaining stuff later." I guess you were referring to this as well

paologallinaharbur (approved):

charts/newrelic-infra-operator/templates/admission-webhooks/mutatingWebhookConfiguration.yaml
invidian: Yes, we should add `reinvocationPolicy: IfNeeded` I think.

invidian: Added.

paologallinaharbur:
Before merging I would wait for https://github.com/newrelic/newrelic-infra-operator/pull/40 and https://github.com/newrelic/newrelic-infra-operator/pull/50 to be merged and possibly for a new release

invidian:
>Before merging I would wait for newrelic/newrelic-infra-operator#40 and newrelic/newrelic-infra-operator#50 to be merged and possibly for a new release.

We should make the repository public as well :)

charts/newrelic-infra-operator/README.md
```diff
@@ -6,11 +6,53 @@ This chart will deploy the [New Relic Infrastructure Operator][1], which injects
 
 ## Configuration
 
+
 | Parameter                     | Description                                                  | Default                    |
 | ----------------------------- | ------------------------------------------------------------ | -------------------------- |
 | `cluster`                     | The cluster name for the Kubernetes cluster.                 |                            |
+| `image.repository`            | The container to pull.                                       | `newrelic/newrelic-infra-operator`   |
```
invidian, line 13: I'd fix indentation of table with http://markdowntable.com/ :)

charts/newrelic-infra-operator/values.yaml
```diff
@@ -90,19 +90,19 @@ config:
...
-      defaultValue: unknown
-      fromLabel: eks.amazonaws.com/fargate-profile
-
-## agentConfig contains all the 
-    agentConfig:
+      customAttributes:
+      - name: computeType
+        defaultValue: serverless
+      - name: fargateProfile
+        # To allow inclusion on non-fargate clusters for testing.
```
invidian, line 103: This shouldn't be there I guess? We could add those customAttributes only when `global.fargate=true`.

charts/newrelic-infra-operator/README.md
```diff
@@ -7,51 +7,52 @@ This chart will deploy the [New Relic Infrastructure Operator][1], which injects
...
+| `config.infraAgentInjection.policies[].namespaceSelector`                       | Selector on Namespace labels.                                                                                                                                                                                                                                                                                                               |                                                                                                      |
+| `config.infraAgentInjection.policies[].namespace`                               | Namespace name. if set only pods belonging to such namespace matches the policy.                                                                                                                                                                                                                                                             |                                                                                                      |
```
invidian, line 41:
```suggestion
| `config.infraAgentInjection.policies[].namespace`                               | Namespace name. If set only pods belonging to such namespace matches the policy.                                                                                                                                                                                                                                                             |                                                                                                      |
```

charts/newrelic-infra-operator/README.md
```diff
@@ -7,51 +7,52 @@ This chart will deploy the [New Relic Infrastructure Operator][1], which injects
...
+| `affinity`                                                                      | Node affinity to use for scheduling.                                                                                                                                                                                                                                                                                                         | `{}`                                                                                                 |
+| `config.ignoreMutationErrors`                                                   | if true it instruments the operator to ignore injection error instead of failing.                                                                                                                                                                                                                                                            | `true`                                                                                               |
```
invidian, line 37: I wonder if this is the right place to document the syntax of the configuration. Maybe it would be better to document it in operator docs instead?

charts/newrelic-infra-operator/README.md
```diff
@@ -7,51 +7,52 @@ This chart will deploy the [New Relic Infrastructure Operator][1], which injects
...
+| `config.infraAgentInjection.policies[].namespace`                               | Namespace name. if set only pods belonging to such namespace matches the policy.                                                                                                                                                                                                                                                             |                                                                                                      |
+| `config.infraAgentInjection.customAttributes[].name`                            | name of custom attribute to include.                                                                                                                                                                                                                                                                                                         |                                                                                                      |
```
invidian: And all descriptions below should be capitalized as well.
```suggestion
| `config.infraAgentInjection.customAttributes[].name`                            | Name of custom attribute to include.                                                                                                                                                                                                                                                                                                         |                                                                                                      |
```

charts/newrelic-infra-operator/README.md
```diff
@@ -7,51 +7,52 @@ This chart will deploy the [New Relic Infrastructure Operator][1], which injects
...
+| `config.infraAgentInjection.agentConfig.image.pullPolicy`                       | The sidecar image pull policy.                                                                                                                                                                                                                                                                                                              | `IfNotPresent`                                                                                       |
+| `config.infraAgentInjection.agentConfig.image.tag`                              | The infrastructure agent repository for the sidecar container.                                                                                                                                                                                                                                                                              | `2.4.0-unprivileged`                                                                                 |
```
invidian, line 47: Perhaps tag should be documented after repository? It makes sense from both logical point of view and from alphabetical order.
```suggestion
| `config.infraAgentInjection.agentConfig.image.tag`                              | The infrastructure agent image tag for the sidecar container.                                                                                                                                                                                                                                                                              | `2.4.0-unprivileged`                                                                                 |
```

charts/newrelic-infra-operator/README.md
```diff
@@ -7,51 +7,52 @@ This chart will deploy the [New Relic Infrastructure Operator][1], which injects
...
+| `config.infraAgentInjection.agentConfig.image.repository`                       | The infrastructure agent repository for the sidecar container.                                                                                                                                                                                                                                                                              | `newrelic/infrastructure-k8s`                                                                        |
+| `config.infraAgentInjection.agentConfig.podSecurityContext.RunAsUser`           | runAsUser UID for Pod Security Context.                                                                                                                                                                                                                                                                                                      |                                                                                                      |
```
invidian, line 49: I don't think field should be capitalized.
```suggestion
| `config.infraAgentInjection.agentConfig.podSecurityContext.runAsUser`           | runAsUser UID for Pod Security Context.                                                                                                                                                                                                                                                                                                      |                                                                                                      |
```

charts/newrelic-infra-operator/README.md
paologallinaharbur: 👍

paologallinaharbur:
@roobre You can not formally approve since you opened the PR, however can you take a look?

charts/newrelic-infra-operator/templates/_helpers.tpl
```diff
@@ -133,3 +133,16 @@ Return the customSecretLicenseKey
...
+*/}}
+{{- define "newrelic.fargate" -}}
+{{- if .Values.global }}
+  {{- if .Values.global.fargate }}
+    {{- .Values.global.fargate -}}
+  {{- end -}}
+{{- else if .Values.fargate }}
+  {{- .Values.fargate -}}
+{{- end -}}
+{{- end -}}
```
invidian, line 148: Missing newline.

charts/newrelic-infra-operator/templates/_helpers.tpl
```diff
@@ -133,3 +133,16 @@ Return the customSecretLicenseKey
     {{- .Values.customSecretLicenseKey | default "" -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Returns fargate
+*/}}
+{{- define "newrelic.fargate" -}}
+{{- if .Values.global }}
```
invidian, line 141: I'd love to simplify this, but it seems surprisingly complicated then :smile:

charts/newrelic-infra-operator/templates/configmap.yaml
```diff
@@ -1,8 +1,24 @@
...
+{{- if include "newrelic.fargate" . }}
+  agentConfig:
+{{- if not .Values.config.infraAgentInjection.agentConfig.customAttributes }}
+    customAttributes:
+      - name: computeType
+        defaultValue: serverless
+      - name: fargateProfile
+        fromLabel: eks.amazonaws.com/fargate-profile
+{{- end -}}
+{{- if not .Values.config.infraAgentInjection.agentConfig.policies }}
```
invidian: I think those conditions are not needed, merge should retain the values from config?

charts/newrelic-infra-operator/templates/configmap.yaml
```diff
@@ -1,8 +1,24 @@
 {{- define "fargate-config" -}}
 infraAgentInjection:
   resourcePrefix: {{ template "newrelic-infra-operator.fullname" . }}
+{{- if include "newrelic.fargate" . }}
+  agentConfig:
+{{- if not .Values.config.infraAgentInjection.agentConfig.customAttributes }}
+    customAttributes:
```
invidian: I wonder if we should have `default-config`, including `fargate-config` template, so setting `resourcePrefix` will be done indepedently from fargate settings. That would require 3 levels of merging though.

charts/newrelic-infra-operator/values.yaml
```diff
@@ -77,32 +77,32 @@ config:
...
 ## Within a policy PodSelectors, NamespaceSelector and NamespaceName are ANDed, any of these, if not specified, is ignored.
-    policies:
-    ## The following policy matches all pods belonging to any fargate profile
-    - podSelector:
-        matchExpressions:
-          - key: "label.eks.amazonaws.com/fargate-profile"
-            operator: Exists
-      # NamespaceName: "my-namespace"
-      # NamespaceSelector: LabelSelector{}
+## The following policy are injected if global.fargate=true and matches all pods belonging to any fargate profile
```
invidian, line 84:
```suggestion
## The following policy is injected if global.fargate=true and matches all pods belonging to any fargate profile.
```

charts/newrelic-infra-operator/values.yaml
```diff
@@ -77,32 +77,32 @@ config:
...
+#    policies:
+#    - podSelector:
+#        matchExpressions:
+#        - key: "label.eks.amazonaws.com/fargate-profile"
+#          operator: Exists
+## Also NamespaceName and NamespaceSelector can be leveraged
```
invidian, line 90:
```suggestion
## Also NamespaceName and NamespaceSelector can be leveraged.
```

charts/newrelic-infra-operator/templates/configmap.yaml
paologallinaharbur: I think it is safer and more explicit to show that if defined we are not trying to inject them

charts/newrelic-infra-operator/templates/_helpers.tpl
paologallinaharbur: I copied it from a different chart 😅

charts/newrelic-infra-operator/templates/configmap.yaml
invidian: Fair :+1:

charts/newrelic-infra-operator/templates/configmap.yaml
paologallinaharbur: I would avoid adding an other level if not strictly needed 😅

charts/newrelic-infra-operator/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+
+## Example
+
+Make sure you have [added the New Relic chart repository.](../../README.md#install)
+
+Then, to install this chart, run the following command:
+
+```sh
+helm upgrade --install newrelic/newrelic-infra-operator --set cluster=my_cluster_name
```
invidian, line 67: License is also required, right?

charts/newrelic-infra-operator/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Configure in which pods the sidecar should be injected
+
+Policies are available in order to configure in which pods the sidecar should be injected.
+Each policy is evaluated independently and if at least one policy matches the operator will inject the sidecar.
+
+Policies are composed by `namespaceSelector` checking the labels of the Pod namespace, `podSelector` checking
+the labels of the Pod and `namespace` checking the namespace name. Each of those, if specified, are ANDed.
+
+By default, the policies are configured in order to inject the sidecar in each pod belonging to a fargate profile.
```
invidian, line 78:
```suggestion
By default, the policies are configured in order to inject the sidecar in each pod belonging to a Fargate profile.
```

charts/newrelic-infra-operator/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+>Moreover, it is possible to add the label `infra-operator.newrelic.com/disable-injection` to Pods to exclude injection
+for single Pods that otherwise would be selected by the policies.
```
invidian, line 81:
```suggestion
for a single Pod that otherwise would be selected by the policies.
```

charts/newrelic-infra-operator/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+Please make sure to configure policies correctly avoiding injecting sidecar for pods running on EC2 nodes
```
invidian, line 83:
```suggestion
Please make sure to configure policies correctly to avoid injecting sidecar for pods running on EC2 nodes
```

charts/newrelic-infra-operator/README.md
```diff
@@ -0,0 +1,133 @@
...
+already monitored by the infrastructure DaemonSet.
+
+## Configure the sidecar with labelsSelectors
+
+It is also possible to configure `resourceRequirements` and `extraEnvVars` based on the labels of the mutating Pod.
+
+The current configuration increases the resource requirements for sidecar injected on `KSM` instances. Moreover, 
+injectes disable the `DISABLE_KUBE_STATE_METRICS` environmentVariable for pods not running on `KSM` instances
```
invidian, line 91:
```suggestion
injectes disable the `DISABLE_KUBE_STATE_METRICS` environment variable for Pods not running on `KSM` instances
```

charts/newrelic-infra-operator/README.md
```diff
@@ -0,0 +1,133 @@
...
+to decrease the load on the ApiServer.
```
invidian, line 92:
```suggestion
to decrease the load on the API server.
```

invidian (commented):
Just some nits in the readme, otherwise looks good!

charts/newrelic-infra-operator/README.md
paologallinaharbur: yes I am adding it, it was also missing from the readme

paologallinaharbur: as well as "cluster"

paologallinaharbur: also missing release name from the command

charts/newrelic-infra-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,87 @@
...
+    {{- include "newrelic-infra-operator.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicas }}
+  selector:
+    matchLabels:
+      {{- include "newrelic-infra-operator.appLabel" . | nindent 6 }}
+  template:
+    metadata:
+      {{- if .Values.podAnnotations }}
+      annotations:
```
invidian, line 15: I'd add https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments to automatically restart operator on configuration change.

paologallinaharbur: I agree, it is quite annoying

paologallinaharbur: @invidian added

charts/newrelic-infra-operator/values.yaml
```diff
@@ -0,0 +1,191 @@
...
+          limits:
+            memory: 100M
+            cpu: 200m
+          requests:
+            memory: 50M
+            cpu: 100m
+## extraEnvVars to pass to the injected sidecar.
+        extraEnvVars:
+          DISABLE_KUBE_STATE_METRICS: "true"
+#            NRIA_VERBOSE: "1"
```
invidian, line 134: Indentation could be fixed here, so when you remove `#`, then config is valid.

invidian (approved):
Tested on EKS Fargate cluster and everything works seamlessly.

PR #348, opened 2021-04-28
exclude daemonset from fargate nodes if global.fargate

### LINKED ISSUE #210: newrelic-infrastructure: needs affinity values

roobre:
#### Is this a new chart

No

Adds `global.fargate`, which will be used as a toggle to indicate whether the bundle is being deployed in an EKS fargate environment.

#### Checklist
- [x] Chart Version bumped

---

Fixes #210


paologallinaharbur:
can we add as well in this Pr the computeType envVar?

charts/newrelic-infrastructure/templates/daemonset.yaml
```diff
@@ -258,6 +258,17 @@ spec:
       {{- if $.Values.priorityClassName }}
       priorityClassName: {{ $.Values.priorityClassName }}
       {{- end }}
+      {{- if .Values.global.fargate }}
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: eks.amazonaws.com/compute-type
```
invidian, line 267: Rather than hardcoding it here, perhaps we can have some syntax sugar for exclude rules for DaemonSet? Or `nodeAffinity` even as a variable + `global.faraget` sugar?

roobre: Doesn't sound bad, let me give it a try

charts/newrelic-infrastructure/README.md
```diff
@@ -12,6 +12,7 @@ This chart will deploy the New Relic Infrastructure agent as a Daemonset.
...
 | `global.customSecretName` - `customSecretName` | Name of the Secret object where the license key is stored                                                                                                                                                                         |                                 |
 | `global.customSecretLicenseKey` - `customSecretLicenseKey` | Key in the Secret object where the license key is stored.                                                                                                                                                             |                                 |
+| `global.fargate`               | Must be set to `true` when deploying in an EKS Fargate environment. Prevents DaemonSet pods from being scheduled in Fargate nodes.  |                                 |
 | `config`                       | A `newrelic.yml` file if you wish to provide.                                                                                                                                                                                                     |                                 |
```
paologallinaharbur, line 16: are we doing something with this?

paologallinaharbur (approved):
LGTM

charts/newrelic-infrastructure/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic Infrastructure Agent as a DaemonSet
 name: newrelic-infrastructure
-version: 2.4.1
+version: 2.4.2
```
paologallinaharbur, line 4: actually isn't better 2.5.0?

invidian (approved):
I'd personally prefer to get an error in case of "conflicting" config, rather than warning, otherwise LGTM.

charts/newrelic-infrastructure/Chart.yaml
paologallinaharbur: Also because we are bumping nri-bundle to 2.9.0

invidian (commented):
>Agreed, but I'm not super sure on how to reliably detect whether the user has specified this already, so I'm going for the easy approach. You're welcome to give this a go :)

I'm not sure if this is actually a question on how to implement it, but it can be done for example like this:
```yaml
{{- if (and (eq (len .Values.nodeAffinity) 1) .Values.global.fargate) }}
{{- required "global.fargate and nodeAffinity values are mutually exclusive" "" }}
{{- end }}
```

However, If other charts also use `global.fargate`, it may not be convenient for the user to unset `global.fargate` only because of single chart. So either we implement some sort of indirection, which will allow unsetting `global.fargate` only for this chart or we actually rely only on the warning.

charts/newrelic-infrastructure/templates/daemonset.yaml
```diff
@@ -6,8 +6,7 @@ metadata:
   labels: {{ include "newrelic.labels" . | indent 4 }}
   name: {{ template "newrelic.fullname" . }}
   {{- if .Values.daemonSet.annotations }}
-  annotations:
-{{ toYaml .Values.daemonSet.annotations | indent 4 }}
+  annotations: {{ toYaml .Values.daemonSet.annotations | nindent 4 }}
```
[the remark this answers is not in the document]
roobre: I think just the opposite, `indent` makes things a bit confusing to read, and is easier to make indentation mistakes 😅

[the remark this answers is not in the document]
invidian:
```yaml
annotations: {{ toYaml .Values.daemonSet.annotations | nindent 4 }}
labels: {{ toYaml .Values.daemonSet.labels | nindent 4 }}
keys: {{ toYaml .Values.daemonSet.keys | nindent 4 }}
annotations: {{ toYaml .Values.daemonSet.annotations | nindent 4 }}
labels: {{ toYaml .Values.daemonSet.labels | nindent 4 }}
keys: {{ toYaml .Values.daemonSet.keys | nindent 4 }}
```

vs

```yaml
annotations:
  {{ toYaml .Values.daemonSet.annotations | nindent 4 }}
labels:
  {{ toYaml .Values.daemonSet.labels | nindent 4 }}
keys:
  {{ toYaml .Values.daemonSet.keys | nindent 4 }}
annotations:
  {{ toYaml .Values.daemonSet.annotations | nindent 4 }}
labels:
  {{ toYaml .Values.daemonSet.labels | nindent 4 }}
keys:
  {{ toYaml .Values.daemonSet.keys | nindent 4 }}
```

Does it make more sense when there is more keys perhaps? IMO former starts getting blurry with all the lines, it becomes a single blob of text while the latter has clear distinction between each element.

PR #359, opened 2021-05-05
[nri-prometheus] Adding config options to scrape endpoints

paologallinaharbur:
#### Is this a new chart
no

Adds the new  config options included into https://github.com/newrelic/nri-prometheus/pull/148 updating as well the image

Updated as well documentation https://github.com/newrelic/docs-website/pull/2126

#### Checklist
- [x] Chart Version bumped


charts/nri-prometheus/README.md
```diff
@@ -42,3 +42,17 @@ helm install newrelic/nri-prometheus \
...
+since `scrape_services` is set to `true` and `scrape_endpoints` to `false`. 
+
+In order to change this behaviour set `scrape_endpoints` to `true` configuring `nri-promtheus` to 
+scrape the underlying endpoints, as Prometheus server natively does, instead of directly the services.
+
+Please notice that depending on the number of endpoints behind the services in the cluster the load and the data injested can increase considerably, 
+monitor and, if needed, increase resource requirements.
+
+Moreover, even if it is possible to set both `scrape_services` and `scrape_endpoints` to true to assure retrocompatibility, 
+it would lead to duplicate data.
```
roobre, line 58:
```suggestion
## Scraping services and endpoints

When a service is labeled or anotated with `scrape_enabled_label` (defaults to `prometheus.io/scrape`),
`nri-prometheus` will attempt to hit the service directly, rather than the endpoints behind it.

This is the default behavior for compatibility reasons, but is known to cause issues if more than one endpoint
is behind the service, as metric queries wil be load-balanced as well leading to unaccurate histograms.

In order to change this behaviour set `scrape_endpoints` to `true` and `scrape_services` to `false`.
This will instruct `nri-promtheus` to scrape the underlying endpoints, as Prometheus server does.

Existing users that are switching to this behavior should note that, depending on the number of endpoints
behind the services in the cluster the load and the metrics reported by those, data ingestion might see
an increase when flipping this option. Resource requirements might also be impacted, again depending on the number of new targets.

While it is technically possible to set both `scrape_services` and `scrape_endpoints` to true, we do no recommend
doing so as it will lead to redundant metrics being processed,
```

roobre (approved):

PR #369, opened 2021-05-14
[newrelic-infrastructure] fix NPD in .Values.global

roobre:
Despite being a convention, .Values.global is not automatically initialized to anything. When trying to render this chart alone, attempting to check for truthyness of .Values.global.fargate caused a NPD.
This commit adds a helper similar to newrelic.nrStaging taking care of checking if .Values.global is defined before attempting to deference it.

#### Is this a new chart

Nope

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure/templates/_helpers.tpl
+++ b/charts/newrelic-infrastructure/templates/_helpers.tpl
@@ -140,6 +140,19 @@ Returns nrStaging
 {{- end -}}
 
 {{/*
+Returns fargate
+*/}}
+{{- define "newrelic.fargate" -}}
+{{- if .Values.global }}
+  {{- if .Values.global.fargate }}
+    {{- .Values.global.fargate -}}
+  {{- end -}}
+{{- else if .Values.fargate }}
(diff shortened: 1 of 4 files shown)


charts/newrelic-infrastructure/templates/_helpers.tpl
```diff
@@ -139,6 +139,19 @@ Returns nrStaging
...
+{{/*
+Returns fargate
+*/}}
+{{- define "newrelic.fargate" -}}
+{{- if .Values.global }}
+  {{- if .Values.global.fargate }}
+    {{- .Values.global.fargate -}}
+  {{- end -}}
+{{- else if .Values.fargate }}
+  {{- .Values.nrStaging -}}
```
paologallinaharbur, line 151: I am a bit confused why .Values.nrStaging?

roobre: Mistake on my part, good catch!

paologallinaharbur (approved):

PR #376, opened 2021-05-19
[nri-bundle] deploy ksm without prometheusScrape labels

roobre:
#### Is this a new chart

No

By default, the Kubernetes wizard will deploy the New Relic solution with `nri-prometheus` turned on, and configured to scrape anything in the cluster with the "prometheus label". Since KSM is also deployed with this label enabled, `nri-prometheus` will attempt to scrape it as well, leading to unnecessary ingest of metrics as KSM is already parsed by `nri-kubernetes`.

This PR disables the prometheus scrape flag of KSM.

I'm still testing this, although no wrong side-effects are expected.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/nri-bundle/Chart.yaml
+++ b/charts/nri-bundle/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.0"
 description: A Helm chart to deploy New Relic integrations bundled together
 name: nri-bundle
-version: 2.11.1
+version: 2.12.0
 home: https://github.com/newrelic/helm-charts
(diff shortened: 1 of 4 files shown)


roobre:
Marking as draft for now

paologallinaharbur:
@roobre We could follow this approach to avoid any breaking change https://github.com/newrelic/helm-charts/pull/436/files

Fred was trying to standardize chart to use `lowDataMode` FF, we could do the same

roobre:
I would have preferred for `lowDataMode` to trigger this rather than filtering metrics after scraping them, but unfortunately that doesn't seem to be possible with Helm :(
Since the use case is solved I'll be closing this one for now :)

PR #391, opened 2021-06-07
ci/bump_chart: use bot token to open PR with

roobre:
#### Is this a new chart

No

Previously, version-bumping PRs were created with the auto-generated GH token, which caused said PRs to not run their workflows due to a limitation from GH.

This PR allows the create PR action to run with a different userl which should circumvent this limitation.

  - fixes #

#### Checklist
- [ ] Chart Version bumped


roobre:
This approach would supply a privileged, non-volatile token to a third party action. Since this poses a security risk, we will need to follow a different approach.

PR #412, opened 2021-06-18
synthetics-minion/kmullaney

keegoid-nr:
#### Checklist


keegoid-nr:
I'm having trouble getting the minion to be scheduled to run on a node and pass the Lint and Test Charts check. The node says it is not ready.

paologallinaharbur:
It is a timeout due to an old version of the kindCluster action, we are about to merge a fix for that.

The following in the workflow should fix the issue
```
uses: helm/kind-action@v1.2.0
```

paologallinaharbur:
I updated your branch to make use of the updated action, now it is not failing 👍

charts/synthetics-minion/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: synthetics-minion
 description: New Relic Synthetics Containerized Private Minion (CPM)
 type: application
-version: 1.0.34
+version: 1.0.35
```
cleon-nr, line 5: We'll need to bump this to `1.0.36`.

cleon-nr (approved):
@keegoid-nr

Thanks for these changes. They look good to me!

Since some of these updates are related to the top level `README`, this PR will also require approval from the owner of that directory as well.

roobre (approved):
Changes to `CONTRIBUTING.md` lgtm :)

charts/synthetics-minion/Chart.yaml
cleon-nr: Strangely it's showing that the last version is `1.0.34` from this diff, but we already have `1.0.35` merged in. Perhaps your fork just needs to be synced up again.

keegoid-nr: bumped version

keegoid-nr: the versions keep changing faster than this pr can get merged

Issue #404, opened 2021-06-14
Rewrite chart-bump-action


roobre:
**Is your feature request related to a problem? Please describe.**

Our current solution for automatically bumping charts is a custom action written in Node. This action is hard to maintain and very often gets security warnings, given the high number of dependencies.

Additionally, since this workflow also uses a third party action for creating pull requests, we cannot supply it a custom token to allow it to run CI checks due to security concerns.

**Describe the solution you'd like**

It would be great to rewrite the chart-bump action in go, and automate PR creation with the `gh` tool rather than a third-party action.

kang-makes:
This is now done by the renovate bot.

PR #413, opened 2021-06-18
Reimplement version bumper

roobre:
#### Is this a new chart

No

Deprecates the old chart-version-bumper action with a new version written in go with less dependencies.

Fix https://github.com/newrelic/helm-charts/issues/404


kang-makes:
Related to #404 which is closed.

Closing.

PR #420, opened 2021-06-30
[CI/CD] bumping version of kind cluster and increasing timeout

paologallinaharbur:
#### Is this a new chart
no
The ci/cd sometimes fails since the kind cluster is not ready yet


paologallinaharbur:
Actually bumping the action is enough to speed it up and therefore I would keep it in the same PR

PR #421, opened 2021-07-02
[newrelic-infrastructure] Change update strategy values to add more flexibility

### LINKED ISSUE #410: [newrelic-infrastructure] Enable configuring updateStrategy better for the daemonset
In big environments rolling update of the daemonset will be fery slow.
It would be nice to be able to set other values, not just updateStrategy type

See [this example](https://github.com/aws/aws-node-termination-handler/blob/main/config/helm/aws-node-termination-handler/templates/daemonset.linux.yaml#L11) for reference.

gdziwoki:
#### Is this a new chart
No
resolves https://github.com/newrelic/helm-charts/issues/410
adds ability to set all parameters for updateStrategy
  - fixes #410 

Not sure how to bump chart version, so I didn't
#### Checklist
- [x] Chart Version bumped


paologallinaharbur:
I fear this would represent a breaking change  since `updateStrategy` passes from being a string to be a dictionary.

However, I totally understand that the naming of UpdateStrategy was already really unfortunate.
I wonder if it is better going with a small breaking change or "deprecate it" and add a new  "updateStrategyDictionary" or directly `maxUnavailable`

By the way you added the windows support since you needed it in your use-case? I am asking since I imagine it being useful only in a mixed windows-linux scenario where the two daemonset needs different updateStrategy

gdziwoki:
Using `updateStrategy` as dictionary is more future-proof, if there would be any other parameters added. Yes it's a breaking change in that sense, that's why I waasn't sure about the chart versioning.
We don't do mixed envs, but it seemed like a good practice.

paologallinaharbur:
I agree, I will raise it with the team. 
Bumping the major version for such change could be too much, but strictly speaking it is what we should do.
We might end-up hold it for now waiting for other breaking changes and pack them together

paologallinaharbur:
Ok, I spoke with the team, basically since it is a fix of a something that it was broken and misleading a major version change is not needed and we can move forward.
On the other hand, we would get rid of the two different configuration options for Linux and Windows, the use case does not seem to be strong enough to add such complexity

gdziwoki:
Removing two different versions for Windows and Linux is fine by me.
Should I bump the chart version?

roobre:
@gdziwoki Yes please, otherwise the CI will complain :)
As @paologallinaharbur said, I think a minor version bump would be good here, since even if this can break configurations, I think it's more a bug that anything else.

charts/newrelic-infrastructure/README.md
```diff
@@ -39,7 +39,7 @@ This chart will deploy the New Relic Infrastructure agent as a Daemonset.
...
 | `tolerations`                  | List of node taints to tolerate (requires Kubernetes >= 1.6)                                                                                                                                                                                      | See Tolerations below           |
-| `updateStrategy`               | Strategy for DaemonSet updates (requires Kubernetes >= 1.6)                                                                                                                                                                                       | `RollingUpdate`                 |
+| `updateStrategy`               | Update strategy for the all DaemonSets (Linux and Windows (requires Kubernetes >= 1.6)                                                                                                                                                                                       | `type=RollingUpdate,rollingUpdate.maxUnavailable=1`                 |
```
roobre, line 42:
```suggestion
| `updateStrategy`               | Update strategy the DaemonSets (Linux and Windows) (requires Kubernetes >= 1.6)                                                                                                                                                                                       | `type=RollingUpdate,rollingUpdate.maxUnavailable=1`                 |
```

roobre (commented):
Regardin this comment:

> requires Kubernetes >= 1.6

is this correct, meaning `1.6` and not `1.16`? If this is the case (meaning `1.6`) I would drop it since we do not plan support for versions older than 1.15 in the near future.

gdziwoki:
@roobre @paologallinaharbur anything else I need to do to get this merged?
looks like you released 2.4.8 in the meantime

paologallinaharbur (approved):
Thanks again for the contribution and the fixes

PR #426, opened 2021-07-07
[newrelic-pixie] Add support for talking to Pixie's staging

douglascamata:
#### Is this a new chart

No.

When the chart is configured to send data to New Relic's staging, it will talk to Pixie's staging environment.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-pixie/templates/deployment.yaml
+++ b/charts/newrelic-pixie/templates/deployment.yaml
@@ -62,6 +62,10 @@ spec:
         - name: NR_OTLP_HOST
           value: "staging.otlp.nr-data.net:4317"
         {{- end }}
+        {{- if (include "newrelic-pixie.pixieStaging" .) }}
+        - name: PIXIE_ENDPOINT
+          value: "staging.withpixie.dev:443"
+        {{- end }}
         {{- if .Values.resources }}
         resources:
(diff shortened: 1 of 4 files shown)


gsanchezgavier (approved):
lgtm

charts/newrelic-pixie/templates/deployment.yaml
```diff
@@ -61,6 +61,10 @@ spec:
         {{- if (include "newrelic-pixie.nrStaging" .) }}
         - name: NR_OTLP_HOST
           value: "staging.otlp.nr-data.net:4317"
+        {{- end }}
+          {{- if (include "newrelic-pixie.nrStaging" .) }}
```
roobre, line 65: Seems like there are two extra spaces here

roobre:
Looks like `nrStaging` will trigger both the use of pixie staging and NR staging. Is this the intended behavior? I wonder if there would be an use case for using staging for one thing and not the other.

charts/newrelic-pixie/templates/deployment.yaml
douglascamata: Thanks IntelliJ!

roobre (approved):

PR #436, opened 2021-07-17
[nri-prometheus] Pomi config update for Guided Install metric exclusion

bpschmitt:
#### Is this a new chart

No

Adds a configuration option for KSM and cAdvisor metric exclusion when using the Guided Install.  These metrics are collected by the Infra Daemonset by default.

N/A

Default setting is currently `guidedInstall: false` so that we don't change the behavior for anyone using the chart outside of the Guided Install.

#### Checklist
- [X] Chart Version bumped


charts/nri-prometheus/values.yaml
```diff
@@ -170,71 +175,95 @@ config:
...
-  #         to_metrics: "kube_node_"
-  #         match_by:
-  #           - namespace
-  #           - node
+  #
   # integration definition files required to map metrics to entities
   # definition_files_path: /etc/newrelic-infra/definition-files
 
+custom_transformations:
+  transformations: {}
```
roobre, line 183: Should be initialized to `[]` if it's an array
```suggestion
  transformations: []
```

charts/nri-prometheus/values.yaml
```diff
@@ -170,71 +175,95 @@ config:
...
-  #       - from_metric: "kube_node_labels"
-  #         to_metrics: "kube_node_"
-  #         match_by:
-  #           - namespace
-  #           - node
+  #
   # integration definition files required to map metrics to entities
   # definition_files_path: /etc/newrelic-infra/definition-files
 
+custom_transformations:
```
roobre, line 182: Since we use CamelCase for values, I think this would be better in that style (even if pomi uses sneak case internally).

charts/nri-prometheus/values.yaml
```diff
@@ -170,71 +175,95 @@ config:
...
+    #       match_by:
+    #         - namespace
+    #         - service
+    #     - from_metric: "kube_node_labels"
+    #       to_metrics: "kube_node_"
+    #       match_by:
+    #         - namespace
+    #         - node
+
+guided_install_transformations:
```
roobre, line 245: Same as above, I'd switch to camelCase for this key

charts/nri-prometheus/values.yaml
```diff
@@ -20,6 +20,11 @@
 
 prometheusScrape: "true"
 
+# This parameter enables the default metric set
+# ingested by nri-prometheus when using the Guided Install.
+# Default: false
+guidedInstall: false
```
roobre, line 26:
```suggestion
# This parameter will be set to true if installing from guided install.
# If set, the `config` entry present in this file will be ignored,
# and instead a default one with a different metric set configured will be used.
# Default: false
guidedInstall: false
```

charts/nri-prometheus/templates/configmap.yaml
```diff
@@ -11,3 +11,8 @@ data:
 {{- if .Values.config }}
 {{ toYaml .Values.config | indent 4 -}}
 {{ end }}
+{{- if .Values.guidedInstall }}
+{{ toYaml .Values.guided_install_transformations | indent 4 -}}
+{{ else }}
+{{ toYaml .Values.custom_transformations | indent 4 -}}
+{{ end }}
```
roobre, line 18: I'm not much of a fan of the idea of merging keys in the yaml this way. For example, it would be possible for an unknowledgeable user to add a `transformations` section to `.Values.config`, causing a yaml syntax error.

Since with the guided install it's very uncommon to have a custom values.yml file, what do you think about creating a different template, containing a whole config file for guided install, and leaving this as it is?

Something like:

```suggestion
{{- if .Values.guidedInstall }}
{{ include "guidedInstallConfig" . | indent 4}}
{{- else if .Values.config }}
{{ toYaml .Values.config | indent 4 -}}
{{- end }}
```

This way it would be very hard to accidentally mix configurations, and the default guided install config file could be on a separate template rather than in the values. IMO this would be better since it's not intended to be modified/overridden by the user.

charts/nri-prometheus/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 2.7.0
 description: A Helm chart to deploy the New Relic Prometheus OpenMetrics integration
 name: nri-prometheus
-version: 1.8.2
+version: 1.8.3
```
roobre, line 5: This will change some behavior so I'd definitely bump a minor

roobre (changes requested):

charts/nri-prometheus/values.yaml
bpschmitt: Updated.

charts/nri-prometheus/values.yaml
bpschmitt: Removed this in the latest iteration.

charts/nri-prometheus/values.yaml
bpschmitt: Remoted this in the latest iteration.

charts/nri-prometheus/Chart.yaml
bpschmitt: Bumped chart to `1.9.0`

charts/nri-prometheus/templates/configmap.yaml
bpschmitt: @roobre I made some modifications to this in the latest update.  

- Created a `static` directory which contains the full YAML for the Guided Install config
- If `guidedInstall` is `true` in `values.yaml` then `static/guidedinstalldefaults.yaml` will be loaded into the POMI `config.yaml` ConfigMap
- If `guidedInstall is `false` in `values.yaml` then the `config:` section will be used.

Overall, I think this will be cleaner and we can easily edit the Guided Install defaults if we decide to change them in the static file.

charts/nri-prometheus/templates/configmap.yaml
```diff
@@ -8,6 +8,10 @@ apiVersion: v1
...
-{{- if .Values.config }}
-{{ toYaml .Values.config | indent 4 -}}
-{{ end }}
+{{- if .Values.guidedInstall }}
+    {{- .Files.Get "static/guidedinstalldefaults.yaml" | indent 4 -}}
+{{ else }}
+    {{- if .Values.config }}
+    {{ toYaml .Values.config | indent 4 | trim  -}}
+    {{ end }}
+{{ end }}
```
roobre, line 17: It should be possible to compact else and if
```suggestion
{{- else if .Values.config }}
    {{ toYaml .Values.config | indent 4 | trim }}
{{- end }}

```

charts/nri-prometheus/static/guidedinstalldefaults.yaml
```diff
@@ -0,0 +1,134 @@
...
+          pod: "podName"
+          deployment: "deploymentName"
+    ignore_metrics:
+      # Ignore the following metrics.
+      # These metrics are already collected by the New Relic Kubernetes Integration.
+      - prefixes:
+        - kube_
+        - container_
+        - machine_
+        - cadvisor_
```
roobre, line 134: Missing newline before EOF

```suggestion
        - cadvisor_

```

roobre (commented):
Overall looking pretty good, thanks a lot @bpschmitt 
Just left two minor comments :)

bpschmitt:
@roobre are you waiting on me for anything here?

paologallinaharbur (approved):

paologallinaharbur:
I actually realised that the new injected config is static, not only for the transformation section. Is it the expected behaviour?

paologallinaharbur:
closing in favour or #518

PR #438, opened 2021-07-20
[nri-kube-events] Use the NRIA_IS_FORWARD_ONLY to avoid entity collision.

gsanchezgavier:
#### Is this a new chart
no
Replace the secure forward mode for the forward only mode to avoid entity creation and collision. 
  - fixes https://github.com/newrelic/nri-kube-events/issues/56

Staging test [link](https://staging-one.newrelic.com/-/0nVjYPGbXR0)
Also [check](https://staging-one.newrelic.com/-/01qwL5LXgR5) that cluster explorer still has the same functionality regarding the events and linkage to the entities that they belong. 
#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/nri-kube-events/templates/deployment.yaml
+++ b/charts/nri-kube-events/templates/deployment.yaml
@@ -75,6 +75,11 @@ spec:
             - name: NRIA_PROXY
               value: "{{ .Values.proxy }}"
             {{- end }}
+            # Using FORWARD_ONLY mode to avoid the entity creation for the events.
+            - name: NRIA_IS_SECURE_FORWARD_ONLY
+              value: "false"
+            - name: NRIA_IS_FORWARD_ONLY
+              value: "true"
           volumeMounts:
             - mountPath: /var/db/newrelic-infra/data
               name: tmpfs-data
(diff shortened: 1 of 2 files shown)


charts/nri-kube-events/templates/deployment.yaml
```diff
@@ -75,6 +75,11 @@ spec:
             - name: NRIA_PROXY
               value: "{{ .Values.proxy }}"
             {{- end }}
+            # Using FORWARD_ONLY mode to avoid the entity creation for the events.
+            - name: NRIA_IS_SECURE_FORWARD_ONLY
+              value: "false"
```
roobre, line 80: Is it better to set this to `false`, or empty? Maybe @newrelic/caos can provide some insight.

roobre (approved):
LGTM!

charts/nri-kube-events/templates/deployment.yaml
gsanchezgavier: my first approach was to set it to empty but i saw an error log on the agent sayin that was not able to convert it to boolean type, so i followed the `false`.

gsanchezgavier:
```
level=error msg="unable to interpret environment variables" component=Configuration error="envconfig.Process: assigning NRIA_IS_SECURE_FORWARD_ONLY to IsSecureForwardOnly: converting '' to type bool. details: strconv.ParseBool: parsing \"\": invalid syntax"
```

PR #443, opened 2021-07-26
[synthetics-minion] Add an option to exclude jobs/log resource from role

thilinapiy:
#### Is this a new chart: 
No

Default behaviour - no changes
If `removeJobsLog` is set to `true`, it will remove following api resource from the `role`.

There is no API resource called `jobs/log` in the main Kubernetes API and it is causing an installation failure.     

  - fixes # N/A

#### Checklist
- [X] Chart Version bumped


cleon-nr (approved):
Thanks for your contribution! This looks good to me.

PR #448, opened 2021-07-30
add proxy variable to env vars

cherepsv:
proxy option exists in values, but there is no any usage of it.
tested with this env var - now looks better

  - fixes #

#### Checklist
- [ ] Chart Version bumped


paologallinaharbur:
Hello, we decided to reduce as much as possible the number of variable exposed. Moreover there are different types of proxies supported by the agent and each one of them would require a different env vars.

The proxy you see in the values is mounted as a config map directly in  /etc/newrelic-infra.yml of the container

https://docs.newrelic.com/docs/integrations/kubernetes-integration/installation/configure-kubernetes-proxy/

paologallinaharbur:
closing for now, feel free to reopen if the approach does not work for you

PR #457, opened 2021-08-04
newrelic-infrastructure: make updateStrategy backwards-compatible

roobre:
#### Is this a new chart

No

Fixes a regression introduced in #421 which caused upgrade to fail if `--reuse-values` were specified in the CLI.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure/templates/_helpers.tpl
+++ b/charts/newrelic-infrastructure/templates/_helpers.tpl
@@ -155,14 +155,17 @@ Returns fargate
 {{/*
 Returns the updateStrategy, either .Values.updateStrategy directly if it is an object, or wrapped if it is a string
 This is done to keep compatibility with old values and --reuse-values.
+Defining updateStrategy as a string is deprecated and will be removed in a future version of the chart.
 */}}
 {{- define "newrelic.updateStrategy" -}}
 {{- if .Values.updateStrategy }}
 {{- if eq "string" (printf "%T" .Values.updateStrategy) }}
 updateStrategy:
   type: {{ .Values.updateStrategy }}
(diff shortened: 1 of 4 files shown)


charts/newrelic-infrastructure/templates/_helpers.tpl
```diff
@@ -152,6 +152,24 @@ Returns fargate
...
+{{/*
+Returns the updateStrategy, either .Values.updateStrategy directly if it is an object, or wrapped if it is a string
+This is done to keep compatibility with old values and --reuse-values.
+*/}}
+{{- define "newrelic.updateStrategy" -}}
+{{- if .Values.updateStrategy }}
+{{- if eq "string" (printf "%T" .Values.updateStrategy) }}
+updateStrategy:
+  type: {{ .Values.updateStrategy }}
+  rollingUpdate:
```
gsanchezgavier, line 164: isn't needed to check if the update strategy was rollingUpdate?

roobre: Theoretically yes, but this is as it was hardcoded before 🤔 
I can do the check here though, just to be in the safe side.

douglascamata (approved):

charts/newrelic-infrastructure/templates/_helpers.tpl
gsanchezgavier: oh i didn't know that, otherwise I would just complete the `type` and nothing else thinking that this would be using the default values for rollingUpdate and recreate

gsanchezgavier: if that was hardcoded befor , i don't think that needs the check

gsanchezgavier (approved):

PR #459, opened 2021-08-06
[newrelic-pixie] Add proxy configuration for Pixie integration.

fryckbos:
#### Is this a new chart
No

It adds a proxy configuration field to the Pixie integration Helm chart.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-pixie/templates/deployment.yaml
+++ b/charts/newrelic-pixie/templates/deployment.yaml
@@ -66,6 +66,12 @@ spec:
         - name: PIXIE_ENDPOINT
           value: "staging.withpixie.dev:443"
         {{- end }}
+        {{- if .Values.proxy }}
+        - name: HTTP_PROXY
+          value: "{{ .Values.proxy }}"
+        - name: HTTPS_PROXY
+          value: "{{ .Values.proxy }}"
+        {{- end }}
         {{- if .Values.resources }}
         resources:
 {{ toYaml .Values.resources | indent 10 }}
--- a/charts/newrelic-pixie/templates/deployment.yaml
+++ b/charts/newrelic-pixie/templates/deployment.yaml
@@ -68,9 +68,9 @@ spec:
         {{- end }}
         {{- if .Values.proxy }}
         - name: HTTP_PROXY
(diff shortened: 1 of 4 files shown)


charts/newrelic-pixie/values.yaml
```diff
@@ -28,3 +28,6 @@ resources:
   requests:
     cpu: 100m
     memory: 250M
+
+proxy: {}
```
roobre, line 32: Should this be an object (`{}`) or an string? From the `deployment.yaml` template I understand it should be a string, right? Or are we planning to split http and https proxies?

charts/newrelic-pixie/templates/deployment.yaml
```diff
@@ -66,6 +66,12 @@ spec:
         - name: PIXIE_ENDPOINT
           value: "staging.withpixie.dev:443"
         {{- end }}
+        {{- if .Values.proxy }}
+        - name: HTTP_PROXY
+          value: "{{ .Values.proxy }}"
+        - name: HTTPS_PROXY
+          value: "{{ .Values.proxy }}"
+        {{- end }}
```
roobre, line 74: I would sugges to use `quote` since it will escape inner quotes rather than producing potentially incorrect yaml.

```suggestion
        {{- if .Values.proxy }}
        - name: HTTP_PROXY
          value: {{ .Values.proxy | quote }}
        - name: HTTPS_PROXY
          value: {{ .Values.proxy | quote }}
        {{- end }}
```

charts/newrelic-pixie/values.yaml
fryckbos: According to the docs, both an empty string and empty object should work: https://helm.sh/docs/chart_template_guide/control_structures/

roobre: We've seen helm complaining when switching types, for example overriding an empty object with an array, but in this case it doesn't seem to matter 😕 

In any case, if Helm is fine with it, so I am :)

fryckbos:
Fixed the linter issue.

PR #462, opened 2021-08-13
Synthetics minion/kmullaney/initcontainer resources

keegoid-nr:
#### Is this a new chart
no

Statefulset failed to create minion pod due to failed quota missing resource limits

```log
Warning  FailedCreate      6s (x12 over 16s)  statefulset-controller  create Pod synthetics-minion-0 in StatefulSet synthetics-minion failed error: pods "synthetics-minion-0" is forbidden: failed quota: compute-resources: must specify limits.cpu,limits.memory,requests.cpu,requests.memory
```

The `initContainer` in our Helm chart as of chart version 1.0.35 does not specify resource limits for memory or cpu. If a resource quota is defined for the namespace, for example

```sh
kubectl get quota -n YOUR_NAMESPACE
```

The statefulset will fail to create the minion pod due to the missing resource limits.

The solution is to specify the resource limits in the statefulset for the `initContainer`:

```yaml
      initContainers:
        - name: "update-mounted-subpath-permissions"
          resources:
            requests:
              cpu: "100m"
              memory: "100Mi"
            limits:
              cpu: "200m"
              memory: "200Mi"
```

#### Checklist
- [x] Chart Version bumped


charts/synthetics-minion/templates/statefulset.yaml
```diff
@@ -93,6 +93,13 @@ spec:
               value: {{ include "synthetics-minion.user-defined-variables-path" . | quote }}
           volumeMounts:
           {{- include "synthetics-minion.volumeMounts" . | nindent 12 }}
+          resources:
+            requests:
+              cpu: "100m"
+              memory: "100Mi"
+            limits:
+              cpu: "200m"
+              memory: "200Mi"
```
keegoid-nr, line 102: These resources need to be here if a namespace quota is in use.

keegoid-nr: I entered what should be reasonable values, but feel free to adjust up if needed.

charts/synthetics-minion/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: synthetics-minion
 description: New Relic Synthetics Containerized Private Minion (CPM)
 type: application
-version: 1.0.38
+version: 1.0.40
```
cleon-nr, line 5: The current version is 1.0.40 so we'll want to bump this to `1.0.41`. The `appVersion` can remain the same though

```suggestion
version: 1.0.41
```

cleon-nr (approved):
LGTM. Thank you!

PR #465, opened 2021-08-16
[synthetics-minion] upgrading authorization API from v1beta1 to v1

kaschaefer:
#### Is this a new chart
No
This PR updates which Authorization API the synthetics-minion uses to v1 from v1beta1.
Fixes customer issues where they can't build the CPM with K8s v1.22

Thank you!

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/synthetics-minion/Chart.yaml
+++ b/charts/synthetics-minion/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: synthetics-minion
 description: New Relic Synthetics Containerized Private Minion (CPM)
 type: application
-version: 1.0.36
+version: 1.0.38
 appVersion: 3.0.48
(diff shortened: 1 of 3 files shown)


roobre:
Hi @kaschaefer,
Thanks a lot for this PR!
We are aware of our charts not being compatible with 1.22 right now, and we are working towards making a coordinated release for all of them to include support for 1.22. Unfortunately this is a nontrivial change, since it involves breaking compatibility with some earlier version that, according to our data, are still in use by a non-negligible amount of users.

We are following up on this and we expect to have a solution soon for our userbase that is already moving to 1.22 :)

In the meantime, I'll be closing this since it will be superseded by a larger PR with the coordinated bump across all charts.

PR #467, opened 2021-08-18
nri-bundle: replace pixie chart with pixie operator chart

roobre:
## Extra steps

After this update, the following manifests need to be applied _before_ installing the helm chart:

- https://github.com/pixie-labs/pixie/blob/main/k8s/operator/crd/base/px.dev_viziers.yaml
- https://github.com/pixie-labs/pixie/blob/main/k8s/operator/helm/crds/olm_crd.yaml

The previously installed CRDs, https://raw.githubusercontent.com/pixie-labs/pixie/main/k8s/vizier_deps/base/nats/nats_crd.yaml, should no longer be needed.


charts/nri-bundle/requirements.lock
```diff
@@ -16,15 +16,15 @@ dependencies:
...
 - name: newrelic-pixie
   repository: file://../newrelic-pixie
-  version: 1.1.0
-- name: pixie-chart-helm2
-  repository: https://pixie-chart-helm2.storage.googleapis.com
-  version: 0.8.2
+  version: 1.1.1
+- name: pixie-operator-helm2-chart
+  repository: https://pixie-operator-charts.storage.googleapis.com
+  version: 0.0.4
```
aimichelle, line 25: Can we actually update this to 0.0.5? We just did an operator release today, so it would be good to pull in that latest version. Thanks!

invidian (commented):
Looks good, though I'd squash all commits together, as they are sort of related.

charts/nri-bundle/requirements.lock
roobre: Done in https://github.com/newrelic/helm-charts/pull/467/commits/17604c77b57c50d542055e1136cf20b7f07056ee :)

invidian:
Can we enable CI for those 2 subcharts to make sure they install fine (right now they won't due to missing CRDs) and that they converge?

invidian:
Note: this will still not work on Kubernetes 1.22 at least because of https://github.com/nats-io/nats-operator/issues/327

PR #494, opened 2021-09-08
[charts/newrelic-infrastructure] Split KSM scraping into Deployment

invidian:
Includes removal of pixie for temporary K8s 1.22 support.

Related to https://github.com/newrelic/nri-kubernetes/issues/201

### DIFF
--- a/charts/newrelic-infrastructure/templates/deployment.yaml
+++ b/charts/newrelic-infrastructure/templates/deployment.yaml
@@ -1,15 +1,14 @@
 {{- if and (include "newrelic.areValuesValid" .) .Values.enableLinux }}
 apiVersion: apps/v1
-kind: DaemonSet
+kind: Deployment
 metadata:
   namespace: {{ .Release.Namespace }}
   labels: {{ include "newrelic.labels" . | indent 4 }}
-  name: {{ template "newrelic.fullname" . }}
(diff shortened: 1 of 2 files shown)


charts/newrelic-infrastructure/templates/deployment.yaml
```diff
@@ -0,0 +1,241 @@
...
+        - name: nri-integrations-cfg-volume
+          configMap:
+            name: {{ template "newrelic.fullname" . }}-integrations-cfg
+        {{- end }}
+      {{- if $.Values.priorityClassName }}
+      priorityClassName: {{ $.Values.priorityClassName }}
+      {{- end }}
+      affinity:
+        podAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
```
roobre, line 223: Definitely out of scope for the PoC but I'd add the reciprocal label to KSM, so the affinity is honored if the KSM pod gets killed.

invidian: True. I mean, this is only to make it work on cluster with multiple nodes in best case scenario :smile: But we can consider adding those affinities in general.

roobre:
Superseded by https://github.com/newrelic/helm-charts/pull/624

PR #503, opened 2021-09-09
nri-bundle: clarify how to configure dependencies and the agent

roobre:
#### Is this a new chart

No

Configuring the infrastructure agent through Helm is often not clear enough for some users. This PR adds more verbosity to the config examples to hopefully be more clear in how to achieve this.


gsanchezgavier (approved):
thanks! don't forget to bump the version

PR #505, opened 2021-09-09
ci: prevent concurrent releases

roobre:
Previously, merging two PRs to `master` at the same time (or very close in time) might cause two `Release` workflows to trigger in parallel. Since the release is done by pushing to the `gh-pages` branch, which is shared, a write-after-read race might occur, resulting in one of the releases failing or overwriting the other one.

This PR sets the `concurrency` key of the release workflow to instruct GHA to allow only one parallel run of said workflow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - master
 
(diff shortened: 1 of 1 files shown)


douglascamata (approved):
Yes, please :shipit:

PR #170, opened 2020-10-30
Windows FluentBit Helm Chart

andrew-lozoya:
Since v1.5.0, Fluent Bit supports deployment to Windows pods. New Relic offers a [Fluent Bit](https://fluentbit.io/) output [plugin](https://github.com/newrelic/newrelic-fluent-bit-output) to easily forward your logs to [New Relic Logs](https://docs.newrelic.com/docs/logs/new-relic-logs/get-started/introduction-new-relic-logs). This plugin is also provided in a [standalone Docker image](https://github.com/andrew-lozoya/newrelic-fluent-bit-output/blob/master/Dockerfile) that can be installed in a [Kubernetes](https://kubernetes.io/) cluster in the form of a [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/), which we refer as the Kubernetes plugin.

This document explains how to install it in your cluster, either using a [Helm](https://helm.sh/) chart (recommended). Starting with chart version 2.7.1, chart deploys the daemonset on windows nodes which collects std{out;err} logs of the containers running on windows nodes.

PR #509, opened 2021-09-16
[newrelic-logging] Add support for Windows containers

jsubirat:
#### Is this a new chart
No.

The PR adds support to ship logs from Windows containers. This PR includes the changes contributed by @andrew-lozoya in this PR: https://github.com/newrelic/helm-charts/pull/170, but simplified so that we only need to install one Helm chart both for Windows and Linux.

- The main introduced change is the `daemonset-windows.yml` file. Some observations:
  - It creates **multiple** `DaemonSet`s, because we need a separate `DaemonSet` for each Windows version that we need to support. Right now, we only support Windows Server 2019 LTSC, therefore it will only create one `DaemonSet`.
  - It uses the same version of the `newrelic-fluent-bit-output` Docker image as the Linux Helm chart, but note that the image tag contains a suffix as the [Linux](https://hub.docker.com/layers/newrelic/newrelic-fluentbit-output/1.9.2/images/sha256-cf37c73e985875a65516317bb736737d71ca735602a19fdbdcfe68e9e8caddd5?context=explore) and [Windows](https://hub.docker.com/layers/newrelic/newrelic-fluentbit-output/1.9.2-windows-ltsc-2019/images/sha256-488c71004c6a5618c83e042bfac8214d48c4e0a830d55410712b43de3384e3b6?context=explore) images are different.
  - We ensure that the `DaemonSet` will only be deployed to nodes having the `kubernetes.io/os=windows` and `node.kubernetes.io/windows-build=SOME_BUILD` by using a `nodeSelector`. Note that we need to constraint this because **Windows containers can only be executed on hosts running the exact same Windows version and build number**. This limitation justifies many of the changes in this PR.
- We needed to un-split the `ConfigMap` into a single `fluent-bit.conf` file, because Fluent Bit running on a Windows container does not support the `@INCLUDE` directive. This way, the same `ConfigMap` can be reused across the two `DaemonSets` (Linux and Windows).
- As a result, we've had to add the `FB_DB` environment variable in order to provide different DB paths for Windows and Linux. Reminder: this is the DB where Fluent Bit will store up to which offset it has read from each log file. This file needs to be mounted _outside_ of any container (in the _host_ filesystem) in order to avoid losing this data across container redeployments.
- We now provide flags to enable/disable entirely the support for Linux and/or Windows. This is specially targeted for those users only having Linux machines that don't need to create the Windows `DaemonSet`.

#### Checklist
- [x] Chart Version bumped


charts/newrelic-logging/templates/daemonset-windows.yaml
```diff
@@ -0,0 +1,139 @@
...
+            - C:\fluent-bit\bin\out_newrelic.dll
+          volumeMounts:
+            - mountPath: C:\fluent-bit\etc
+              name: fluent-bit-config
+            - mountPath: C:\var\log
+              name: varlog
+              # We need to also mount this because the logs in C:\var\logs are actually symlinks to C:\ProgramData.
+              # So, in order to be able to read these logs, the reading process needs to also have access to C:\ProgramData.
+            - mountPath: C:\ProgramData
+              name: progdata
```
danybmx, line 108: 👍🏻 Nice finding!

danybmx (changes requested):
Hey @jsubirat and @andrew-lozoya nice work! kudos!! 👏🏻 👏🏻 

I've just left a comment on values file for `enableWindows` that I think should be changed, if not, ping me and I'll approve this.

Thanks both for take care of windows containers!! 🪟🪟

charts/newrelic-logging/templates/configmap.yaml
```diff
@@ -41,10 +35,9 @@ data:
     [FILTER]
         Name           kubernetes
         Match          kube.*
-        Kube_URL       https://kubernetes.default.svc:443
+        Kube_URL       https://kubernetes.default.svc.cluster.local:443
```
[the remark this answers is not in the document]
jsubirat: It took me a while to figure that out! Actually, I'm going to put this as a comment, thanks Dani!

charts/newrelic-logging/values.yaml
```diff
@@ -42,6 +45,21 @@ image:
   # pullSecrets:
     # - name: regsecret
 
+# By default, the Linux DaemonSet will always be deployed, while the Windows DaemonSet(s) won't.
+enableLinux: true
+enableWindows: true
```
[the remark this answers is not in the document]
jsubirat: Indeed, this is a leftover from my tests, I will be reverting this. Thanks!

charts/newrelic-logging/templates/daemonset-windows.yaml
jsubirat: @andrew-lozoya 👏

danybmx (approved):
kudos!! 🏅🏅

charts/newrelic-logging/templates/daemonset-windows.yaml
```diff
@@ -0,0 +1,139 @@
...
+{{ toYaml $.Values.image.pullSecrets | indent 8 }}
+      {{- end }}
+      containers:
+        - name: {{ template "newrelic-logging.name" $ }}
+          image: "{{ $.Values.image.repository }}:{{ $.Values.image.tag | default $.Chart.AppVersion  }}-{{ .imageTagSuffix }}"
+          imagePullPolicy: "{{ $.Values.image.pullPolicy }}"
+          securityContext: {}
+          env:
+            - name: ENDPOINT
+              {{- if (include "newrelic.nrStaging" $) }}
```
tejunior, line 53: Do we need to expose staging? Maybe we can expose only US and EU

tejunior (approved):

charts/newrelic-logging/templates/daemonset-windows.yaml
jsubirat: We are also doing this in the Linux daemonset, we can handle this, if needed, in a separate PR.

charts/newrelic-logging/templates/daemonset.yaml
```diff
@@ -122,10 +125,16 @@ spec:
                     values:
                       - fargate
       {{- end }}
-      {{- if $.Values.nodeSelector }}
       nodeSelector:
+        {{- if $.Values.nodeSelector }}
```
paologallinaharbur, line 129: I do not think that $ is needed here since we are not inside a loop

paologallinaharbur: also in the if next to it is the same

jsubirat: Indeed, fixed!

PR #510, opened 2021-09-17
newrelic-infrastructure: fix use of updateStrategy in daemonset-windows

roobre:
#### Is this a new chart

No

Fixes an error where the `updateStrategy` helper was not being called with the right argument in the daemonset-windows template.

#### Checklist
- [x] Chart Version bumped


invidian (approved):
OK

PR #511, opened 2021-09-17
Add metrics adapter chart

invidian:


charts/newrelic-k8s-metrics-adapter/Chart.yaml
```diff
@@ -0,0 +1,20 @@
...
+name: newrelic-k8s-metrics-adapter
+version: 0.1.0
+appVersion: 0.1.0
+home: https://hub.docker.com/r/newrelic/newrelic-k8s-metrics-adapter
+sources:
+  - https://github.com/newrelic/newrelic-k8s-metrics-adapter
+engine: gotpl
+icon: https://newrelic.com/assets/newrelic/source/NewRelic-logo-square.svg
+maintainers:
+  - name: douglascamata
```
paologallinaharbur, line 12: davidbrota and douglascamata are no longer part of the team

charts/newrelic-k8s-metrics-adapter/values.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+personalAPIKey:
+
+accountID:
+externalMetrics:
+#  test1:
+#    query: "SELECT latest(cpuUsedCores) from K8sContainerSample "
+#    addClusterFilter: true
+
+# If not set, it will be automatically derived from global.licenseKey.
```
paologallinaharbur, line 23: does it work as well with the not global licenseKey?

charts/newrelic-k8s-metrics-adapter/templates/configmap.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "newrelic-k8s-metrics-adapter.fullname" . }}
+  labels:
+    {{- include "newrelic-k8s-metrics-adapter.labels" . | nindent 4 }}
+data:
+  config.yaml: |
+    accountID: {{ .Values.accountID | required "accountID is required" }}
```
paologallinaharbur, line 9: are we including them separately to perform checks?

charts/newrelic-k8s-metrics-adapter/templates/_helpers.tpl
```diff
@@ -0,0 +1,104 @@
...
+{{- end -}}
+
+{{/*
+Select a value for the region
+*/}}
+{{- define "newrelic-k8s-metrics-adapter.region" -}}
+{{- if .Values.region -}}
+{{ .Values.region | upper }}
+{{- else -}}
+{{ include "newrelic-k8s-metrics-adapter.licenseKey" . | substr 0 2 | upper }}
```
paologallinaharbur, line 102: I believe this is true only for EU licenses, no eu ones do not have us prefix

charts/newrelic-k8s-metrics-adapter/values.yaml
```diff
@@ -0,0 +1,89 @@
...
+# can be shared by other newrelic product's charts.
+#
+# global:
+#   licenseKey:
+#   cluster:
+#   fargate
+
+personalAPIKey:
+
+accountID:
```
paologallinaharbur, line 17: isn't better to include all config maps options under config even if we access them one by one?

paologallinaharbur (changes requested):
the eu logic do not work for us licenses

charts/newrelic-k8s-metrics-adapter/values.yaml
gsanchezgavier: Maybe we can add a note like this
```suggestion
# If not set, it will be automatically derived from global.licenseKey.
# For US-based accounts, the region is: `US`.
# For EU-based accounts, the region is: `EU`.
# For Staging accounts, the region is: 'Staging' this is also automatically derived form `global.nrStaging`
```

charts/newrelic-k8s-metrics-adapter/templates/apiservice/apiservice.yaml
```diff
@@ -0,0 +1,22 @@
...
+  annotations:
+    certmanager.k8s.io/inject-ca-from: {{ printf "%s/%s-root-cert" .Release.Namespace (include "newrelic-k8s-metrics-adapter.fullname" .) | quote }}
+    cert-manager.io/inject-ca-from: {{ printf "%s/%s-root-cert" .Release.Namespace (include "newrelic-k8s-metrics-adapter.fullname" .) | quote }}
+{{- end }}
+spec:
+  service:
+    name: {{ template "newrelic-k8s-metrics-adapter.fullname" . }}
+    namespace: {{ .Release.Namespace }}
+  group: external.metrics.k8s.io
+  version: v1beta1
```
gsanchezgavier, line 17: shouldn't we add the `insecureSkipTLSVerify` configurable parameter ?

charts/newrelic-k8s-metrics-adapter/Chart.yaml
invidian: Updated.

charts/newrelic-k8s-metrics-adapter/values.yaml
invidian: Right, we can wrap it with `config: {}`, I don't mind.

charts/newrelic-k8s-metrics-adapter/templates/configmap.yaml
invidian: Also to dynamically select a value for region.

charts/newrelic-k8s-metrics-adapter/values.yaml
invidian: It should work fine with non global license key, as it make use of the common template:
```
{{/*
Return the licenseKey
*/}}
{{- define "newrelic-k8s-metrics-adapter.licenseKey" -}}
{{- if .Values.global}}
  {{- if .Values.global.licenseKey }}
      {{- .Values.global.licenseKey -}}
  {{- else -}}
      {{- .Values.licenseKey | default "" -}}
  {{- end -}}
{{- else -}}
    {{- .Values.licenseKey | default "" -}}
{{- end -}}
{{- end -}}
```

charts/newrelic-k8s-metrics-adapter/templates/apiservice/apiservice.yaml
invidian: I'm not sure. We either generate the certificate using hook OR delegate that to cert-manager, so TLS can always be enabled.

invidian:
CI is finally passing :tada:

gsanchezgavier (approved):
lgtm

invidian:
Before we merge, we may want to revisit https://github.com/newrelic/helm-charts/pull/511/commits/9741be60fef79f7d3e9c5e0d0fb6abfa9df2da20 and perhaps squash commits.

gsanchezgavier:
We could make PR to this branch as we talked and merge it when all is ready

gsanchezgavier:
just rebased from master to get 1.22 changes

charts/newrelic-k8s-metrics-adapter/ci/test-values.yaml
```diff
@@ -0,0 +1,38 @@
...
+          - key: test-key
+            operator: In
+            values:
+            - test-value
+tolerations:
+- key: "key1"
+  operator: "Exists"
+  effect: "NoSchedule"
+
+image:
```
paologallinaharbur, line 36: we should remember to update this

PR #517, opened 2021-09-22
[newrelic-pixie] Introducing low data mode & bump to v1.4.0

fryckbos:
#### Is this a new chart
No

* Bump the Pixie integration version to 1.4.0
* Introduces low data mode

* Communication to Pixie now uses end-to-end encryption
* OTLP protobuf updated to v0.9.0

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-pixie/Chart.yaml
+++ b/charts/newrelic-pixie/Chart.yaml
@@ -9,9 +9,11 @@ sources:
 engine: gotpl
 icon: https://newrelic.com/assets/newrelic/source/NewRelic-logo-square.svg
 maintainers:
-  - name: paologallinaharbur
-  - name: davidbrota
+  - name: alvarocabanas
+  - name: carlossscastro
   - name: gsanchezgavier
+  - name: kang-makes
+  - name: paologallinaharbur
   - name: roobre
 keywords:
   - newrelic
--- a/charts/newrelic-pixie/Chart.yaml
+++ b/charts/newrelic-pixie/Chart.yaml
@@ -1,15 +1,18 @@
 apiVersion: v1
 description: A Helm chart for the New Relic Pixie integration.
 name: newrelic-pixie
-version: 1.3.2
-appVersion: 1.3.0
+version: 1.4.0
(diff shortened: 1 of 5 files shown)


charts/newrelic-pixie/Chart.yaml
```diff
@@ -1,15 +1,18 @@
...
+version: 1.4.0
+appVersion: 1.4.0
 home: https://hub.docker.com/u/newrelic
 sources:
   - https://github.com/newrelic/
 engine: gotpl
 icon: https://newrelic.com/assets/newrelic/source/NewRelic-logo-square.svg
 maintainers:
-  - name: douglascamata
+  - name: paologallinaharbur
```
paologallinaharbur, line 12: The list changed https://github.com/newrelic/helm-charts/blob/master/charts/nri-bundle/Chart.yaml#L8

charts/newrelic-pixie/templates/deployment.yaml
```diff
@@ -58,6 +58,14 @@ spec:
         - name: VERBOSE
           value: "true"
         {{- end }}
+        {{- if (include "newrelic-pixie.lowDataMode" .) }}
+        - name: COLLECT_INTERVAL_SEC
```
paologallinaharbur, line 62: Does it makes sense to always expose those env vars and add the default to be those if lowDataMode=true?
Or we prefer not to provide that flexibility? No idea if it can create issues

charts/newrelic-pixie/Chart.yaml
paologallinaharbur: 😢

fryckbos: Updated.

charts/newrelic-pixie/templates/deployment.yaml
fryckbos: These are certainly variables we can expose in the Helm chart if there is demand of that. For now I would keep it simple.

paologallinaharbur (approved):

PR #518, opened 2021-09-22
[nri-prometheus] Introducing low data mode

fryckbos:
#### Is this a new chart
No

This PR is based on the PR from Brad (https://github.com/newrelic/helm-charts/pull/436) but standardises on the lowDataMode attribute to enable this mode.

#### Checklist
- [x] Chart Version bumped


charts/nri-prometheus/templates/configmap.yaml
```diff
@@ -8,6 +8,8 @@ apiVersion: v1
 data:
   config.yaml: |
     cluster_name: {{ include "nri-prometheus.cluster" . }}
-{{- if .Values.config }}
-{{ toYaml .Values.config | indent 4 -}}
-{{ end }}
+{{- if (include "nri-prometheus.lowDataMode" .) }}
+    {{- .Files.Get "static/lowdatamodedefaults.yaml" | indent 4 -}}
```
paologallinaharbur, line 12: If `lowDataMode=true` how a customer could change any of the injected parameters?

Do we need to inject a whole new static config or only to enforce:
``` yaml
transformations:
  - description: "Low data mode defaults"
    ignore_metrics:
      # Ignore the following metrics.
      # These metrics are already collected by the New Relic Kubernetes Integration.
      - prefixes:
        - kube_
        - container_
        - machine_
        - cadvisor_
```

charts/nri-prometheus/static/lowdatamodedefaults.yaml
```diff
@@ -0,0 +1,124 @@
...
+# Default: "5m"
+# telemetry_emitter_delta_expiration_age: "5m"
+
+# How often must the telemetry emitter check for expired delta entries.
+# Default: "5m"
+# telemetry_emitter_delta_expiration_check_interval: "5m"
+
+# Whether the integration should run in verbose mode or not.
+# Default: false
+verbose: false
```
paologallinaharbur, line 35: If I am not wrong customers setting lowData=true will not be able to change these defaults parameters, is it the expected behaviour?

charts/nri-prometheus/templates/configmap.yaml
paologallinaharbur: I modified the PR to overwrite only the transformation list

charts/nri-prometheus/static/lowdatamodedefaults.yaml
paologallinaharbur: I modified the PR to overwrite only the transformation list

charts/nri-prometheus/README.md
```diff
@@ -12,6 +12,7 @@ This chart will deploy the New Relic's Prometheus OpenMetrics Integration.
...
 | `global.customSecretName` - `customSecretName`             | Name of the Secret object where the license key is stored                                                                                                                                                                             |                                        |
 | `global.customSecretLicenseKey` - `customSecretLicenseKey` | Key in the Secret object where the license key is stored.                                                                                                                                                                             |                                        |
+| `global.lowDataMode` - `lowDataMode` | Excludes KSM and cAdvisor metrics when enabled overwriting config.transformations                                                                                                                                                                                                      |           `false`                      |
```
fryckbos, line 15: I think this needs to be a bit more clear. What do you think about:

Excludes KSM and cAdvisor metrics when enabled. The low data mode overwrites `config.transformations` and `config.require_scrape_enabled_label_for_nodes`, if you want to set these fields in the config you need to turn low data mode off.

paologallinaharbur: I agree, I am removing `require_scrape_enabled_label_for_nodes` only because it is true by default and there is no reason to fix it in the static file

paologallinaharbur (approved):
LGTM

gsanchezgavier (approved):

PR #525, opened 2021-09-27
nri-bundle: update newrelic-logging to its latest version

roobre:
#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[mychartname]`)

PR #527, opened 2021-09-28
nri-bundle: Add Kubernetes 1.22 support

roobre:
#### ⚠️ Breaking changes ⚠️ 

This PR *introduces Kubernetes 1.22.x* support for the New Relic Kuberentes solution, and *removes support for kubernetes 1.15.x and below*.

Users may still use previous versions of the `nri-bundle` and other charts to instrument their clusters, and limited support will be offered for them. Development for said older versions is considered stopped and New Relic does guarantee feature parity or bugfixes (security or otherwise) for those older versions.

#### Is this a new chart

No

Bumps versions in individual charts and the bundle to point to the latest released versions of all Kubernetes components, which include support for Kubernetes 1.22.

Additionally, some changes to the charts themselves are also included to make them compatible with said version.

#### Checklist
- [x] Chart Version bumped

Includes changes from #468

### DIFF
--- a/charts/newrelic-infrastructure/Chart.yaml
+++ b/charts/newrelic-infrastructure/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic Infrastructure Agent as a DaemonSet
 name: newrelic-infrastructure
-version: 2.6.4
-appVersion: 2.7.1
+version: 2.7.0
+appVersion: 2.8.1
 home: https://hub.docker.com/r/newrelic/infrastructure-k8s/
 sources:
   - https://github.com/newrelic/helm-charts/tree/master/charts/newrelic-infrastructure
--- a/charts/nri-bundle/Chart.yaml
+++ b/charts/nri-bundle/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.0"
 description: A Helm chart to deploy New Relic integrations bundled together
 name: nri-bundle
-version: 2.23.0
+version: 3.0.0
 home: https://github.com/newrelic/helm-charts
 icon: https://newrelic.com/assets/newrelic/source/NewRelic-logo-square.svg
 maintainers:
--- a/charts/nri-metadata-injection/Chart.yaml
+++ b/charts/nri-metadata-injection/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic metadata injection webhook.
 name: nri-metadata-injection
-version: 1.5.4
-appVersion: 1.5.0
+version: 2.0.0
+appVersion: 1.6.0
 home: https://hub.docker.com/r/newrelic/k8s-metadata-injection
 sources:
   - https://github.com/newrelic/k8s-metadata-injection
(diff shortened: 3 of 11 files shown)


paologallinaharbur:
Same nri-bundle version of https://github.com/newrelic/helm-charts/pull/525/files

charts/newrelic-infrastructure/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic Infrastructure Agent as a DaemonSet
 name: newrelic-infrastructure
-version: 2.6.4
-appVersion: 2.7.1
+version: 2.7.0
```
gsanchezgavier, line 4: no major bump? just to double ensure the breaking change communication

charts/nri-bundle/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.0"
 description: A Helm chart to deploy New Relic integrations bundled together
 name: nri-bundle
-version: 2.23.0
+version: 2.24.0
```
gsanchezgavier, line 5: no major bump? just to double ensure the breaking change communication

charts/nri-metadata-injection/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic metadata injection webhook.
 name: nri-metadata-injection
-version: 1.5.4
-appVersion: 1.5.0
+version: 1.6.0
```
gsanchezgavier, line 4: should bump the major because of the breaking change?

roobre: Good point, done!

charts/nri-bundle/Chart.yaml
roobre: Bumping major here as well

roobre:
> Same nri-bundle version of #525

@paologallinaharbur Yes, this PR supersedes #525 and the latter will be closed after merging this.

charts/nri-metadata-injection/values.yaml
```diff
@@ -16,8 +16,8 @@ image:
     # - name: regsecret
 
 jobImage:
-  repository: jettech/kube-webhook-certgen
-  tag: v1.5.0
+  repository: k8s.gcr.io/ingress-nginx/kube-webhook-certgen
```
paologallinaharbur, line 19: this was used also in the newrelic-infra operator, has it been fixed there already?

paologallinaharbur: https://github.com/newrelic/helm-charts/blob/65447698c9798a0b1ead680788de3a2804885f92/charts/newrelic-infra-operator/values.yaml#L26

roobre: Afaik the PR @invidian made to support signing other objects is still pending, but this one should work for 1.22.

paologallinaharbur (approved):

charts/nri-metadata-injection/values.yaml
invidian: Yeah, this one works for 1.22.

PR #533, opened 2021-09-30
[metric-adapter-chart] add verbose logs

gsanchezgavier:


invidian (approved):
`-v=10` seems quite verbose, but OK :+1:

PR #551, opened 2021-10-20
[newrelic-logging] Made fluent-bit.conf and parsers.conf ConfigMap configurable from values.yaml file

bpschmitt:
#### Is this a new chart

No

#### What this PR does / why we need it:

The changes in this PR provide users the ability to customize the fluent-bit.conf and parsers.conf files using the values.yaml file.  Previously, this was statically set in the configmap.yaml file.  This will provide much better flexibility for custom fluent-bit configurations.

#### Which issue this PR fixes

N/A

#### Special notes for your reviewer:

This closely mimics the [fluent-bit helm chart](https://github.com/fluent/helm-charts/blob/main/charts/fluent-bit/templates/configmap.yaml)

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [X] Chart Version bumped
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[mychartname]`)

PR #565, opened 2021-11-01
[newrelic-logging] Parameterise launch command

### LINKED ISSUE #564: [newrelic-logging] Allow command to be overridden

I would like to be able to override the command used to launch Fluent Bit so that I can pass alternative launch commands.  That would allow me to pass additional command line arguments to launch additional plugins or launch with a completely different command.

Having the command defined in the values.yaml 

**Additional context**

We would like to launch Fluent Bit with [Fluent Bit Watcher](https://github.com/fluent/fluentbit-operator/blob/5694f2ac3fbf76525630bc363f79882937b71720/cmd/fluent-bit-watcher/Dockerfile).  This allows changes to the ConfigMap to trigger a restart of Fluent Bit and get the updated config.

Our specific requirement is to enable integration with existing internal tooling which we use to manage Kubernetes namespace quotas.  This will allow us to set throttling quotas using [Fluent Bit's throttling plugin](https://docs.fluentbit.io/manual/pipeline/filters/throttle) and automatically reload Fluent Bit when the throttling config changes.  This specific use case is probably quite niche, but imagine it could be more generally useful to pass different command line flags.

neilfordyce:
Allow launch command overrides to alter default launch command.

#### Is this a new chart
No

Makes it possible to override the launch command and pass different command line flags.  Some possible reasons described in #564.

  - fixes #564

If this seems worthwhile, should we do this for the windows daemonset too?

#### Checklist
- [x] Chart Version bumped


danybmx:
Hi @neilfordyce,

Our target with these helm-charts is more about common cases and this is not a really common one (as you mentioned). Also, our image doesn't include fluent-bit-watcher, and neither bash/sh... or any other command at the moment of writing this comment.

For plugin configurations and so on, there is another PR open https://github.com/newrelic/helm-charts/pull/551 that is under review.

So sadly, we're going to close this PR by now.

Regards and thanks a lot for your contribution.

PR #567, opened 2021-11-03
[newrelic-logging] Fixes config map to not exclude message when running in lowDataMode (docker & cri)

jodstrcil:

When docker is used log messages are wrapped in the log attribute
When CRI is used log messages are wrapped in the message attribute

We need both to in the Allowlist_key

  - fixes Low Data Mode to allow logs. 

#### Checklist
- [x] Chart Version bumped


jsubirat (approved):
Thanks for fixing this, Jo!

PR #572, opened 2021-11-08
Add override hostname short if unprivileged

alvarocabanas:
When on unprivileged mode, add the variable NRIA_OVERRIDE_HOSTNAME_SHORT with the nodeName in order to make the lookup of the FDQN with that instead of the pod name.


paologallinaharbur (approved):

charts/newrelic-infra-operator/values.yaml
```diff
@@ -113,7 +113,7 @@ config:
 ## Image of the infrastructure agent to be injected.
       image:
         repository: newrelic/infrastructure-k8s
-        tag: 2.8.1-unprivileged
```
paologallinaharbur, line 116: can you bump as well the info in the docs https://github.com/newrelic/helm-charts/blob/b01b6fe6cb622377f3dda61aca102cc13a87e0dc/charts/newrelic-infra-operator/README.md?plain=1#L53

charts/nri-bundle/requirements.yaml
```diff
@@ -48,4 +48,4 @@ dependencies:
   - name: newrelic-infra-operator
     repository: file://../newrelic-infra-operator
     condition: newrelic-infra-operator.enabled
-    version: 0.3.0
```
paologallinaharbur, line 51: We usually do not bound together nri-bundle with its dependencies, in the future it is better to open different Prs 😄

paologallinaharbur (approved):

charts/newrelic-infra-operator/values.yaml
alvarocabanas: done

Issue #110, opened 2020-08-26
[nri-kube-events] nri-kube-events chart should support custom attributes


sjammy6:
nri-kube-events helm chart should support custom_sttributes just like the rest of the product family (nri-infrastructure). This feature is needed for us to monitor and set up alerts based on the custom_attributes

Reference  code for implementation:
https://github.com/newrelic/helm-charts/blob/master/charts/newrelic-infrastructure/values.yaml#L91
https://github.com/newrelic/helm-charts/blob/master/charts/newrelic-infrastructure/templates/configmap.yaml#L10
https://github.com/newrelic/helm-charts/blob/master/charts/newrelic-infrastructure/templates/daemonset.yaml#L142

bpschmitt:
@sjammy6 I've opened a PR with the updates you requested.

https://github.com/newrelic/helm-charts/pull/588

davidgit:
Duplicate of #598

PR #598, opened 2021-11-19
[nri-kube-events] Add custom attributes support through the agent

kang-makes:
#### Is this a new chart

No

Custom Attributes are needed for the Kubernetes rearchitecture so a patch to send Custom Attributes in every metric has been added to the agent: https://github.com/newrelic/infrastructure-agent/pull/825

This PR adds custom attributes to kube-events which currently is not supported.

https://github.com/newrelic/helm-charts/issues/110

#### Checklist
- [x] Chart Version bumped


kang-makes:
Force push to rebase to the last commit of `master`.

Blocked by newrelic/infrastructure-agent#911

charts/nri-kube-events/templates/_helpers.tpl
```diff
@@ -141,3 +141,31 @@ licenseKey and cluster are set.
 {{- $customSecretLicenseKey := include "nri-kube-events.customSecretLicenseKey" . -}}
 {{- and (or $licenseKey (and $customSecretName $customSecretLicenseKey)) $cluster}}
 {{- end -}}
+
+{{/*
+Return the cluster name
+*/}}
+{{- define "newrelic.cluster" -}}
+{{- if .Values.cluster -}}
+  {{- .Values.cluster -}}
```
paologallinaharbur, line 150: Cluster name is populated already in a different way (I do not fully remember how).

I would avoid adding it also as a custom attribute. Therefore I would have merely customAttributesWithoutClusterName.

kang-makes: You are right. Here: https://github.com/newrelic/helm-charts/pull/598/files/0ce03b02a64c24836fef4d55cbf3bea37bcaa068#diff-131da0b26df332b77efd10d0695ed21bff8234051410998ae18562a7d0a43b79R74-R87

And here: https://github.com/newrelic/helm-charts/blob/0ce03b02a64c24836fef4d55cbf3bea37bcaa068/charts/nri-kube-events/templates/configmap.yaml#L13

I am removing those changes.

paologallinaharbur (approved):

charts/nri-kube-events/templates/deployment.yaml
```diff
@@ -85,6 +85,10 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: spec.nodeName
+            {{- if .Values.customAttributes }}
+            - name: NRIA_CUSTOM_ATTRIBUTES
+              value: {{ include "newrelic.customAttributes" . | quote }}
+            {{- end }}
```
roobre, line 91: I think the scope of `"newrelic.customAttributes"` could be reduced by passing just the customAttributes directly:

```suggestion
            {{- with .Values.customAttributes }}
            - name: NRIA_CUSTOM_ATTRIBUTES
              value: {{ include "newrelic.customAttributes" . | quote }}
            {{- end }}
```

WDYT?

kang-makes: This was approved on Wednesday and we said that this was going to be released today.

I am sorry to say that I am not going last-minute changes :(

PR #593, opened 2021-11-17
[newrelic-logging] Made fluent-bit.conf and parsers.conf ConfigMap configurable from values.yaml file

bpschmitt:
#### Is this a new chart

No

#### What this PR does / why we need it:

The changes in this PR provide users the ability to customize the fluent-bit.conf and parsers.conf files using the values.yaml file. Previously, this was statically set in the configmap.yaml file. This will provide much better flexibility for custom fluent-bit configurations.

#### Which issue this PR fixes

A cleanup and continuation of https://github.com/newrelic/helm-charts/pull/551 (previously closed by me).

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [X] Chart Version bumped
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[mychartname]`)

PR #604, opened 2021-11-24
[nri-bundle] Fix kube state metric log spamming in 1.22

### LINKED ISSUE #582: [nri-bundle] KSM creates noisy logs in Kubernetes 1.22
KSM `1.9.8` will attempt to fetch beta resources that no longer exist in `1.22`. which causes log messages like the following to be emmited frequently:

```
E1109 16:56:14.735035       1 reflector.go:156] pkg/mod/k8s.io/client-go@v0.0.0-20191109102209-3c0d1af94be5/tools/cache/reflector.go:108: Failed to list *v1beta1.CertificateSigningRequest: the server could not find the requested resource
```

This creates noise and should be avoided.

kang-makes:
#### Is this a new chart

No

Starting 1.22 lots of v1beta1 APIs where deprecated and we are still using KSM v1 which try to ask the API Server of an object that does not exists. We have in roadmap to support KSM v2 but right now the only way to solve this is to disable `certificatesigningrequest` collector creating an out-of-tree chart for KSM.

  - fixes #582

#### Checklist
- [x] Chart Version bumped


paologallinaharbur:
Oh 😢 , I hoped it was possible to use the standard one passing `Values.collectors.certificatesigningrequests =false`:
```
{{  if .Values.collectors.certificatesigningrequests  }}
        - --collectors=certificatesigningrequests
{{  end  }}

Including a whole new chart for KSM could be bit problematic
```

kang-makes:
Logging https://github.com/newrelic/helm-charts/pull/593 introduced an error that was not properly tested.

I have fixed it so we can merge this fix for KSM and I will help add a proper update to the logging chart so this is testing before merging it.

kang-makes:
@paologallinaharbur Even if you change the values, users values previously deployed by the user would be used and collector could still being used.

Because the chart is not maintained anymore, in case we have to do a future change it is already in the chart so I could be easily changed and updated.

Changing to V2 is on roadmap so this is a temporary patch until we can have enough time to tackle this problem correctly.

kang-makes:
To do one thing at a time, I have separated fixing logging chart to other PR: https://github.com/newrelic/helm-charts/pull/605

Then I will have to rebase and test it again.

paologallinaharbur:
> Even if you change the values, user's values previously deployed would be used and collector could still being used.

This is true in case they manually added "kube-state-metrics.collectors.certificatesigningrequests =false" value setting it to true, no? 🤔 
Otherwise, even if they use "--reuse-values", the new default value at nri-bundle level will overwrite the old one at ksm level

kang-makes:
You are right. And I realised that you were right and that is the reason I created this commit before your comment: https://github.com/newrelic/helm-charts/pull/604/commits/924e87ffb47d342957fd472891d01599b840f80e

This way only discrimates 1.22 which is in the end what we want to patch until we find time to support KSM v2.

This change forces us to include the chart.

roobre:
Regarding embedding a custom KSM chart, the discussion seems more fundamental:

Should we honor KSM defaults when we bundle it as a dependency, or should we consider it a pure dependency and make changes to it through the `nri-bundle` values file as we see fit?

Implications for this are wide and we should discuss and pick a stance. @kang-makes will organize a meeting to discuss this.

kang-makes:
Moving this to draft so the reminded bot does not bother us with this.

PR #605, opened 2021-11-24
[newrelic-logging] Add testing to newrelic logging chart

kang-makes:
#### Is this a new chart

No.

In the PR #593 a bug to this chart was introduced because there is not enough ci testing at your chart level, so now that we are trying to bump `nri-bundle` this chart is failing.

This PR try to add some testing while only this chart is being modified.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-logging/Chart.yaml
+++ b/charts/newrelic-logging/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to deploy New Relic Kubernetes Logging as a DaemonSet, supporting both Linux and Windows nodes and containers
 name: newrelic-logging
-version: 1.10.5
+version: 1.10.6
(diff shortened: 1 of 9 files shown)


kang-makes:
Here you can see that various `values.yaml` are being tested: https://github.com/newrelic/helm-charts/runs/4311946263?check_suite_focus=true#step:7:29

So this should pop problems while templating this chart with different configurations and.

jsubirat (approved):
Thank you very much for your contribution!

gonzalobarbitta:
I am facing similar issue when trying to update to latest version of the chart.
It looks like named template was renamed from `newrelic.lowDataMode` to `newrelic-logging.lowDataMode`. but manifests were not updated. FWIW, changes look good on my end :)

PR #224, opened 2020-12-01
[newrelic-logging] Adds option for additional filter from values file

elialgranti:
#### Is this a new chart
No
#### What this PR does / why we need it:
This change allows adding additional filters to fluent bit from the values file. Currently this requires downloading and modifying the chart.
I've used this to add logtype attributes to logs.
#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version bumped
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[mychartname]`)

PR #609, opened 2021-11-29
[newrelic-logging] add support for extra log filters

els-ipatel:
#### Is this a new chart

No, this is a feature addition to existing `newrelic-logging` chart.

Currently if custom log filtering rules want to be applied, a full copy of `fluentBit.config.filters` would need to be made in order to maintain the default configuration provided (i.e. https://github.com/newrelic/helm-charts/blob/master/charts/newrelic-logging/values.yaml#L65). This PR introduces the ability to specify `extraFilters` value, where custom filters would be appended to the default filter configuration. This is especially useful during chart upgrades, as the default filter values may change in future versions, with the use of a separate parameter for custom filters, chart users only need to be concerned about maintaining their specific configuration. 

  - fixes #

This PR is similar to https://github.com/newrelic/helm-charts/pull/224 - which seems to be stale.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-logging/templates/configmap.yaml
+++ b/charts/newrelic-logging/templates/configmap.yaml
@@ -17,6 +17,9 @@ data:
     {{- else }}
       {{- .Values.fluentBit.config.filters | nindent 4 }}
     {{- end }}
+    {{- if .Values.fluentBit.config.extraFilters }}
+      {{ .Values.fluentBit.config.extraFilters | indent 4}}
+    {{- end }}
     {{- if .Values.fluentBit.config.outputs }}
       {{- .Values.fluentBit.config.outputs | nindent 4 }}
     {{- end }}
--- a/charts/newrelic-logging/Chart.yaml
+++ b/charts/newrelic-logging/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to deploy New Relic Kubernetes Logging as a DaemonSet, supporting both Linux and Windows nodes and containers
 name: newrelic-logging
-version: 1.10.5
+version: 1.11.0
 appVersion: 1.10.0
 home: https://github.com/newrelic/kubernetes-logging
 icon: https://newrelic.com/assets/newrelic/source/NewRelic-logo-square.svg
--- a/charts/newrelic-logging/README.md
+++ b/charts/newrelic-logging/README.md
@@ -145,6 +145,7 @@ See [values.yaml](values.yaml) for the default values
 | `fluentBit.config.service` | Contains fluent-bit.conf Service config                                                                                             |    |
 | `fluentBit.config.inputs` | Contains fluent-bit.conf Inputs config                                                                                             |    |
 | `fluentBit.config.filters` | Contains fluent-bit.conf Filters config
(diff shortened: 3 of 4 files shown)


jsubirat (commented):
Hi @els-ipatel , many thanks for your contribution! We've been considering your change, and we'd be more inclined in introducing a more general setting, `fluentBit.config.extraConfig`, that would allow the user to specify additional configuration but not limited to filters: they would then be able to specify extra INPUTs or OUTPUTs, for instance. Would it be possible for you to rename `fluentBit.config.extraFilters` to `fluentBit.config.extraConfig`, instead? Thank you very much in advance.

els-ipatel:
@jsubirat I did think of this approach initially but decided not to pursue as the construct of this section of the fluenbit config is in the form of a pipeline where config stanzas are positional. My thoughts were this could get quite tricky to keep track of if using a single value for multiple config options and in some cases would not work - e.g. if there is a requirement to add INPUT, FILTER and OUTPUT custom config. At the same time I appreciate this is a NewRelic specific chart and the intended use may not stretch to this type of requirement?

jsubirat:
> @jsubirat I did think of this approach initially but decided not to pursue as the construct of this section of the fluenbit config is in the form of a pipeline where config stanzas are positional. My thoughts were this could get quite tricky to keep track of if using a single value for multiple config options and in some cases would not work - e.g. if there is a requirement to add INPUT, FILTER and OUTPUT custom config. At the same time I appreciate this is a NewRelic specific chart and the intended use may not stretch to this type of requirement?

Hi @els-ipatel , thank you for your answer. Indeed, what you mention makes sense. Could I please ask you to then create three separate "extraXXX" sections: `extraInputs`, `extraFilters` and `extraOutputs`?

els-ipatel:
@jsubirat updated PR as requested.

charts/newrelic-logging/templates/configmap.yaml
```diff
@@ -12,14 +12,23 @@ data:
     {{- if .Values.fluentBit.config.inputs }}
       {{- .Values.fluentBit.config.inputs | nindent 4 }}
     {{- end }}
+    {{- if .Values.fluentBit.config.extraInputs }}
+      {{ .Values.fluentBit.config.extraInputs | indent 4}}
```
jsubirat, line 16: Hi @els-ipatel , I tested the Helm chart installation with extraInputs/extraFilters/extraOutputs and found that this `indent` needs to be a `nindent` instead (like the rest in this file). Otherwise, the resulting Fluent Bit configuration has an invalid indentation. Could you please correct this? Thank you.

jsubirat (changes requested):

charts/newrelic-logging/templates/configmap.yaml
jsubirat: Also please note that a `-` is missing after the `{{-` in these lines for the resulting indentation to be correct. Here is how it should look like everywhere:
```
{{- .Values.fluentBit.config.extraInputs | indent 4 }}
```

els-ipatel: fixed.

jsubirat (approved):
Thank you @els-ipatel for your patience addressing all the comments. It looks good, now! I proceed to merge it.

PR #617, opened 2021-12-14
[newrelic-logging] allow metadata env vars

### LINKED ISSUE #613: [newrelic-logging] Allow additionalEnvVars that use valueFrom
At present it's possible to provide `additionalEnvVars` to set [environment variables][1] on the Fluent Bit container. However, these can only be provided as value strings. There's no provision to allow an [EnvVarSource][2]. For example, we would like to pass in the pod name as an environment variable to make use of in the FluentBit configuration file:

```
additionalEnvVars:
- name: K8S_POD_NAME
  valueFrom:
    fieldRef:
      fieldPath: metadata.name
```

The root of this appears to be [this code in the chart][3] that iterates over `additionalEnvVars` and explicitly looks for the "value" field and uses it to create a new object with a name and a value field.
```
{{- range .Values.fluentBit.additionalEnvVariables }}
- name: {{ .name }}
  value: {{ .value }}
{{- end }}
```
If this instead passed through the environment items unmodified (or explicitly passed through the valuesFrom field) this would enable setting such metadata-derived environment variables.

I want to use the pod name in the Fluent Bit configuration, specifically to set it as a label on metrics emitted by the prometheus remote write plugin. Since every pod has a different name, I can't set this via a static environment variable. It needs to be derived from the metadata.

I want the chart to pass through each item in `additionalEnvVars` directly.

A hard-coded environment variable set to the pod name would also solve my immediate problem, but it wouldn't be very general.

**Additional context**

See also https://github.com/newrelic/helm-charts/issues/81 which would benefit from such a change.

[1]: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#envvar-v1-core
[2]: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#envvarsource-v1-core
[3]: https://github.com/newrelic/helm-charts/blob/334ce7ae4e9753acea9e35c5b923a6795cea262e/charts/newrelic-logging/templates/daemonset.yaml#L87-L90

annettejanewilson:
Problem: [Issue #613][1]. We cannot use `valueFrom` in
`additionalEnvVariables` to add metadata like the pod id as an
environment variable. The code only passes through the `name` and
`value` fields, not `valueFrom`.

Solution: Don't destructure the additionalEnvVariables list at all. Just
emit the whole thing in place.

#### Is this a new chart

No.

Allows the use of `valueFrom` in `additionalEnvVars` provided to Fluent Bit containers.

* fixes #613 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-logging/Chart.yaml
+++ b/charts/newrelic-logging/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart to deploy New Relic Kubernetes Logging as a DaemonSet, supporting both Linux and Windows nodes and containers
 name: newrelic-logging
-version: 1.11.4
+version: 1.11.5
 appVersion: 1.14.0
 home: https://github.com/newrelic/kubernetes-logging
 icon: https://newrelic.com/assets/newrelic/source/NewRelic-logo-square.svg
--- a/charts/newrelic-logging/templates/_helpers.tpl
+++ b/charts/newrelic-logging/templates/_helpers.tpl
@@ -172,3 +172,16 @@ Returns if the template should render, it checks if the required values are set.
 {{- $customSecretKey := include "newrelic-logging.customSecretKey" . -}}
 {{- and (or $licenseKey (and $customSecretName $customSecretKey))}}
 {{- end -}}
+
+{{/*
+If additionalEnvVariables is set, renames to extraEnv. Returns extraEnv.
+*/}}
+{{- define "newrelic-logging.extraEnv" -}}
+{{- if .Values.fluentBit }}
+  {{- if .Values.fluentBit.additionalEnvVariables }}
+    {{- toYaml .Values.fluentBit.additionalEnvVariables -}}
+  {{- else if .Values.fluentBit.extraEnv }}
+    {{- toYaml .Values.fluentBit.extraEnv  -}}
+  {{- end -}}
+{{- end -}}
+{{- end -}}
--- a/charts/newrelic-logging/templates/daemonset-windows.yaml
+++ b/charts/newrelic-logging/templates/daemonset-windows.yaml
@@ -91,10 +91,7 @@ spec:
               value: {{ $.Values.fluentBit.k8sLoggingExclude | quote }}
             - name: LOW_DATA_MODE
               value: {{ include "newrelic-logging.lowDataMode" $ | default "false" | quote }}
-            {{- range $.Values.fluentBit.additionalEnvVariables }}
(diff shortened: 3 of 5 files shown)


roobre:
Hey @annettejanewilson,
Thanks for your contribution!
My current understanding is that this would break compatibility with users who already have the extra env vars specified in the old format, and such a breakage would require a major version bump and approval from @newrelic/logging.

Alternatively it could be possible to write a helper to "shim" the old list format to the new one (which I agree is better).

annettejanewilson:
Hi. It's not clear to me how this breaks backwards compatibility. Could you give an example of something that currently works but that would be broken by this change?

annettejanewilson:
Specifically, without this change, `additionalEnvVariables` must be an array of objects, each with a `name` field and a `value` field. Each such entry is copied into `env` in the container spec. Such an array would still be entirely valid after the change. The only difference (as far as I intend) is that this _also_ allows you to opt to provide a `valueFrom` field instead of a `value` field, with it working in the obvious way. I don't think this stops anything from working.

roobre:
Hey @annettejanewilson,

You are totally right here, I definitely misread the change somehow!

I'll ping @newrelic/logging for approval here :)

achupryn:
Hello NR team, do you have any news regarding that feature request?

we also need that functionality to add env vars from k8s secrets (credentials for additional output) - https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables 

Thanks

roobre:
This would be for @newrelic/logging to prioritize.
We would recommend to use the recently-released common library, as it makes very easy to do this: https://github.com/newrelic/helm-charts/blob/master/library/common-library/DEVELOPERS.md

kang-makes:
In the latest changes we made during the last month we named this `env` and `envFrom`. Here are a couple of examples:
 - https://github.com/newrelic/nri-kubernetes/blob/main/charts/newrelic-infrastructure/values.yaml#L88-L95
 - https://github.com/newrelic/newrelic-k8s-metrics-adapter/blob/main/charts/newrelic-k8s-metrics-adapter/values.yaml#L95-L102
 
The reviewer may want to keep consistency through all the charts :)

jdearana-nr:
Hi @annettejanewilson, please, see @jsubirat suggestion to "_Could you at least extend the aforementioned commented example in values.yml? Thank you very much._"

danybmx (approved):
Thanks for taking care of this @annettejanewilson, @luckslovez!

PR #570, opened 2021-11-04
[newrelic-logging] Allow configuration of kubernetes filter Buffer_Size param

Voziv:
#### What this PR does / why we need it:

We were experiencing the error `[2021/11/04 13:49:51] [ warn] [http_client] cannot increase buffer: current=32000 requested=64768 max=32000` on our newrelic logging pods.

Increasing the buffersize on the kubernetes filter to 64k solved this issue.

I've added a helm chart value for configuring this setting.

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version bumped
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[mychartname]`)

PR #621, opened 2021-12-15
[newrelic-logging] Configurable Kubernetes filter HTTP buffer size

rlees85:
# Allow Kubernetes filter http buffer size to be configured since many pod templates now exceed 32k.

#### Is this a new chart
No

Allow Kubernetes filter http buffer size to be configured since many pod templates now exceed 32k.

Cannot find open issue. Fixes issue where logs are dropped for certain pods that contain a large template. The following is in the newrelic logging agent logs:
```
[2021/11/04 13:49:51] [ warn] [http_client] cannot increase buffer: current=32000 requested=64768 max=32000
```
Fixes the issue without having to maintain an entire custom filters block in Helm values.

#### Checklist
- [x] Chart Version bumped


rlees85:
Related: https://github.com/newrelic/helm-charts/pull/570

rediphone:
Hi there, this would be helpful for me, please could we get this in I have problems with logs that this would resolve - thanks!

rediphone:
Hi guys, sorry to chase but it would be great to get this in - i have an issue which this PR would resolve. if there is a problem please can we discuss what this issue is

thanks very much

rediphone:
any possibility to  progress this please 
@jsubirat
@tejunior

roobre:
Hey there,

Just my two cents here, I think we might benefit to a more flexible solution that allows users to define arbitrary options in the `values.yaml` that get appended to the filters. This should scale better and would not require users to open a PR for each new Fluentbit flag they want to change.

WDYT @newrelic/logging ?

tejunior (approved):

tejunior:
More resources about this issue:
https://github.com/fluent/fluent-bit/issues/2033
https://github.com/fluent/fluent-bit/issues/3112

rlees85:
@roobre I also like this idea, but sadly I don't think that PR will permit it since we are wanting to add settings to an existing (Kubernetes) filter rather than adding new filter blocks. I am not sure if fluentd/fluentbit stacks "same" filter blocks.

PR #623, opened 2021-12-17
newrelic-infrastructure - Charts for new nri-kubernetes architecture

alvarocabanas:
Helm Charts for the new architecture.


charts/newrelic-infrastructure/templates/daemonset-control-plane.yaml
```diff
@@ -0,0 +1,199 @@
...
+      containers:
+        - name: control-plane
+          image: {{ .Values.images.scraper.repository }}:{{ .Values.images.scraper.tag | default .Chart.AppVersion }}
+          imagePullPolicy: {{ .Values.images.scraper.pullPolicy }}
+          securityContext:
+            runAsUser: {{ .Values.runAsUser }}
+            runAsGroup: {{ .Values.runAsGroup }}
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+          env:
```
roobre, line 52: I miss the DNS override env in this line`

roobre (changes requested):
## PARTIAL REVIEW
## TO BE CONTINUED LATER

Looking good overall, I love how a lot of stupid nits have finally been fixed ✨ 

Some things I think we could improve:
- All agent sidecars share a lot of logic. Could we create a helper for them?
- I think it's better to create a new chart rather than replace the old one. This will allow us to select one or the other from `nri-bundle` to allow opt-out or opt-in from the release candidate.
- We could leverage https://github.com/norwoodj/helm-docs to generate the readme
- We are not dropping capabilities by default, which is a good security practice, especially in unprivileged mode
-

charts/newrelic-infrastructure/templates/daemonset-control-plane.yaml
```diff
@@ -0,0 +1,199 @@
...
+        {{- end }}
+        - name: tmpfs-data
+          emptyDir: {}
+        - name: tmpfs-user-data
+          emptyDir: {}
+        - name: tmpfs-tmp
+          emptyDir: {}
+        - name: tmpfs-cache
+          emptyDir: {}
+        {{- if .Values.config }}
```
roobre, line 176: Relfecting what we discussed during the sync meeting: It might seem weird to have both env vars and config exposed to the user.

I don't have a clear opinion on how to adress this. Thoughts?

charts/newrelic-infrastructure/templates/daemonset-control-plane.yaml
```diff
@@ -0,0 +1,199 @@
...
+        - name: config
+          configMap:
+            name: {{ template "newrelic.fullname" . }}
+            items:
+              - key: newrelic-infra.yml
+                path: newrelic-infra.yml
+        {{- end }}
+      {{- with $.Values.priorityClassName }}
```
roobre, line 184: `$` seems redundant.

charts/newrelic-infrastructure/templates/daemonset-kubelet.yaml
```diff
@@ -0,0 +1,294 @@
+{{- if and (.Values.scrapersConfig.kubelet.enabled) (include "newrelic.areValuesValid" .) .Values.enableLinux }}
```
roobre, line 1: As mentioned above, I'd rather fail than skip a template.

charts/newrelic-infrastructure/templates/daemonset-kubelet.yaml
```diff
@@ -0,0 +1,294 @@
...
+                  name: {{ template "newrelic.fullname" . }}-config
+                  key: license
+                  {{- else }}
+                  name: {{ include "newrelic.customSecretName" . }}
+                  key: {{ include "newrelic.customSecretLicenseKey" . }}
+                  {{- end }}
+            {{- with .Values.eventQueueDepth }}
+            - name: "NRIA_EVENT_QUEUE_DEPTH"
+              value: {{ .| quote }}
+            {{- end }}
```
roobre, line 128: This affects how the agent sends data so I believe it should be in all sidecars. Thankfully I don't really see a reason to have it per-sidecar, we can probably share the value for all.

charts/newrelic-infrastructure/templates/daemonset-control-plane.yaml
```diff
@@ -0,0 +1,199 @@
...
+            - name: "NRIA_DISPLAY_NAME"
+            {{- if .Values.prefixDisplayNameWithCluster }}
+              value: "{{ include "newrelic.cluster" . }}:$(NRI_KUBERNETES_NODE_NAME)"
+            {{- else }}
+              valueFrom:
+                fieldRef:
+                  apiVersion: "v1"
+                  fieldPath: "spec.nodeName"
+            {{- end }}
+            {{- end }}
```
roobre, line 122: I never fully understood what this does. Could we document (with a comment) what is it used for, and the implications of changing it?

charts/newrelic-infrastructure/templates/daemonset-kubelet.yaml
```diff
@@ -0,0 +1,294 @@
...
+              valueFrom:
+                fieldRef:
+                  apiVersion: "v1"
+                  fieldPath: "spec.nodeName"
+            {{- end }}
+            {{- end }}
+            {{- with .Values.customAttributes }}
+            - name: "NRIA_CUSTOM_ATTRIBUTES"
+              value: {{ . }}
+            {{- end }}
```
roobre, line 157: This should be in all agent containers as well.

Moreover I think the current way of specifying custom attributes, as a json string, is quite awful: We could improve it by letting the user define these as a yaml and then rendering it with `toJson`:

```
# values.yml
customAttributes:
  foo: bar
  team: awesome
```

```suggestion
            {{- with .Values.customAttributes }}
            - name: "NRIA_CUSTOM_ATTRIBUTES"
              value: {{ . | toJson | quote }}
            {{- end }}
```

roobre: This would break the previous sturcture, but we can create a helper and check if the user defined a string with `kindOf`.

charts/newrelic-infrastructure/templates/daemonset-kubelet.yaml
```diff
@@ -0,0 +1,294 @@
...
+          {{- with .Values.resources }}
+          resources: {{ toYaml . | nindent 12 }}
+          {{- end }}
+      volumes:
+        {{- if .Values.privileged }}
+        - name: dev
+          hostPath:
+            path: /dev
+        {{- if and (.Values.openshift.enabled) (eq .Values.openshift.version "4.x") }}
```
roobre, line 223: As we commented, we should discuss with caos if this is supported and how to tell the agent which .sock to use.

charts/newrelic-infrastructure/templates/daemonset-kubelet.yaml
```diff
@@ -0,0 +1,294 @@
...
+        - name: host-crio-socket
+          hostPath:
+            path: /var/run/crio.sock
+        {{- else }}
+        - name: host-docker-socket
+          hostPath:
+            path: /var/run/docker.sock
```
roobre, line 230: As we discussed, I'd love to be able to fully disable this in unprivileged and tell the agent to not log an error. Something to check with CAOS I think.

charts/newrelic-infrastructure/templates/clusterrole.yaml
```diff
@@ -31,12 +37,12 @@ rules:
...
 metadata:
-  labels: {{ include "newrelic.labels" . | indent 4 }}
+  labels: {{ include "newrelic.labels" . | nindent 4 }}
   name: {{ template "newrelic.fullname" . }}-secrets
 rules:
   - apiGroups: [""]
     resources:
       - "secrets"
-    verbs: ["get", "list"]
+    verbs: ["get", "list", "watch"]
```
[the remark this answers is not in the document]
roobre: @gsanchezgavier has more context on this :)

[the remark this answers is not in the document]
gsanchezgavier: yes we should be able to get the list of namespaces where the secrets are from the controlplane config under the MTLS configuration.

PR #534, opened 2021-09-30
newrelic-infrastructure: add support for privileged flag as a superglobal

roobre:
#### What this PR does / why we need it:

Allows the `newrelic-infrastructure` chart to respond to `global.privileged`, which will override `privileged` (local) if set.

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version bumped
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[mychartname]`)

PR #624, opened 2021-12-20
Newrelic infrastructure v3

alvarocabanas:
New nri-kubernetes architecture beta helm chart


charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,393 @@
...
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+          - matchExpressions:
+              - key: node-role.kubernetes.io/control-plane
+                operator: Exists
+          - matchExpressions:
+              - key: node-role.kubernetes.io/master
+                operator: Exists
```
paologallinaharbur, line 51:
```
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
          - matchExpressions:
              - key: node-role.kubernetes.io/control-plane
                operator: Exists
          - matchExpressions:
              - key: node-role.kubernetes.io/controlplane
                operator: Exists
          - matchExpressions:
              - key: node-role.kubernetes.io/etcd
                operator: Exists
          - matchExpressions:
              - key: node-role.kubernetes.io/master
                operator: Exists
```

We should add this if we want to support rancher out of the box

paologallinaharbur: Note: rancher uses "controlplane" minikue "controlplane" 😢

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,393 @@
...
+    pullPolicy: IfNotPresent
+    tag: 2.7.6
+    repository: newrelic/infrastructure-bundle
+  scraper:
+    pullPolicy: IfNotPresent
+    tag: 0.1.1
+    repository: newrelic/nri-kubernetes
+
+controlPlane:
+  enabled: false
```
paologallinaharbur, line 34: why false by default?

charts/newrelic-infrastructure-beta/templates/control-plane/serviceaccount.yaml
```diff
@@ -0,0 +1,13 @@
+{{- $namespaces := include "newrelic.roleBindingNamespaces" . | fromYaml -}}
+{{if $namespaces.roleBindingNamespaces}}
```
paologallinaharbur, line 2:
```
{{- if and (.Values.controlPlane.enabled) (.Values.rbac.create) (.Values.serviceAccount.create)}}
```

charts/newrelic-infrastructure-beta/values.yaml
alvarocabanas: Added it

charts/newrelic-infrastructure-beta/values.yaml
alvarocabanas: don't we want it false by default? should it be true?

charts/newrelic-infrastructure-beta/templates/control-plane/serviceaccount.yaml
alvarocabanas: Fixed

charts/newrelic-infrastructure-beta/README.md
```diff
@@ -0,0 +1,208 @@
+# newrelic-infrastructure
```
roobre, line 1: For the new values, we could consider using this https://github.com/norwoodj/helm-docs

charts/newrelic-infrastructure-beta/templates/NOTES.txt
```diff
@@ -0,0 +1,34 @@
...
+  nodeAffinity:
+    requiredDuringSchedulingIgnoredDuringExecution:
+      nodeSelectorTerms:
+        - matchExpressions:
+            - key: eks.amazonaws.com/compute-type
+              operator: NotIn
+              values:
+                - fargate
+{{- end }}
+{{- end }}
```
roobre, line 34: Unrelated to this PR but there is likely a bug in here: The first if (`if .Values.image`) should not cover the fargate warning part.

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,234 @@
...
+
+
+{{/* Generate mode label */}}
+{{- define "newrelic.mode" }}
+{{- if .Values.privileged -}}
+privileged
+{{- else -}}
+unprivileged
+{{- end }}
+{{- end -}}
```
roobre, line 35: We are not using this anymore, right?

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,234 @@
...
+{{- define "newrelic.labels" -}}
+helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+app.kubernetes.io/name: {{ include "newrelic.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+mode: {{ template "newrelic.mode" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
```
roobre, line 47: I would suggest to run `helm create` on a different directory and see which labels are the new standard generated by helm. Then we can incorporate those here :)

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,234 @@
...
+{{- if .Values.global -}}
+  {{- if .Values.global.cluster -}}
+      {{- .Values.global.cluster -}}
+  {{- else -}}
+      {{- .Values.cluster | default "" -}}
+  {{- end -}}
+{{- else -}}
+  {{- .Values.cluster | default "" -}}
+{{- end -}}
+{{- end -}}
```
roobre, line 70: Non-global value should have preference over global one. Feel free to copypaste from here: https://github.com/newrelic/helm-charts/pull/543/files#diff-96fa95a04ce4a058fe069a796881e93c225000191df8e909267c073290bc9738R67

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,234 @@
...
+*/}}
+{{- define "newrelic.licenseSecretKey" -}}
+{{ include "newrelic.licenseCustomSecretKey" . | default "licenseKey" }}
+{{- end -}}
+
+{{/*
+Returns nrStaging
+*/}}
+{{- define "newrelic.nrStaging" -}}
+{{- if .Values.global }}
```
roobre, line 130: Same as above

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
roobre: Btw, there should be no need to `| default ""`. The chart will error if we return nothing, when we pipe this to `| required` in the template.

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,234 @@
...
+{{- else if .Values.nrStaging }}
+  {{- .Values.nrStaging -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Returns fargate
+*/}}
+{{- define "newrelic.fargate" -}}
+{{- if .Values.global }}
```
roobre, line 143: And above.

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,234 @@
...
+controlPlane:
+  enabled: true
+  {{- if .Values.controlPlane.scraperConfig.controlPlane }}
+    {{- .Values.controlPlane.scraperConfig.controlPlane | toYaml | nindent 2 }}
+  {{- end }}
+  {{- range $key, $val := .Values.controlPlane.scraperConfig }}
+    {{- if ne $key "controlPlane" -}}
+    {{- nindent 0 $key }}: {{ $val | quote }}
+    {{- end }}
+  {{- end }}
```
roobre, line 163: We can simplify this by using [`omit`](https://masterminds.github.io/sprig/dicts.html), which will as a bonus avoid the problem of what happens if `$val` is not a string.

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,234 @@
...
+{{- end }}
+
+{{/* kubelet scraper config */}}
+{{- define "newrelic.kubelet.scraperConfig" -}}
+kubelet:
+  enabled: true
+  {{- if .Values.kubelet.scraperConfig.kubelet }}
+    {{- .Values.kubelet.scraperConfig.kubelet | toYaml | nindent 2 }}
+  {{- end }}
+  {{- range $key, $val := .Values.kubelet.scraperConfig }}
```
roobre, line 173: As above

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
roobre: Although I don't fully get what we're trying to achieve here.

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,234 @@
...
+{{- end -}}
+
+{{/* ksm scraper config */}}
+{{- define "newrelic.ksm.scraperConfig" -}}
+ksm:
+  enabled: true
+  {{- if .Values.ksm.scraperConfig.ksm }}
+    {{- .Values.ksm.scraperConfig.ksm | toYaml | nindent 2 }}
+  {{- end }}
+  {{- range $key, $val := .Values.ksm.scraperConfig }}
```
roobre, line 196: Aaand as above

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,234 @@
...
+
+{{/*
+Returns the list of namespaces where secrets need to be accessed by the controlPlane Scraper to do mTLS Auth
+*/}}
+{{- define "newrelic.roleBindingNamespaces" -}}
+{{ $namespaceList := list }}
+{{- range $components := .Values.controlPlane.scraperConfig.controlPlane }}
+    {{- range $autodiscover := $components.autodiscover }}
+        {{- range $endpoint := $autodiscover.endpoints }}
+            {{- if and ($endpoint.auth) }}
```
roobre, line 211: Leftover `and`

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,234 @@
...
+            {{- if $endpoint.auth.mtls }}
+            {{- if $endpoint.auth.mtls.secretName }}
+            {{- $namespace := $endpoint.auth.mtls.secretNamespace | default "default" -}}
```
roobre, line 214: Just pinging @gsanchezgavier to check if this default makes sense. Do we allow empty namespaces in the config with this meaning?

charts/newrelic-infrastructure-beta/templates/configmap.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: ConfigMap
+metadata:
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "newrelic.labels" . | nindent 4 }}
+  name: {{ template "newrelic.fullname" . }}
+data:
+    newrelic-infra.yml: |
+{{ toYaml .Values.config | indent 6 }}
+{{ end }}
+{{ if .Values.integrations_config }}
```
roobre, line 12: I would split this CM in two: `integrations-configmap` and `agent-configmap`. Moreover `integrations-configmap` could me moved into the `kubelet` folder.

charts/newrelic-infrastructure-beta/templates/control-plane/configmap.yaml
```diff
@@ -0,0 +1,11 @@
+{{ if .Values.controlPlane.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "newrelic.labels" . | nindent 4 }}
+  name: {{ template "newrelic.fullname" . }}-control-plane
+data:
+  nri-kubernetes-config.yml: |
+{{ toYaml (merge (include "newrelic.controlPlane.scraperConfig" . | fromYaml) .Values.common.scraperConfig) | indent 6 }}
```
roobre, line 10: This line could probably use a comment, I have no idea of what's going on hahaha

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,213 @@
...
+            runAsUser: {{ .Values.runAsUser }}
+            runAsGroup: {{ .Values.runAsGroup }}
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+          env:
+            - name: "NRIA_OVERRIDE_HOSTNAME_SHORT"
+              valueFrom:
+                fieldRef:
+                  apiVersion: "v1"
+                  fieldPath: "spec.nodeName"
```
roobre, line 56: This should be for the agent only, right?

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,213 @@
...
+            - name: NRI_VERBOSE
+              value: "1"
+            {{- end }}
+            {{- with .Values.controlPlane.extraEnv }}
+            {{- tpl . $ | nindent 12 }}
+            {{- end }}
+          {{- with .Values.controlPlane.extraEnvFrom }}
+          envFrom:
+            {{- tpl . $ | nindent 12 }}
+          {{- end }}
```
roobre, line 81: This still assumes extraEnvFrom as a string.

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,213 @@
...
+          volumeMounts:
+            - name: nri-kubernetes-config
+              mountPath: /etc/newrelic-infra/integrations.d/nri-kubernetes-config.yml
+              subPath: nri-kubernetes-config.yml
+            - mountPath: /tmp
+              name: tmpfs-tmp
```
roobre, line 87: Do we use `/tmp` in the _integration_? @paologallinaharbur @gsanchezgavier

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,213 @@
...
+          image: {{ .Values.images.forwarder.repository }}:{{ .Values.images.forwarder.tag }}
+          imagePullPolicy: {{ .Values.images.forwarder.pullPolicy }}
+          securityContext:
+            runAsUser: {{ .Values.runAsUser }}
+            runAsGroup: {{ .Values.runAsGroup }}
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+          ports:
+            - containerPort: 8001
+          env:
```
roobre, line 104: Hostname override short missing here :)

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,213 @@
...
+            - name: "NRIA_DISPLAY_NAME"
+            {{- if .Values.prefixDisplayNameWithCluster }}
+              value: "{{ include "newrelic.cluster" . }}:$(K8S_NODE_NAME)"
+            {{- else }}
+              valueFrom:
+                fieldRef:
+                  apiVersion: "v1"
+                  fieldPath: "spec.nodeName"
+            {{- end }}
+            {{- end }}
```
roobre, line 126: I'd love a comment here explaining what this achieves

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,213 @@
...
+            {{- if (include "newrelic.customAttributes" .) }}
+            - name: "NRIA_CUSTOM_ATTRIBUTES"
+              value: {{ include "newrelic.customAttributes" . }}
+            {{- end }}
```
roobre, line 130: I love how this looks! 💖

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,213 @@
...
+            - name: NRIA_IS_SECURE_FORWARD_ONLY
+              value: "false"
+            - name: NRIA_IS_FORWARD_ONLY
+              value: "true"
+            {{- with .Values.controlPlane.extraEnv }}
+            {{- tpl . $ | nindent 12 }}
+            {{- end }}
+          {{- with .Values.controlPlane.extraEnvFrom }}
+          envFrom:
+            {{- tpl . $ | nindent 12 }}
```
roobre, line 154: Assuming string

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,213 @@
...
+            name: {{ template "newrelic.fullname" . }}-control-plane
+            items:
+              - key: nri-kubernetes-config.yml
+                path: nri-kubernetes-config.yml
+        - name: tmpfs-data
+          emptyDir: {}
+        - name: tmpfs-user-data
+          emptyDir: {}
+        - name: tmpfs-tmp
+          emptyDir: {}
```
roobre, line 186: We could probably coalesce all these into one `tmpfs` volume and mount `subPath`s of it.

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,213 @@
...
+        {{- end }}
+        {{- with .Values.controlPlane.extraVolumes }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      {{- with .Values.priorityClassName }}
+      priorityClassName: {{ . }}
+      {{- end }}
+      {{- with .Values.controlPlane.affinity.nodeAffinity }}
+      affinity:
+        nodeAffinity: {{ . | toYaml | nindent 10 }}
```
roobre, line 203:
```suggestion
        nodeAffinity:
          {{- . | toYaml | nindent 10 }}
```

charts/newrelic-infrastructure-beta/templates/ksm/deployment.yaml
```diff
@@ -0,0 +1,206 @@
...
+        {{- end }}
+        mode: {{ include "newrelic.mode" . }}
+    spec:
+      {{- with .Values.images.pullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "newrelic.serviceAccountName" . }}
+      containers:
+        - name: forwarder
```
roobre, line 36: Sometimes we put the `forwarder` container first, while others we put the integration container first. Mostly a nit but can hurt readability a bit. Could we unify this criteria?

charts/newrelic-infrastructure-beta/templates/ksm/deployment.yaml
```diff
@@ -0,0 +1,206 @@
+{{- if and (.Values.ksm.enabled) (.Values.enableLinux) }}
```
roobre, line 1: I understand the motivation for`.Values.enableLinux` in other components, but not here: KSM scraper should not care about windows or linux.

I think we should completely get rid of any windows reference for this chart, and add it later if required. The old chart with windows-specific clauses will still be there.

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,398 @@
...
+windowsOsList:
+  # - version: 1809
+  #   imageTag: 2.2.0-windows-1809-alpha
+  #   buildNumber: 10.0.17763
+  # - version: 1909
+  #   imageTag: 2.2.0-windows-1909-alpha
+  #   buildNumber: 10.0.18363
+  - version: 2004
+    imageTag: 2.2.0-windows-2004-alpha
+    buildNumber: 10.0.19041
```
roobre, line 16: I would get rid of windows stuff in the new chart until (and if) we add it properly.

charts/newrelic-infrastructure-beta/values.yaml
roobre: There can be a weird interaction if it is enabled, set to `daemonSet`, but `privileged` is false. We need to plan this carefully @gsanchezgavier

roobre: (In any case I think it should be on by default if possible)

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,398 @@
...
+  extraEnvFrom: []
+  extraVolumes: []
+  extraVolumeMounts: []
+  resources:
+    limits:
+      memory: 300M
+    requests:
+      cpu: 100m
+      memory: 150M
+  scraperConfig:
```
roobre, line 69: I'd call this simply
```suggestion
  config:
```

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,398 @@
...
+    #    interval: 35s
+    #    timeout: 60s
+    controlPlane:
```
roobre, line 72: Having `.Values.controlPlane.config.controlPlane` seems redundant. Could we have just `.Values.controlPlane.config`?

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,398 @@
...
+      etcd:
+        enabled: true
+        autodiscover:
```
roobre, line 75: I'd definitely expect some comments here, this config is massively complex.

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,398 @@
...
+            namespace: kube-system
+            matchNode: true
+            endpoints:
+              - url: https://localhost:10257
+                insecureSkipVerify: true
+                auth:
+                  type: bearer
+                  #           mtls:
+                  #           secretName: secret-name
+                  #           secretNamespace: secret-namespace
```
roobre, line 170: Weird indentation

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,398 @@
...
+  extraVolumes: []
+  extraVolumeMounts: []
+  resources:
+    limits:
+      memory: 300M
+    requests:
+      cpu: 100m
+      memory: 150M
+  scraperConfig:
+    kubelet: {}
```
roobre, line 231: Ditto, I'd prefer if we could have just `config: {}`

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,398 @@
...
+#    timeout: 40s
+    ksm:
+      namespace: ""
+      port: 0
+      scheme: ""
+      selector: ""
+      staticURL: ""
+
+common:
+  scraperConfig:
```
roobre, line 285:
```suggestion
common:
  config:
```

I'd also move this to the beginning, I think it reads out better.

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,398 @@
...
+# K8s DaemonSets update strategy.
+updateStrategy:
+  type: RollingUpdate
+  rollingUpdate:
+    maxUnavailable: 1
+
+# Kubelet's agent envs
+# Custom attributes to be passed to the New Relic agent
+customAttributes:
+  clusterName: $(CLUSTER_NAME)
```
roobre, line 299: Customers could delete this by accident, and we need it for integrations to report sensical things. Would it make sense to add the special `clusterName` key with a helper template?

`customAttributes: merge .Values.customAttributes (dict "clusterName" "$(CLUSTER_NAME)")`

Or something like that

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,398 @@
...
+# may be restricted.  See OpenShift docs for more information.
+# https://www.openshift.com/blog/a-guide-to-openshift-and-uids
+runAsUser: 1000
+# For unprivileged mode - default is 2000.
+runAsGroup: 2000
+
+### Openshift ###
+# Set openshift.enabled true when installing in an OpenShift Cluster
+openshift:
+  enabled: false
```
roobre, line 328: Do we have any openshift-specific behavior still?

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,398 @@
...
+podSecurityContext: {}
+# fsGroup: 2000
+
+securityContext: {}
+  # capabilities:
+  #   drop:
+  #   - ALL
+  # readOnlyRootFilesystem: true
+  # runAsNonRoot: true
+# runAsUser: 1000
```
roobre, line 342: This `securityContext` seems redundant with `runAsUser` above.

charts/newrelic-infrastructure-beta/values.yaml
paologallinaharbur: adding the CRI-O socket

charts/newrelic-infrastructure-beta/ci/test-values.yaml
```diff
@@ -0,0 +1,3 @@
+global:
```
kang-makes, line 1: This shouldn't be "only global" and global should be more tested at `nri-bundle` level than here. But you should add as many values you want.

This only test that the manifests templates correctly. Not even if they has sense/are Kubernetes cluster. But stop lots of footguns.

charts/newrelic-infrastructure-beta/templates/NOTES.txt
```diff
@@ -0,0 +1,26 @@
+Your deployment of the New Relic Infrastructure agent is complete. You can check on the progress of this by running the following command:
```
kang-makes, line 1: I would add "If you have not used `--wait` during the installation of this helm chart" to this sentence. And I would add it to the docs so Helm waits until it is deployed.

This has make me remind of creating a (Helm Chart test)[https://helm.sh/docs/topics/chart_tests/] so we can also tests things like "the secrets you have put are correct" or "the license key is correct, valid and allow to ingest data" (if possible).

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,182 @@
...
+Return the cluster name
+*/}}
+{{- define "newrelic.cluster" -}}
+{{- if .Values.cluster -}}
+  {{- .Values.cluster -}}
+{{- else if .Values.global -}}
+  {{- if .Values.global.cluster -}}
+    {{- .Values.global.cluster -}}
+  {{- end -}}
+{{- end -}}
```
kang-makes, line 69: Avoid lots of if-else.. All this curly braces are more difficult to read than a `required "error message" (default of two values)`

```suggestion
{{- required "A valid .Values.cluster or .Values.global.cluster is required" (default .Values.cluster .Values.global.cluster) -}}
```

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,182 @@
...
+{{- end -}}
+
+{{/*
+Return local licenseKey if set, global otherwise
+*/}}
+{{- define "newrelic.licenseKey" -}}
+{{- if .Values.licenseKey -}}
```
kang-makes, line 76: Same as above

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,182 @@
...
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return the name of the secret holding the License Key
+*/}}
+{{- define "newrelic.licenseCustomSecretName" -}}
+{{- if .Values.customSecretName -}}
+  {{- .Values.customSecretName -}}
+{{- else if and .Values.global -}}
```
kang-makes, line 91: Maybe here there is no possibility of doing the above proposed change. But why have you used a `and` and then put another `if` in other line?

```suggestion
{{- else if (and .Values.global .Values.global.customSecretName) -}}
```

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
kang-makes: Remember to put a `global` with an empty `cluster` not only here but in the `nri-bundle`.

charts/newrelic-infrastructure-beta/templates/secret.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if not (include "newrelic.licenseCustomSecretName" .) }}
+{{- if (include "newrelic.licenseKey" . | required "licenseKey or customSecretName and customSecretLicenseKey must be set" ) }}
```
kang-makes, line 2: Templates should only template. `required` should be in `helpers`. The helper should provide directly the value needed to template.

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,182 @@
...
+*/}}
+{{- define "newrelic.licenseSecretName" -}}
+{{ include "newrelic.licenseCustomSecretName" . | default (printf "%s-license" (include "newrelic.fullname" . )) }}
+{{- end -}}
+
+{{/*
+Return the name key for the License Key inside the secret
+*/}}
+{{- define "newrelic.licenseCustomSecretKey" -}}
+{{- if .Values.customSecretLicenseKey -}}
```
kang-makes, line 109: Same as above. Move the `required` below to here.

charts/newrelic-infrastructure-beta/templates/agent-configmap.yaml
```diff
@@ -0,0 +1,11 @@
+{{ if .Values.config }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "newrelic.labels" . | nindent 4 }}
+  name: {{ template "newrelic.fullname" . }}-agent
+data:
+    newrelic-infra.yml: |
```
kang-makes, line 9: This should have 2 spaces. The `indent` below 4. We should add a YAML linter to ther charts and the templated exist from the `ci` values...

charts/newrelic-infrastructure-beta/templates/control-plane/clusterrole.yaml
```diff
@@ -0,0 +1,45 @@
+{{- if and (.Values.controlPlane.enabled) (.Values.rbac.create) }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels: {{ include "newrelic.labels" . | nindent 4 }}
```
kang-makes, line 5: This is going to fail.

charts/newrelic-infrastructure-beta/templates/control-plane/clusterrole.yaml
```diff
@@ -0,0 +1,45 @@
...
+    verbs:
+      - use
+  {{- end -}}
+{{- $namespaces := include "newrelic.roleBindingNamespaces" . | fromYaml -}}
+{{- if $namespaces.roleBindingNamespaces}}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels: {{ include "newrelic.labels" . | nindent 4 }}
```
kang-makes, line 37: Here too.

charts/newrelic-infrastructure-beta/templates/control-plane/clusterrolebinding.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if and (.Values.controlPlane.enabled) (.Values.rbac.create) }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  labels: {{ include "newrelic.labels" . | nindent 4 }}
```
kang-makes, line 5: All the `labels` are not correct.
```suggestion
  labels:
    {{ include "newrelic.labels" . | nindent 4 }}
```

charts/newrelic-infrastructure-beta/templates/control-plane/configmap.yaml
```diff
@@ -0,0 +1,16 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "newrelic.labels" . | nindent 4 }}
+  name: {{ template "newrelic.fullname" . }}-control-plane
+data:
+  nri-kubernetes-config.yml: |
+    {{- .Values.common.config | toYaml | nindent 4 }}
+    controlPlane:
+      enabled: true
+    {{- if .Values.controlPlane.config -}}
+    {{- .Values.controlPlane.config | toYaml | nindent 6 }}
```
kang-makes, line 14: This will not get the `config` key.
```suggestion
    config:
      {{- .Values.controlPlane.config | toYaml | nindent 6 }}
```

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,211 @@
+{{- if and (.Values.controlPlane.enabled) (not (include "newrelic.fargate" .)) }}
+apiVersion: apps/v1
+kind: {{ .Values.controlPlane.kind }}
+metadata:
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "newrelic.labels" . | nindent 4 }}
```
kang-makes, line 6: As all the `labels` this is not rendering correctly.

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,211 @@
...
+  template:
+    metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        app.kubernetes.io/name: {{ include "newrelic.name" . }}-control-plane
+        app.kubernetes.io/instance: {{ .Release.Name }}-control-plane
+        {{- with .Values.podLabels }}
```
kang-makes, line 28: Add also the `.global.podLabels` that came as a PR today to the legacy chart.

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,211 @@
...
+      app.kubernetes.io/name: {{ include "newrelic.name" . }}-control-plane
+      app.kubernetes.io/instance: {{ .Release.Name }}-control-plane
+  template:
+    metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        app.kubernetes.io/name: {{ include "newrelic.name" . }}-control-plane
```
kang-makes, line 26: This is not using `{{ include "newrelic.labels" . | nindent 4 }}`

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,211 @@
...
+      serviceAccountName: {{ template "newrelic.serviceAccountName" . }}-control-plane
+      {{- if .Values.privileged }}
+      hostNetwork: true
+      dnsPolicy: ClusterFirstWithHostNet
+      {{- end }}
+      containers:
+        - name: control-plane
+          image: {{ .Values.images.scraper.repository }}:{{ .Values.images.scraper.tag | default .Chart.AppVersion }}
+          imagePullPolicy: {{ .Values.images.scraper.pullPolicy }}
+          securityContext:
```
kang-makes, line 46: All the `securityContext`should be in the values to be idiomatic with the `helm` standards. Take a look to a freshly made `helm create newrelic` in a temporary folder and take it as an example.

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,211 @@
...
+            {{- end }}
+            {{- if (include "newrelic.customAttributes" .) }}
+            - name: "NRIA_CUSTOM_ATTRIBUTES"
+              value: {{ include "newrelic.customAttributes" . }}
+            {{- end }}
+            - name: NRIA_LICENSE_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "newrelic.licenseSecretName" . | required "licenseKey or customSecretName and customSecretLicenseKey must be set" }}
+                  key: {{ include "newrelic.licenseSecretKey" . | required "licenseKey or customSecretName and customSecretLicenseKey must be set" }}
```
kang-makes, line 134: `required` should be in the helper. Helpers should provide directly what is needed to render.

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,211 @@
...
+            {{- if (include "newrelic.nrStaging" .) }}
+            - name: NRIA_STAGING
+              value: "true"
+            {{- end }}
+            {{- with .Values.eventQueueDepth }}
+            - name: "NRIA_EVENT_QUEUE_DEPTH"
+              value: {{ .| quote }}
+            {{- end }}
```
kang-makes, line 142: We talked today about something like `.Values.common.agentConfig` here we should use or a configMap o a `range key value` from the YAML and paint them all.

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
kang-makes: Are we sure the `affinity` key is going to exist and not return a `nil` and break the templating?

charts/newrelic-infrastructure-beta/templates/control-plane/rolebinding.yaml
```diff
@@ -0,0 +1,18 @@
+{{- $namespaces := include "newrelic.roleBindingNamespaces" . | fromYaml -}}
```
kang-makes, line 1: Add for clarity.
```suggestion
{{- $namespaces := (include "newrelic.roleBindingNamespaces" . | fromYaml) -}}
```

charts/newrelic-infrastructure-beta/templates/kubelet/daemonset.yaml
```diff
@@ -0,0 +1,273 @@
...
+          resources: {{ toYaml . | nindent 12 }}
+          {{- end }}
+      volumes:
+        {{- if .Values.privileged }}
+        - name: dev
+          hostPath:
+            path: /dev
+        - name: host-docker-socket
+          hostPath:
+            path: /var/run/docker.sock
```
kang-makes, line 211: Is this still necessary? What about other CRI? Because we have some OpenShift compatibility issues here then...

charts/newrelic-infrastructure-beta/templates/kubelet/daemonset.yaml
```diff
@@ -0,0 +1,273 @@
...
+          emptyDir: {}
+        {{- end }}
+        - name: tmpfs-tmp
+          emptyDir: {}
+        - name: nri-kubernetes-config
+          configMap:
+            name: {{ template "newrelic.fullname" . }}-kubelet
+            items:
+              - key: nri-kubernetes-config.yml
+                path: nri-kubernetes-config.yml
```
kang-makes, line 231: Could we avoid using a simple file from the `configMap`? If we could mount the whole `configMap` as a filder with its files, the cluster will eventually change the file if it is changed by hand.

If not we should use an annotation with the config related to each integration hashed to force the restart of the pod: https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

charts/newrelic-infrastructure-beta/templates/kubelet/daemonset.yaml
```diff
@@ -0,0 +1,273 @@
...
+        {{- end }}
+        {{- with .Values.kubelet.extraVolumes }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      {{- with $.Values.priorityClassName }}
+      priorityClassName: {{ . }}
+      {{- end }}
+      {{- if .Values.kubelet.affinity.nodeAffinity }}
+      affinity:
+        nodeAffinity: {{ .Values.kubelet.affinity.nodeAffinity | toYaml | nindent 10 }}
```
kang-makes, line 253:
```suggestion
        nodeAffinity:
          {{ .Values.kubelet.affinity.nodeAffinity | toYaml | nindent 10 }}
```

charts/newrelic-infrastructure-beta/templates/kubelet/integrations-configmap.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: ConfigMap
+metadata:
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "newrelic.labels" . | nindent 4 }}
+  name: {{ template "newrelic.fullname" . }}-integrations-cfg
+data:
+{{ range .Values.integrations_config -}}
+{{ .name | indent 2 }}: |
+    ---
+{{ toYaml .data | indent  4 }}
```
kang-makes, line 12:
```suggestion
  {{- range $key, $value := .Values.integrations_config -}}
  {{ $key }}: |
    {{- toYaml $value | indent  4 }}
```

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,387 @@
...
+
+# Accounts and subaccounts created before July 20, 2020 will report, by default, process metrics unless this config option is explicitly set to "false"
+# On the other hand New Relic accounts created after July 20, 2020 will **not** send, by default, any process metrics unless this config option is explicitly set to "true",
+# https://docs.newrelic.com/docs/release-notes/infrastructure-release-notes/infrastructure-agent-release-notes/new-relic-infrastructure-agent-1120
+# enableProcessMetrics: "false"
+
+# This can be set, the default is shown below
+# logFile: /var/log/nr-infra.log
+
+verboseLog: false
```
kang-makes, line 312: Could we add the verbosity as part of each integration?

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,387 @@
...
+  create: true
+  annotations: {}
+  # If not set and create is true, a name is generated using the fullname template
+  name: ""
+
+# For unprivileged mode - default is 1000.
+# Can also be modified for OpenShift clusters where the runAsUser range
+# may be restricted.  See OpenShift docs for more information.
+# https://www.openshift.com/blog/a-guide-to-openshift-and-uids
+runAsUser: 1000
```
kang-makes, line 324: This should be the whole security context and it should be a security context for each integration.

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,387 @@
...
+# For unprivileged mode - default is 2000.
+runAsGroup: 2000
+
+podAnnotations: {}
```
kang-makes, line 328: We have a `common` for configuration, right? Does it make sense if this is at `.global`, `.common` and each integration level?

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
gsanchezgavier: for some reason i was not notified by this ping 😞 , or perhaps it lost in the nested mails sorry.

yes we should we made the change that empty namespaces for secrets are not allowed anymore, and we agree that instead of having this logic (which was the old one) we are just going to left the mtls example comment , which is already there i think

charts/newrelic-infrastructure-beta/templates/control-plane/configmap.yaml
gsanchezgavier: i think that `config:` is not needed actually 
https://github.com/newrelic/nri-kubernetes/blob/develop/internal/config/config.go#L47

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,211 @@
...
+          image: {{ .Values.images.scraper.repository }}:{{ .Values.images.scraper.tag | default .Chart.AppVersion }}
+          imagePullPolicy: {{ .Values.images.scraper.pullPolicy }}
+          securityContext:
+            runAsUser: {{ .Values.runAsUser }}
+            runAsGroup: {{ .Values.runAsGroup }}
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+          env:
+            - name: "NRI_KUBERNETES_HTTPSERVERPORT"
+              value: "8001"
```
gsanchezgavier, line 53: i thought that we were using the config `HTTPServerPort` to configure both , the agent and the integration to hit the agent endpoint, otherwise this config might be changed but this will fail since the agent will be listening in other port

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
gsanchezgavier: not that i know , even calculated metrics are using in memory storage . Just found this that we might have the chance to burn 😀 https://github.com/newrelic/nri-kubernetes/tree/develop/src/storage

gsanchezgavier: I guess that this was the one needed for the `/tmp`

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
```diff
@@ -0,0 +1,211 @@
...
+            - name: NRIA_STAGING
+              value: "true"
+            {{- end }}
+            {{- with .Values.eventQueueDepth }}
+            - name: "NRIA_EVENT_QUEUE_DEPTH"
+              value: {{ .| quote }}
+            {{- end }}
+            # Using FORWARD_ONLY mode to avoid the creating a host entity for the pod.
+            - name: NRIA_IS_SECURE_FORWARD_ONLY
+              value: "false"
```
gsanchezgavier, line 145: i would remove this and keep the mode is being used or leave a comment why is here

gsanchezgavier (commented):
great job Alvaro, i did a review with focus on the controlplane related stuff .

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
gsanchezgavier: https://github.com/newrelic/nri-kubernetes/pull/297

charts/newrelic-infrastructure-beta/README.md
```diff
@@ -0,0 +1,207 @@
...
+| controlPlane.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[1].matchExpressions[0].key | string | `"node-role.kubernetes.io/controlplane"` |  |
+| controlPlane.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[1].matchExpressions[0].operator | string | `"Exists"` |  |
+| controlPlane.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[2].matchExpressions[0].key | string | `"node-role.kubernetes.io/etcd"` |  |
+| controlPlane.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[2].matchExpressions[0].operator | string | `"Exists"` |  |
+| controlPlane.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[3].matchExpressions[0].key | string | `"node-role.kubernetes.io/master"` |  |
+| controlPlane.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[3].matchExpressions[0].operator | string | `"Exists"` |  |
+| controlPlane.annotations | object | `{}` |  |
+| controlPlane.config.apiServer.autodiscover[0].endpoints[0].auth.type | string | `"bearer"` |  |
```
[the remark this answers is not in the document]
kang-makes: The tool uses the comments from the `values.yaml` file. I would let this comment to have a follow up but I won't do it until the resto of the chart (and possible changes in the values) are finished.

charts/newrelic-infrastructure-beta/templates/kubelet/integrations-configmap.yaml
roobre: It _should_ be like this, but the `integrations_config` thing has a very weird format in the values:
```yaml
# integrations_config:
#   - name: nri-redis.yaml
#     data:
#       discovery:
```
So we need to use that weird range.

Not sure if it's worth it changing it now, and adding it to the compat layer. WDYT?

charts/newrelic-infrastructure-beta/templates/secret.yaml
roobre: As above, I still prefer the required outside (it will highlight better the line where the error is). Is there a recommendation on the contrary?

charts/newrelic-infrastructure-beta/values.yaml
kang-makes: As OpenShift is now "enough compliant" why not create a variable with the CRI so we could support differente CRI in the future instead on focusing on a distribution.

charts/newrelic-infrastructure-beta/templates/kubelet/daemonset.yaml
kang-makes: Let me be. It was like 9 in the night...

charts/newrelic-infrastructure-beta/templates/kubelet/daemonset.yaml
```diff
@@ -0,0 +1,273 @@
...
+      {{- else if include "newrelic.fargate" . }}
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: eks.amazonaws.com/compute-type
+                    operator: NotIn
+                    values:
+                      - fargate
```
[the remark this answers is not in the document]
kang-makes: So is a NTH?

charts/newrelic-infrastructure-beta/templates/kubelet/integrations-configmap.yaml
kang-makes: Compatibility layer. Another NTH.

charts/newrelic-infrastructure-beta/templates/secret.yaml
kang-makes: Let's continue this conversation [above](https://github.com/newrelic/helm-charts/pull/624/files#r774742800)

kang-makes (commented):
It is looking pretty good. Almost all the comments are nits.

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
kang-makes: Is a NTH. Resolving this conversation.

charts/newrelic-infrastructure-beta/templates/control-plane/daemonset.yaml
roobre: Seems like we're still using it, I'd address this on a separate PR.

charts/newrelic-infrastructure-beta/templates/control-plane/rolebinding.yaml
```diff
@@ -0,0 +1,19 @@
+{{- $namespaces := (include "newrelic.roleBindingNamespaces" . | fromYaml) -}}
```
roobre, line 1: Do we need `if .Values.rbac.create` here?

charts/newrelic-infrastructure-beta/templates/ksm/configmap.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if .Values.ksm.enabled -}}
```
roobre, line 1: While these make sense conceptually, I'm not sure if we should let users disable `ksm` and/or `kubelet` components. If they do, cluster explorer (and any other experience we build in the future) will certainly not work, and the UI will return unmeaningful errors.

If we chose to allow users to flip this switch we should at least print a warning in `NOTES.txt`. WDYT?

charts/newrelic-infrastructure-beta/templates/ksm/deployment.yaml
```diff
@@ -0,0 +1,203 @@
+{{- if (.Values.ksm.enabled) }}
```
roobre, line 1: Same as above.

charts/newrelic-infrastructure-beta/templates/kubelet/daemonset.yaml
```diff
@@ -0,0 +1,275 @@
...
+        checksum/nri-kubernetes: {{ include (print $.Template.BasePath "/kubelet/configmap.yaml") . | sha256sum }}
+        {{- if not (include "newrelic.licenseCustomSecretName" .) }}
+        checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
+        {{- end }}
+        {{- if .Values.common.agentConfig }}
+        checksum/agent-config: {{ include (print $.Template.BasePath "/agent-configmap.yaml") . | sha256sum }}
+        {{- end }}
+        {{- if .Values.integrations_config }}
+        checksum/integrations_config: {{ include (print $.Template.BasePath "/kubelet/integrations-configmap.yaml") . | sha256sum }}
+        {{- end }}
```
roobre, line 32: The agent supports hot reloading so _in theory_ this should not be necessary. However we know this sometimes behaves erratically on k8s. Should we force a restart of the agent pod in this case? @newrelic/coreint @newrelic/caos

charts/newrelic-infrastructure-beta/templates/kubelet/daemonset.yaml
```diff
@@ -0,0 +1,275 @@
...
+              value: "true"
+            - name: "NRIA_HTTP_SERVER_PORT"
+              value: "8003"
+            - name: NRIA_LICENSE_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "newrelic.licenseSecretName" . | required "licenseKey or customSecretName and customSecretLicenseKey must be set" }}
+                  key: {{ include "newrelic.licenseSecretKey" . | required "licenseKey or customSecretName and customSecretLicenseKey must be set" }}
+            {{- /* Ugly hack to check if .Values.enableProcessMetrics is defined: */}}
+            {{- if ne (print .Values.enableProcessMetrics) "<nil>" }}
```
roobre, line 123: We could use this slightly less ugly hack:
https://github.com/newrelic/helm-charts/pull/534/files#diff-c36211c7f3d9d611c1101c50f0dd080378972f7d7e815d89c56872474b1a240fR27

```suggestion
            {{- Check if enableProcessMetrics is present in the map */}}
            {{- if get .Values.enableProcessMetrics) | kindIs "bool" }}
```

charts/newrelic-infrastructure-beta/templates/kubelet/daemonset.yaml
```diff
@@ -0,0 +1,275 @@
...
+          hostPath:
+            path: /var/log
+        - name: host-volume
+          hostPath:
+            path: /
+        {{- else }}
+        - name: tmpfs-data
+          emptyDir: {}
+        - name: tmpfs-user-data
+          emptyDir: {}
```
roobre, line 223: Why are these two `else`d?

charts/newrelic-infrastructure-beta/templates/kubelet/daemonset.yaml
```diff
@@ -0,0 +1,275 @@
...
+        {{- end }}
+        {{- if .Values.integrations_config }}
+        - name: nri-integrations-cfg-volume
+          configMap:
+            name: {{ template "newrelic.fullname" . }}-integrations-cfg
+        {{- end }}
+        {{- with .Values.kubelet.extraVolumes }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      {{- with $.Values.priorityClassName }}
```
roobre, line 249: Leftover `$`
```suggestion
      {{- with .Values.priorityClassName }}
```

charts/newrelic-infrastructure-beta/values.yaml
```diff
@@ -0,0 +1,389 @@
...
+    pullPolicy: IfNotPresent
+    tag: 1.20.5
+    repository: newrelic/k8s-events-forwarder
+    registry: docker.io
+  agent:
+    pullPolicy: IfNotPresent
+    tag: 2.7.6
+    repository: newrelic/infrastructure-bundle
+    registry: docker.io
+  scraper:
```
roobre, line 17: Since we've gotten rid of the `scraper` name elsewhere, should we change it here as well? Something like `nri-kubernetes`?

charts/newrelic-infrastructure-beta/values.yaml
kang-makes: Good point.

charts/newrelic-infrastructure-beta/templates/kubelet/daemonset.yaml
roobre: Inotify can be unreliable so let's keep it for now.

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,190 @@
...
+{{- define "newrelic.customAttributesWithoutClusterName" -}}
+{{- if kindOf .Values.customAttributes | eq "string" -}}
+{{  .Values.customAttributes }}
+{{- else -}}
+{{ .Values.customAttributes | toJson }}
+{{- end -}}
+{{- end -}}
+
+{{- define "newrelic.customAttributes" -}}
+{{- merge (include "newrelic.customAttributesWithoutClusterName" . | fromJson) (dict "clusterName" "$(CLUSTER_NAME)") | toJson | quote }}
```
kang-makes, line 189: With `$(CLUSTER_NAME)` Kubernetes is using other environment variable from the pod itself. There is no pod with `CLUSTER_NAME` variable so it is going to be empty. Use the exist of `{{ include "newrelic.cluster" . }}`

charts/newrelic-infrastructure-beta/templates/_helpers.tpl
```diff
@@ -0,0 +1,190 @@
...
+{{/* Common labels */}}
+{{- define "newrelic.labels" -}}
+helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+{{ include "newrelic.selectorLabels" . }}
+mode: {{ template "newrelic.mode" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
```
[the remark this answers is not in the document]
kang-makes: Nevermind. I saw that there `{{ include "newrelic.selectorLabels" . }}` is being used.

charts/newrelic-infrastructure-beta/templates/kubelet/daemonset.yaml
```diff
@@ -0,0 +1,274 @@
...
+              name: tmpfs-tmp
+            {{- with .Values.kubelet.extraVolumeMounts }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
+          {{- with .Values.kubelet.resources }}
+          resources: {{ toYaml . | nindent 12 }}
+          {{- end }}
+        - name: agent
+          image: "{{ .Values.images.agent.registry }}/{{ .Values.images.agent.repository }}:{{ .Values.images.agent.tag }}"
+          command: [ "/sbin/tini", "--", "/usr/bin/newrelic-infra" ]
```
kang-makes, line 95: Does not this come from the image's entrypoint itself? I mean, if tomorrow (just a silly example) CAOS team changes the zombie reaper from `tini` to `surpervisord` we are having a failing chart because we are hardcoding it. I would remove it.
```suggestion
```

kang-makes (commented):
The last couple of nits.

charts/newrelic-infrastructure-beta/ci/test-values.yaml
```diff
@@ -0,0 +1,147 @@
...
+      memory: 150M
+
+# K8s DaemonSets update strategy.
+updateStrategy:
+  type: RollingUpdate
+  rollingUpdate:
+    maxUnavailable: 1
+
+enableProcessMetrics: "false"
+logFile: /var/log/nr-infra.log
```
roobre, line 46: This breaks in unprivileged mode, I'd rather get rid of it.

alvarocabanas: I've just removed it.

charts/newrelic-infrastructure-v3/values.yaml
```diff
@@ -353,32 +353,28 @@ prefixDisplayNameWithCluster: false
...
-# filename generated and the Infrastructure agent only looks for YAML files. The data
-# part is the actual integration configuration as described in the spec here:
+# configuration under `integrations`. You just need to create a new entry where
+# the key is the filename of the configuration file and the value is the content of
+# the integration configuration. The key must end in ".yaml" as this will be the
+# filename generated and the Infrastructure agent only looks for YAML files.
+# The data is the actual integration configuration as described in the spec here:
 # https://docs.newrelic.com/docs/integrations/integrations-sdk/file-specifications/integration-configuration-file-specifications-agent-v180
-# For example, if you wanted do to monitor a Redis instance that has a label "app=redis"
+# For example, if you wanted to to monitor a Redis instance that has a label "app=sampleapp"
```
kang-makes, line 362: typo. Fix only if you are going to add another commit.

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
```diff
@@ -186,5 +186,5 @@ Returns Custom Attributes even if formatted as a json string
 {{- end -}}
 
 {{- define "newrelic.customAttributes" -}}
-{{- merge (include "newrelic.customAttributesWithoutClusterName" . | fromJson) (dict "clusterName" (include "newrelic.cluster" .)) | toJson | quote }}
+{{- merge (include "newrelic.customAttributesWithoutClusterName" . | fromJson) (dict "clusterName" (include "newrelic.cluster" .)) | toJson }}
```
kang-makes, line 189: Looks cool. This way we can do `fromJson` from other helper and dirty things like that. Love it.

kang-makes (commented):
Just pointed another nit.

charts/newrelic-infrastructure-v3/values.yaml
```diff
@@ -20,13 +20,90 @@ images:
...
+  # Mount additional volume Mounts into the pod.
+  extraVolumeMounts: []
+  resources:
+    limits:
+      memory: 300M
+    requests:
+      cpu: 100m
+      memory: 150M
+  config: {}
+
```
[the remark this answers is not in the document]
roobre: Same, but given that the controlPlane config is the largest and less comprehensible I think it's slightly better for it to be at the end. Just to keep the largest possible amount of high-level stuff in the page.

roobre (approved):
Ready to merge from my side.

Huge shoutout to @alvarocabanas for tackling this. This is an immensely complex chart and @alvarocabanas did an awesome job, both with the first iteration and also with the massive amount of comments we left here. Kudos!

kang-makes (approved):
I would love that the pipeline would test even more things and better but right now it is way more than the bare minimum to simply pass the tests.

During many and long discussions all the things that could be improved are nice-to-have features.

A week, 88 commits and 250 comments later this is a really good work. Thank you very much for you patience in this PR @alvarocabanas

PR #629, opened 2021-12-23
[newrelic-infrastructure] Add support for custom labels

drewrm:
#### Is this a new chart

No

The purpose of this change is to allow user to add custom labels, for
example to comply with policies that specify all kubernetes objects must have
standard set of labels. For example our standard requires we add a `team` label

#### Checklist

- [ x ] Chart Version bumped
- [ x ] Variables are documented in the README.md
- [ x ] Title of the PR starts with chart name (e.g. `[mychartname]`)


charts/newrelic-infrastructure/templates/_helpers.tpl
```diff
@@ -35,6 +35,15 @@ chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
 heritage: {{.Release.Service }}
 release: {{.Release.Name }}
 mode: {{ template "newrelic.mode" . }}
+{{- if .Values.global }}
+  {{- if .Values.global.customLabels }}
+{{ toYaml .Values.global.customLabels }}
+  {{- end }}
+{{- else }}
+  {{- if .Values.customLabels }}
```
gsanchezgavier, line 43: I would expect local configuration take precedence over global ones, but I see this behavior in the rest of globals here. @roobre  wdyt?

charts/newrelic-infrastructure/templates/serviceaccount.yaml
```diff
@@ -3,11 +3,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   namespace: {{ .Release.Namespace }}
-  labels:
-    app: {{ template "newrelic.name" . }}
-    chart: {{ template "newrelic.chart" . }}
-    heritage: "{{ .Release.Service }}"
-    release: "{{ .Release.Name }}"
+  labels: {{ include "newrelic.labels" . | indent 4 }}
```
gsanchezgavier, line 6: thanks for fixing this

charts/newrelic-infrastructure/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: A Helm chart to deploy the New Relic Infrastructure Agent as a DaemonSet
 name: newrelic-infrastructure
-version: 2.7.3
-appVersion: 2.8.3
+version: 2.7.4
+appVersion: 2.8.4
```
gsanchezgavier, line 5: this version should be a valid nri-kubernetes [released](https://github.com/newrelic/nri-kubernetes/releases) version 
```suggestion
appVersion: 2.8.3
```

gsanchezgavier (changes requested):
Thanks for the contribution!! There is an issue with the appVersion

charts/newrelic-infrastructure/templates/_helpers.tpl
gsanchezgavier: Could you change this logic so both labels are added, meaning global labels will be added to all charts and local labels will be added just in this chart resources.

roobre:
Closing as support for custom labels has been implemented in the v3 chart: https://github.com/newrelic/helm-charts/pull/624

PR #632, opened 2021-12-29
[All] Fix low data mode inheritance

kang-makes:
#### Is this a new chart

No.

This changes the way `lowDataMode` is inherited so you can turn up (or down) low data mode globally but the override that value at each chart level.

This is part of the effort of giving data governance to the user.

#### Checklist
- [x] Chart Version bumped


charts/newrelic-logging/ci/test-with-empty-global.yaml
```diff
@@ -0,0 +1 @@
+global: {}
```
roobre, line 1: Missing trailing newline

charts/newrelic-logging/templates/_helpers.tpl
```diff
@@ -153,12 +153,15 @@ Returns fargate
...
 */}}
 {{- define "newrelic-logging.lowDataMode" -}}
-{{- if .Values.global }}
-  {{- if .Values.global.lowDataMode }}
-    {{- .Values.global.lowDataMode -}}
-  {{- end -}}
-{{- else if .Values.lowDataMode }}
+{{/* `get` will return "" (empty string) if value is not found, and the value otherwise, so we can type-assert with kindIs */}}
+{{- if (get .Values "lowDataMode" | kindIs "bool") -}}
   {{- .Values.lowDataMode -}}
```
roobre, line 158: If `lowDataMode: false`, this will return `"false"` as a string, which is (if I'm not mistaken) evaluated as true when you do

```
{{- if include "newrelic-logging.lowDataMode" . }}
```

charts/newrelic-logging/ci/test-with-empty-global.yaml
kang-makes: Done.

charts/newrelic-logging/templates/_helpers.tpl
kang-makes: You are right. Solved.

kang-makes:
@roobre pointed an error about not templating correctly so I looked at the uses of `(include "nri-prometheus.lowDataMode" .)` and `(include "newrelic-logging.lowDataMode" .)` and I saw that `nri-prometheus` were not templating `lowDataMode` defaults if `config` in the values is empty so I fixed in this pr too as it is related to this change.

The force push is a rebase from `base`.

Ready to re-review :)

charts/newrelic-logging/templates/_helpers.tpl
```diff
@@ -153,12 +153,24 @@ Returns fargate
...
+  {{- if .Values.lowDataMode -}}
+    {{/*
+        We want only to return when this is true, returning `false` here will tamplate "false" (string) when doing
+        an `(include "newrelic-logging.lowDataMode" .)`, whis is not an "empty string" so it is `true` if it is used
+        as an evaluation somewhere else.
+    */}}
+    {{- .Values.lowDataMode -}}
+  {{- end -}}
+{{- else -}}
+{{/* This allows us to use `$global` as an empty dict directly in case `Values.global` does not exists */}}
```
roobre, line 167: I am scared of this comment inside the template, as it will render a newline.

kang-makes: No. It will not. And it was tested before committing ;)

charts/nri-prometheus/templates/configmap.yaml
```diff
@@ -8,11 +8,12 @@ apiVersion: v1
...
-{{- else }}
-    {{ toYaml .Values.config | indent 4 | trim }}
-{{- end }}
-{{- end }}
+
+  {{- $config := .Values.config -}}
+  {{- $lowDataDefault := .Files.Get "static/lowdatamodedefaults.yaml" | fromYaml -}}
+
+  {{- if (include "nri-prometheus.lowDataMode" .) }}
+    {{- mergeOverwrite .Values.config $lowDataDefault | toYaml | nindent 4 }}
```
roobre, line 16: `mergeOverwrite` gives precedence [from right to left](http://masterminds.github.io/sprig/dicts.html), meaning that `$lowDataDefault` is essentially not overrideable if set. Is this intended?

kang-makes: No. Fixed.

kang-makes:
Not that the v3 is merged, the `lowDataMode` can be enabled to the infrastructure bundle v3 too.

Ready to re-review.

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
```diff
@@ -153,6 +153,31 @@ Returns fargate
...
+    {{- .Values.lowDataMode -}}
+  {{- end -}}
+{{- else -}}
+{{/* This allows us to use `$global` as an empty dict directly in case `Values.global` does not exists */}}
+{{- $global := index .Values "global" | default dict -}}
+{{- if get $global "lowDataMode" | kindIs "bool" -}}
+  {{- if $global.lowDataMode -}}
+    {{- $global.lowDataMode -}}
+  {{- end -}}
+{{- end -}}
```
roobre, line 177: We could also just `if .Values.global`, right?
```suggestion
{{- if .Values.global -}}
  {{- if get $global "lowDataMode" | kindIs "bool" -}}
    {{- if $global.lowDataMode -}}
      {{- $global.lowDataMode -}}
    {{- end -}}
  {{- end -}}
{{- end -}}
```

charts/newrelic-infrastructure-v3/templates/control-plane/configmap.yaml
```diff
@@ -1,17 +1,16 @@
...
 metadata:
-  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "newrelic.labels" . | nindent 4 }}
   name: {{ template "newrelic.fullname" . }}-control-plane
+  namespace: {{ .Release.Namespace }}
 data:
   nri-kubernetes.yml: |
-    {{- .Values.common.config | toYaml | nindent 4 }}
+    {{- include "newrelic.commonIntegrationConfig" . | nindent 4 }}
```
roobre, line 12: This might look a bit cryptic. WDYT about making the helper return the overrides? So we can do something like:

```suggestion
    {{- merge (.Values.common.config (include "newrelic.configDefaults" . | fromYaml) ) | toYaml | nindent 4 }}
```

charts/newrelic-infrastructure-v3/values.yaml
```diff
@@ -36,15 +36,20 @@ common:
   # Configuration entries that apply to all instances of the integration: kubelet, ksm and control plane.
   config:
     # -- How often the integration should collect and report data.
-    # Intervals larger than 40s are not supported and will cause the NR UI to not behave properly.
-    interval: 15s
+    # Intervals larger than 40s are not supported and will cause the NR UI to not behave properly. This value takes precedence over `lowDataMode`
+    # but if this value is undefined its default from 15s (without low data mode) to 30s (with low data mode enabled)
+    # interval: 15s
```
roobre, line 41: We can hint `helm-docs` about the default value
```suggestion
    # Intervals larger than 40s are not supported and will cause the NR UI to not behave properly.
    # @default -- `15s` if `lowDataMode == false`, `30s` otherwise.
    # interval: 15s
```

charts/newrelic-infrastructure-v3/values.yaml
```diff
@@ -36,15 +36,20 @@ common:
...
     # -- Timeout for the different APIs contacted by the integration: Kubernetes, KSM, Kubelet, etc.
     timeout: 30s
   # -- Config for the Infrastructure agent.
   # Will be used by the forwarder sidecars and the agent running integrations.
   # See: https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings/
   agentConfig: {}
 
+# -- Enables low data mode sending less data by incrementing the interval from "15s" (the default when `lowDataMode` is `false` or `nil`) to "30s".
+# Can be superseded by `.common.config.interval`
+# lowDataMode: false
```
roobre, line 51:
```suggestion
# lowDataMode -- (bool) Enables low data mode sending less data by incrementing the interval from "15s" (the default when `lowDataMode` is `false` or `nil`) to "30s".
# Can be superseded by `.common.config.interval`
# @default -- false
# lowDataMode: false
```

roobre (commented):
Looking great, just some nits on helm-docs comments and a suggestion to make the config a bit more explicit.

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
kang-makes: I thought that with a variable it was clearer. Fixed.

charts/newrelic-infrastructure-v3/templates/control-plane/configmap.yaml
kang-makes: I am like against of having "logic" in the template and just return from the helper what is has to be in the template. That is the reason I did that. But I have to say that your approach seems cleaner and easier to understand.

Done.

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
roobre: I think in "real code" is clearer, but in templates... Variables in templates feel wrong to me.

Just a personal thought tho. Thanks!

roobre: EDIT: My suggestion is actually wrong, it would not allow for the `else` to evaluate as it should. Please ignore it.

roobre (approved):
Just tested all the combinations I could think of and behaved perfectly fine, awesome job.

Remember to run helm-docs again after the changes we made to the doc comments :)

PR #634, opened 2021-12-30
[nri-kubernetes] NRI kubernetes v3 upgrade shimming

kang-makes:
#### Is this a new chart

No

This is other approach that enables us to leave the compatibility layer separated from the code we really want to maintain.

#### Checklist
- [x] Chart Version bumped


kang-makes:
Rebase is needed but it is ready to start a discussion.

charts/nri-bundle/values.yaml
```diff
@@ -2,10 +2,14 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-# enables newrelic-infrastructure
+# Legacy nri-kubernetes chart, disabled by default
 infrastructure:
   enabled: true
```
roobre, line 7: I'd flip this to false, so new users don't have any hacks enabled by default.

charts/nri-bundle/values.yaml
```diff
@@ -2,10 +2,14 @@
...
 
+# Latest nri-kubernetes chart
+nri-kubernetes:
+  enabled: false
```
roobre, line 11: And this enabled by default.

charts/nri-bundle/requirements.yaml
```diff
@@ -1,9 +1,26 @@
...
     version: 2.7.3
 
+  # New nri-kubernetes chart with compatibility layer.
+  - name: nri-kubernetes
+    alias: newrelic-infrastructure
+    repository: file://../nri-kubernetes
+    # Notes of why this is overengineered:
+    #  * This chart is still using Helm 2 so we cannot set various conditions like `infrastructure.enabled,newrelic-infrastructure.compatibilityLayer`
+    #  * Helm 2 also does not support `import-values` so we have to alias the new chart to the old one to get the values from it.
+    condition: newrelic-infrastructure.compatibilityLayer
```
roobre, line 15: If we don't set this to `false` (or `true`) in the values, helm will ignore the condition and render the chart.

kang-makes: Really good point. I would have to think about it.

kang-makes:
Limitations of having Helm 2, a change in the name of the chart and a change in the values are too complex to a compatibility layer.

We cannot tackle all in the same release. What I tried with this PR is impossible.

Closing.

PR #643, opened 2022-01-06
[nri-prometheus] Fix nri-prometheus-config to include config.transformation content in lowDataMode.

shawnj:
#### Is this a new chart
No

This fix merges the `transformations` content from the `lowDataDefault` yaml with the `config.transformations` content from the default `.Values.config.transformations` content in the `values.yaml` file.

Current configuration overwrites the `.Values.config.transformation` content with the content from the `lowDataDefault` yaml content when `lowDataMode` is `true`.

#### Notes for reviewer:
Needed to use a "manual" approach for merging `transformation` content as the helm `merge` function will not merge different object types.

#### Checklist
- [x] Chart Version bumped


charts/nri-prometheus/templates/configmap.yaml
```diff
@@ -9,10 +9,17 @@ data:
   config.yaml: |
     cluster_name: {{ include "nri-prometheus.cluster" . }}
 {{- if .Values.config }}
-{{- if (include "nri-prometheus.lowDataMode" .) }}
-{{ $lowDataDefault := .Files.Get "static/lowdatamodedefaults.yaml" | fromYaml -}}
-    {{- mergeOverwrite .Values.config $lowDataDefault | toYaml | indent 4 -}}
-{{- else }}
-    {{ toYaml .Values.config | indent 4 | trim }}
+    {{ omit .Values.config "transformations" | toYaml | indent 4 | trim -}}
```
kang-makes, line 12: I would `trim` before `indent` and use `{{-` with `nindent` so it will clean all the YAML and do the indentantion properly an consistently.

charts/nri-prometheus/templates/configmap.yaml
```diff
@@ -9,10 +9,17 @@ data:
...
+    transformations: 
+    {{ .Values.config.transformations | toYaml | indent 4 | trim -}}
+{{- end }}
+{{- if (include "nri-prometheus.lowDataMode" .) -}}
+    {{ $lowDataDefault := .Files.Get "static/lowdatamodedefaults.yaml" | fromYaml }}
+    {{ if .Values.config.transformations }}
+    {{- $lowDataDefault.transformations | toYaml | indent 4 | trim }}
+    {{ else }}
+    {{- $lowDataDefault | toYaml | indent 4 | trim }}
+    {{ end }}
```
kang-makes, line 23: This could lead to two `transformations` in case of `transformations` in the config and `lowDataMode` enabled.

A `merge` with the defaults (in case of `lowDataMode`) with the one that came from `.Values.configtransformations` should be needed or a way to allow overriding `transformations` properly.

kang-makes (changes requested):
There is a possibility that `transformations` could be declared twice making an invalid configuration file. Could I read the template code incorrectly?

kang-makes:
Hhmm... Maybe it is correct and I am reading it incorrectly but a helper could be awesome so we have this obscure logic more readable...

Let me recheck it.

shawnj:
@kang-makes I've added a `mergeTransformations` function to `helpers` to hopefully clear some of the confusion.  Also, I tested the various configurations and since I'm omitting `transformations` from the original config and replacing it there are no duplicate `transformation` sections in the config.

I've also remove the `trim` and cleaned up the whitespace a bit.

kang-makes:
I would like to merge this by myself but I cannot because of a couple of reasons:
- Next time, instead for merging from our `master` branch is better to `rebase` because we can replay the commits easier and I can take ownership or you branch maintaining your authorship of the commits.
- Instead of using you `master` branch on your fork, use other branch to make the rebase easier too.

I would like to take ownership of this and bump the chart version because the `master` version has gone forward this PR but I cannot because of these issues.

Can you bump the version so I can merge it? It would be a shame to lose all this work already done.

shawnj:
@kang-makes Appreciate the guidance on branching for future PR's.  I have bumped the fix version for the `nri-prometheus` chart.

kang-makes (approved):
LGTM

PR #644, opened 2022-01-10
[newrelic-infrastructure-v3] compatibility layer

### LINKED ISSUE #636: infrastructure v3: Implenent the compatibility layer
To ease the transition between `nri-kubernetes` v2 and v3 we will add a set of helpers that translate or validate values form the old to the new chart. So far we have two approaches, both of which I think acceptable:
1. Rename the chart and not add a compatibility helper
2. Keep the old name for the chart and add a compatibility helper

### Option 1 The discarded one
Rename and avoid helper:
We would add the new chart with a new name. In the bundle chart, a helper would perform very basic sanity checks:
- Fail if both old and new are enabled (unlikely to happen)
- Fail if the old chart has any kind of config that does not have an equivalent in the new one.

This makes the upgrade process safe: We will fail and point the user to a simple migration guide if they have complex config. For simple things, like the privileged flag, we check they are equal and go ahead. If more complex options are set, we just fail and point to the migration guide.
Here we assume that most customers will not have custom values. If this is the case, the migration does not need intervention in the happy path. (edited) 

### Option 2 The chosen one

Keep name and add helper:
- We add the new chart, `newrelic-infrastructure-beta` aliased as `newrelic-infrastructure`.
- We add a helper that translates the values and emits a deprecation warning.
In the future, we rename `newrelic-infrastructure-beta` with `newrelic-infrastructure`, making the change breaking after some time.
This will make us stick with the old name.

paologallinaharbur:
#### Is this a new chart
no

 It adds a compatibility layer for what makes sense to keep and add notes into Notes.TXT for the ones that makes sense to directly deprecate

  - fix #636

I tried to keep all compatibility "hacks" in the helpers to be able to easily get rid of them in the future

The following values were **NOT** translated
 - nodeSelector, nodeAffinity, resources, tolerations, image. These were not migrated since there was a 1to3 mapping possibly causing issues
 - Windows related. Windows will be not supported for now
 - apiServerSecurePort , etcdTlsSecretName and etcdTlsSecretNamespace. The structure is completely different and adding it would cause a lot of burden
 - logFile, eventQueueDepth we do not want to expose agent options via ENV var. We could merge the config file, but I would avoid it if possible
 - runAsUser We are passing directly the security context. In strict environment this is a must have, for now there is a warning, but we can include it in the future. Morover it as a default that makes difficult to decide which value to pass

**EDIT**
Added  to translated values also:
- eventQueueDepth, 
- runAsUser, 
- nodeAffinity, 
- nodeSelector

#### Checklist
- [x] Chart Version bumped


charts/newrelic-infrastructure-v3/templates/NOTES.txt
```diff
@@ -13,6 +13,149 @@ Those components are needed to have the full experience on NROne kubernetes expl
...
+###########
+# WARNING #
+###########
+
+You have specified "nodeSelector" into values, however the new version of the chart
+support different nodeSelectors per each component, therefore the value specified has been ignored.
+
+Please use ksm.nodeSelector, controlPlane.nodeSelector and kubelet.nodeSelector.
+
+{{- end }}
```
roobre, line 27: A cheap alternative could be to emit this warning but render the legacy `nodeSelector` for all three components. Do you see any potential drawback for this?

charts/newrelic-infrastructure-v3/templates/NOTES.txt
```diff
@@ -13,6 +13,149 @@ Those components are needed to have the full experience on NROne kubernetes expl
...
+###########
+# WARNING #
+###########
+
+You have specified "nodeAffinity" into values, however the new version of the chart
+support different nodeAffinity options per component, therefore the value specified has been ignored.
+
+Please use ksm.affinity.nodeAffinity, controlPlane.affinity.nodeAffinity and kubelet.affinity.nodeAffinity.
+
+{{- end }}
```
roobre, line 40: Same as above

charts/newrelic-infrastructure-v3/templates/NOTES.txt
```diff
@@ -13,6 +13,149 @@ Those components are needed to have the full experience on NROne kubernetes expl
...
+###########
+# WARNING #
+###########
+
+You have specified "tolerations" into values, however the new version of the chart
+support different tolerations per component, therefore the value specified has been ignored.
+
+Please use ksm.tolerations, controlPlane.tolerations and kubelet.tolerations.
+
+{{- end }}
```
roobre, line 66: Same as above

charts/newrelic-infrastructure-v3/templates/NOTES.txt
```diff
@@ -13,6 +13,149 @@ Those components are needed to have the full experience on NROne kubernetes expl
...
+
+{{- if .Values.runAsUser }}
+
+###########
+# WARNING #
+###########
+
+The "runAsUser" value is no longer supported, please specify it with the value "securityContext.runAsUser".
+
+{{- end }}
```
roobre, line 137: Seems like this should be shim-able with ease, is there any reason to leave it out?

charts/newrelic-infrastructure-v3/templates/NOTES.txt
```diff
@@ -13,6 +13,149 @@ Those components are needed to have the full experience on NROne kubernetes expl
...
+
+{{- if .Values.eventQueueDepth }}
+
+###########
+# WARNING #
+###########
+
+The "eventQueueDepth" value is no longer supported, please specify the option "event_queue_depth" directly in "common.agentConfig".
+
+{{- end }}
```
roobre, line 147: This could also be shim-able into `common.agentConfig`

charts/newrelic-infrastructure-v3/templates/NOTES.txt
```diff
@@ -13,6 +13,149 @@ Those components are needed to have the full experience on NROne kubernetes expl
...
+
+###########
+# WARNING #
+###########
+
+The "apiServerSecurePort" value is no longer supported, please specify it in the section "apiServer.autodiscover[].endpoints".
+
+{{- end }}
+
 {{- if and (include "newrelic.fargate" .) .Values.kubelet.affinity.nodeAffinity }}
```
roobre, line 159: New chart is not compatible with fargate (yet), I'd add a clause to fail if it is specified
```
{{- if include "newrelic.fargate" . }}
Fargate is not supported for the v3 chart, please use v2 instead.
{{- fail "Fargate is not supported for the v3 chart, please use v2 instead" }}
{{- end }}

File: charts/newrelic-infrastructure-v3/templates/control-plane/daemonset.yaml [hunk 7]
```diff
@@ -192,10 +192,20 @@ spec:
       {{- with .Values.priorityClassName }}
       priorityClassName: {{ . }}
       {{- end }}
-      {{- with .Values.controlPlane.affinity.nodeAffinity }}
+      {{- if .Values.controlPlane.affinity.nodeAffinity }}
       affinity:
         nodeAffinity:
-          {{- . | toYaml | nindent 10 }}
+          {{- .Values.controlPlane.affinity.nodeAffinity | toYaml | nindent 10 }}
+      {{- else if include "newrelic.fargate" . }}
```
roobre, line 199: Is it possible to scrape the EKS control plane on Fargate? If it isn't, we could get rid of the fargate logic here.

charts/newrelic-infrastructure-v3/templates/ksm/configmap.yaml
```diff
@@ -1,4 +1,5 @@
-{{- if .Values.ksm.enabled -}}
+{{- $ksmEnabled := ((.Values.ksm.enabled | default (not .Values.ksm.disableKubeStateMetrics))) -}}
```
roobre, line 1: Even if redundant, I'd suggest to wrap this into a helper so we can later identify it as such. Otherwise we might forget that `.Values.ksm.disableKubeStateMetrics` was in fact legacy :)

roobre: Moreover, this line might not behave as the user expect. On a new install, `.Values.ksm.enabled` will default to `true`, effectively overriding an old `.Values.ksm.disableKubeStateMetrics: false`. I think we might need to wrap the logic in a helper to process priority as it should.

charts/newrelic-infrastructure-v3/templates/ksm/deployment.yaml
```diff
@@ -1,4 +1,5 @@
-{{- if (.Values.ksm.enabled) }}
+{{- $ksmEnabled := ((.Values.ksm.enabled | default (not .Values.ksm.disableKubeStateMetrics))) -}}
```
roobre, line 1: Ditto

roobre (commented):
Left some nits and suggestions of things that could be in scope.

More importantly, I think we could simplify some helpers by using a pattern like the following example, for the agent config:

```yaml
data:
  newrelic-infra.yml: |-
    {{- .Values.common.agentConfig | mustMergeOverwrite (include "newrelic.compatibility.agentConfig" . | fromYaml) | toYaml | nindent 4 -}}
{{- end -}}
```

Looks a bit confusing at first with all the `toYaml|fromYaml|toYaml|indent` but I like how the compatibility layer becomes an "overlay" that is applied on top of the defaults.

This might not work for all cases, in which it might result in conflicting options, but I think it could simplify some cases. Moreover removing the overlay later becomes trivial and we don't need to fear introducing any bugs.
WDYT?

charts/newrelic-infrastructure-v3/templates/NOTES.txt
paologallinaharbur: no drawback in this case, I've added that

charts/newrelic-infrastructure-v3/templates/NOTES.txt
paologallinaharbur: nodeAffinity are already present in default in case of control plane. We can give precedence to legacy value.
I am not totally sure we should give precedence to an "old" config that was thought for a different object..
WDYT?

charts/newrelic-infrastructure-v3/templates/NOTES.txt
paologallinaharbur: tolerations are already present in default. We can give precedence to legacy value.
I am not totally sure we should give precedence to an "old" config that was thought for a different object..
WDYT?

charts/newrelic-infrastructure-v3/templates/NOTES.txt
paologallinaharbur: mainly time, added it

charts/newrelic-infrastructure-v3/templates/NOTES.txt
paologallinaharbur: Added it, it was left aside just for time constrain

charts/newrelic-infrastructure-v3/templates/NOTES.txt
paologallinaharbur: why it is not? I would have said that this chart is as compatible as the old one.

charts/newrelic-infrastructure-v3/templates/ksm/configmap.yaml
paologallinaharbur: true, I'll take care of that

charts/newrelic-infrastructure-v3/templates/NOTES.txt
roobre: 🤔 Fair point, I don't like either option:
- If we ignore the old value, we will be deploying kubelet in pods that were not meant to be monitored
- If we honor the old value, we can also deploy the CP daemonset in a bunch of worker nodes

What about applying the legacy affinity/selector to kubelet/ksm deployments only? Keep the default one for the control plane DS (avoid overriding it) and emit a warning.
It increases complexity slightly but I think it could be a good approach.

charts/newrelic-infrastructure-v3/templates/NOTES.txt
roobre: My thought was that were still injecting the "old" image as a sidecar, but it's true that the "rest" of the integration should work well 🤔

charts/newrelic-infrastructure-v3/templates/NOTES.txt
paologallinaharbur: I kept it as unsupported for now

charts/newrelic-infrastructure-v3/templates/NOTES.txt
paologallinaharbur: did that

charts/newrelic-infrastructure-v3/templates/agent-configmap.yaml
```diff
@@ -1,4 +1,4 @@
-{{ if .Values.common.agentConfig }}
+{{- if (include "newrelic.compatibility.agentConfig" . ) -}}
```
paologallinaharbur, line 1: avoiding for the multiple if "(include "newrelic.compatibility.agentConfig" . )" the whole toYaml| FromYaml etc, we can add it, but I see it too verbose

paologallinaharbur: I added it in the other cases

charts/newrelic-infrastructure-v3/tests/configmap_agent_test.yaml
```diff
@@ -0,0 +1,63 @@
+suite: test configmap
+templates:
+  - templates/agent-configmap.yaml
+tests:
```
gsanchezgavier, line 4: I like this unit tests :)

gsanchezgavier: what do you think to move the compatibility related test to an specific folder?

gsanchezgavier (commented):
good job! just left the control plane concern and nits

charts/newrelic-infrastructure-v3/README.md
```diff
@@ -44,7 +44,7 @@ Kubernetes: `>=1.16.0-0`
 | images.forwarder.repository | string | `"newrelic/k8s-events-forwarder"` | Image for the agent sidecar. |
 | images.forwarder.tag | string | `"1.20.5"` | Tag for the agent sidecar. |
 | images.integration.repository | string | `"newrelic/nri-kubernetes"` | Image for the kubernetes integration. |
-| images.integration.tag | string | `"0.1.2"` | Tag for the kubernetes integration. |
+| images.integration.tag | string | `"0.2.0-pre"` | Tag for the kubernetes integration. |
```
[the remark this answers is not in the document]
paologallinaharbur: true

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
```diff
@@ -188,3 +196,121 @@ Returns Custom Attributes even if formatted as a json string
...
+*/}}
+{{- define "newrelic.compatibility.control-plane" -}}
+enabled: true
+{{- with .autodiscover }}
+autodiscover:
+{{ . | toYaml}}
+{{- end -}}
+
+{{- if ( or .staticEndpoint .etcdEndpointUrl) }}
+staticEndpoint:
```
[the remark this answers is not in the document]
paologallinaharbur: Discussed offline, translating CP components seems too complex and error prone since the logic is different

charts/newrelic-infrastructure-v3/tests/configmap_agent_test.yaml
paologallinaharbur: done

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
```diff
@@ -188,3 +196,121 @@ Returns Custom Attributes even if formatted as a json string
 {{- define "newrelic.customAttributes" -}}
 {{- merge (include "newrelic.customAttributesWithoutClusterName" . | fromJson) (dict "clusterName" (include "newrelic.cluster" .)) | toJson }}
 {{- end -}}
+
+
+{{/*
+Returns controlPlane component taking into account legacy values
+*/}}
```
kang-makes, line 203: Helm does not template [anything starting with `_`](https://helm.sh/docs/chart_template_guide/named_templates/#partials-and-_-files) so it could be a good idea to separate the helpers into different files (or the compatibility layer to other file so it could be removed more easily in the future.

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
```diff
@@ -188,3 +196,121 @@ Returns Custom Attributes even if formatted as a json string
...
+{{- define "newrelic.compatibility.control-plane" -}}
+enabled: true
+{{- with .autodiscover }}
+autodiscover:
+{{ . | toYaml}}
```
kang-makes, line 208: Does this not need `indent`?

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
```diff
@@ -188,3 +196,121 @@ Returns Custom Attributes even if formatted as a json string
...
+    type: bearer
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Returns true if .Values.ksm.enabled is true and the legacy disableKubeStateMetrics is not set
+*/}}
+{{- define "newrelic.compatibility.ksm.enabled" -}}
+{{- and .Values.ksm.enabled (not .Values.disableKubeStateMetrics) -}}
```
kang-makes, line 228: Be careful with returning `false`. I had this issue before. `"false"` is `true` in helm/sprig because it is not a empty string: https://github.com/newrelic/helm-charts/pull/632/files#diff-f02666f6644eee28f931b5c6506b680155d232a8617ca542cd4add2be0b56720R158-R163

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
```diff
@@ -188,3 +196,121 @@ Returns Custom Attributes even if formatted as a json string
...
+
+{{/*
+Returns integration configmap data,
+if the new one is defined we ignore the old one
+*/}}
+{{- define "newrelic.integrations" -}}
+{{- if .Values.integrations_config -}}
+{{- range .Values.integrations_config }}
+{{ .name -}}: |-
+  {{- toYaml .data | nindent 2 -}}
```
kang-makes, line 298:
```suggestion
  {{- toYaml .data | nindent 2 }}
```

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
```diff
@@ -188,3 +196,121 @@ Returns Custom Attributes even if formatted as a json string
...
+Returns integration configmap data with legacy fallback
+*/}}
+{{- define "newrelic.compatibility.integrations" -}}
+{{- if (include "newrelic.integrations" .) -}}
+{{- include "newrelic.integrations" . -}}
+
+{{- else if .Values.integrations -}}
+{{- range $k, $v := .Values.integrations }}
+{{ $k | trimSuffix ".yaml" | trimSuffix ".yml" }}.yaml: |-
+    {{- tpl ($v | toYaml) $ | nindent 2 -}}
```
kang-makes, line 313:
```suggestion
    {{- tpl ($v | toYaml) $ | nindent 2 }}
```

kang-makes (commented):
I think that these changes could improve templating, readability and a few corener cases where this chart could fail.

Take some time and let me know what you think :)

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
paologallinaharbur: done!

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
paologallinaharbur: sure, makes sense!

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
paologallinaharbur: without it it was rending the list as:
```
autodiscover:
- item 1
- item2
```

cleaner with indentation, you are right

charts/newrelic-infrastructure-v3/tests/configmap_agent_test.yaml
paologallinaharbur: actually I noticed that tests are ignored if placed in a separate folder 😖

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
paologallinaharbur: however there is a test covering that use case checking that the config map is not created

paologallinaharbur: did as you said, way cleaner

paologallinaharbur:
after talking with Guillermo I had to remove the "translation" for control plane URL. We cannot map them with staticURLs due to possible data duplication, and neither with autodiscover since we do not know the label selector.
We could add multiple entries per each value, but it become way more complex

charts/newrelic-infrastructure-v3/values.yaml
```diff
@@ -24,8 +24,8 @@ images:
   # Image for the New Relic Kubernetes integration.
   integration:
     pullPolicy: IfNotPresent
-    # -- Tag for the kubernetes integration.
-    tag: 0.1.2
+    # -- Tag for the kubernetes integration. Currently, set from the appVersion.
+    tag: ""
```
kang-makes, line 28: Are we sure that this is going to be parsed by renovate? @roobre

Renovate by default will update [helm dependencies](https://docs.renovatebot.com/modules/manager/helmv3/) from `Chart.yaml` and [image version](https://docs.renovatebot.com/modules/manager/helm-values/) from `values.yaml`

I don't know if this change will need further configuration.

kang-makes (commented):
Overall is more than correct. Great work! I pointed a few nits (one not so "not important tho'")

charts/newrelic-infrastructure-v3/values.yaml
paologallinaharbur: I did not think to renovate honestly 😓 
I was just annoyed to have 3 places to update a version -> Readme, values, appVersion in chart.yaml...

If renovate takes care of that I am more than happy to rollback the change 😄

paologallinaharbur: Or nevertheless I can take it out, it it outofscope with respect to the PR

charts/newrelic-infrastructure-v3/templates/control-plane/configmap.yaml
```diff
@@ -8,11 +8,28 @@ metadata:
...
+
+    {{- if  .Values.controlPlane.config.controllerManager.enabled }}
+      controllerManager:
+        {{- toYaml .Values.controlPlane.config.controllerManager | nindent 8 -}}
+    {{- end -}}
+
+    {{- if .Values.controlPlane.config.apiServer.enabled }}
+      apiServer:
+        {{- toYaml .Values.controlPlane.config.apiServer | nindent 8 -}}
+    {{- end -}}
```
roobre, line 34: Is there a reason to expand this by hand?

charts/newrelic-infrastructure-v3/templates/kubelet/daemonset.yaml
```diff
@@ -51,9 +51,9 @@ spec:
           image: "{{ .Values.images.integration.registry }}/{{ .Values.images.integration.repository }}:{{ .Values.images.integration.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.images.integration.pullPolicy }}
           securityContext:
-            {{- toYaml .Values.securityContext | nindent 12 }}
+            {{- .Values.securityContext | mustMergeOverwrite (include "newrelic.compatibility.securityContext" . | fromYaml) | toYaml | nindent 12 }}
           env:
-            - name: "NRI_KUBERNETES_HTTPSERVERPORT"
+            - name: "NRI_KUBERNETES_SINK_HTTP_PORT"
```
roobre, line 56: Is this change intended?

charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
```diff
@@ -0,0 +1,79 @@
...
+{{- end -}}
+{{- if  .Values.kubeStateMetricsNamespace }}
+namespace: {{ .Values.kubeStateMetricsNamespace}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Returns the new value if available, falling back on the legacy one
+*/}}
+{{- define "newrelic.compatibility.valueWithFallback" -}}
```
roobre, line 35: Smart, I like it!

charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
```diff
@@ -0,0 +1,79 @@
...
+{{- toYaml .legacy}}
+{{- end }}
+{{- end -}}
+
+{{/*
+Returns securityContext merged with old runAsUser config
+*/}}
+{{- define "newrelic.compatibility.securityContext" -}}
+{{- if  .Values.runAsUser -}}
+{{- mustMergeOverwrite .Values.securityContext (dict "runAsUser" .Values.runAsUser ) | toYaml }}
```
roobre, line 48: We are already calling `mustMergeOverwrite` outside of the helper, do we need to call it inside as well?

charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
```diff
@@ -0,0 +1,79 @@
...
+{{- else -}}
+{{- .Values.securityContext | toYaml }}
```
roobre, line 50: Same as above, we are already merging outside right?

charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
```diff
@@ -0,0 +1,79 @@
...
+Returns agent configmap merged with old eventQueueDepth config
+*/}}
+{{- define "newrelic.compatibility.agentConfig" -}}
+{{ $config:= (include "newrelic.compatibility.valueWithFallback" (dict "legacy" .Values.config "supported" .Values.common.agentConfig ) | fromYaml )}}
+{{- if .Values.eventQueueDepth -}}
+{{- mustMergeOverwrite $config (dict "event_queue_depth" .Values.eventQueueDepth ) | toYaml }}
+{{- else -}}
+{{- $config | toYaml}}
+{{- end -}}
+{{- end -}}
```
roobre, line 64: This seems a bit complex to me.
Why not just returning the shimmed yaml?
```suggestion
{{- define "newrelic.compatibility.agentConfig" -}}
{{- with .Values.eventQueueDepth -}}
event_queue_depth: {{ . | quote }}
{{- end -}}
{{- end -}}
```

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
```diff
@@ -214,10 +222,25 @@ Returns Custom Attributes even if formatted as a json string
...
+if the new one is defined we ignore the old one
+*/}}
+{{- define "newrelic.integrations" -}}
+{{- if .Values.integrations_config -}}
+{{- range .Values.integrations_config }}
+{{ .name -}}: |-
+  {{- toYaml .data | nindent 2 }}
+{{- end -}}
+{{- end -}}
+{{- end -}}
```
roobre, line 236: Do we need a helper for this? We only render it in one place.

charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
```diff
@@ -0,0 +1,79 @@
...
+{{- define "newrelic.compatibility.integrations" -}}
+{{- if (include "newrelic.integrations" .) -}}
+{{- include "newrelic.integrations" . -}}
+
+{{- else if .Values.integrations -}}
+{{- range $k, $v := .Values.integrations }}
+{{ $k | trimSuffix ".yaml" | trimSuffix ".yml" }}.yaml: |-
+    {{- tpl ($v | toYaml) $ | nindent 2 }}
+{{- end -}}
+{{- end -}}
```
roobre, line 78: Other values are merged while these ones are replaced. Could we merge these as well? And just shim the old format to the new:
```suggestion
{{- range .Values.integrations_config }}
{{ .name }}:
  {{- toYaml .data | nindent 2 }}
{{- end -}}
{{- end -}}
```

And then we just `mustMerge`  .Values.integrations with `include "newrelic.compatibility.integrations"` in the config map.

charts/newrelic-infrastructure-v3/templates/agent-configmap.yaml
roobre: You are right, copying the whole to yaml from yaml merge would be too verbose.

However perhaps we can get rid of this `if` althogether, and render the configMap unconditionally (empty). Would this be a problem?

charts/newrelic-infrastructure-v3/values.yaml
roobre: Renovate should indeed take care, I'd keep the tag in there :)
If we see that the `appVersion` is not handled by it we can then think about something on that regard.

roobre:
One more comment regarding `NOTES.txt`: Customers with a custom `values.yml` might trigger three or four of our messages, which currently have quite some newlines and the big `WARNING` header. Could it be possible to make those shorter?

charts/newrelic-infrastructure-v3/templates/kubelet/daemonset.yaml
paologallinaharbur: yes it is needed after https://github.com/newrelic/nri-kubernetes/pull/320

charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
paologallinaharbur: we are merging three things here:
 - .Values.config
 - .Values.common.agentConfig
 - eventQueueDepth

In your example we are just returning if available "event_queue_depth"

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
paologallinaharbur: yes, I added that since we need the same code it in two different places

charts/newrelic-infrastructure-v3/templates/control-plane/configmap.yaml
paologallinaharbur: basically to avoid putting in the config useless stuff for disabled component, it is also useful for tests. If needed we can have it back as before

charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
paologallinaharbur: We cannot write directly:
```
{{- mustMergeOverwrite .Values.securityContext (dict "runAsUser" .Values.runAsUser ) | toYaml | nindent 12 }}
```
since it would result in a null in case runAsUser is not set, therefore we would need an if merging it in one case and ot in the other, making it really verbose.

```
                        Expected to equal:
                                1000
                        Actual:
                                null
                        Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,2 +1,2 @@
                                -1000
                                +null

```

We could remove easily the mustMergeOverwrite outside the helper, but we would lose the concept of overlay

charts/newrelic-infrastructure-v3/templates/_helpers.tpl
paologallinaharbur: actually they old and new config were mixed up

charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
paologallinaharbur: not at all, I was following the idea that the overlay needed to be a valid config

charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
paologallinaharbur: Moreover we need a reliable function returning the whole agent config since we are checking that in 9-10 places in code:
 - when deciding to create volumes, 
 - when adding the annotation to trigger pod restart

charts/newrelic-infrastructure-v3/templates/agent-configmap.yaml
paologallinaharbur: Actually I prefer the opposite:
 - just for the agent config keep everywhere the simple notation since it is used in many places. In this way we avoid multiple merges, different notations, empty configMaps and complexity. 
   On the down side we are not following the "Overlay pattern" for this config

charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
paologallinaharbur: I simplified it a bit, let me know what you think

paologallinaharbur: I actually followed your advice

charts/newrelic-infrastructure-v3/templates/NOTES.txt
```diff
@@ -4,13 +4,112 @@ Your deployment of the New Relic Infrastructure agent is complete if you have no
...
+
+Notice that you will need as well to set securityContext.readOnlyRootFilesystem to false or add extraVolumes to allow
+the container to write to such location
+{{- end }}
+
+{{- if .Values.apiServerSecurePort }}
 
+-> WARNING LEGACY CONFIG <-
+The "apiServerSecurePort" value is no longer supported, please specify it in the section "apiServer.autodiscover[].endpoints".
 {{- end }}
```
kang-makes, line 113: I think that what Roberto wanted is something like:
```
{{/*
Returns warning messages from the legacy compatibility layer
*/}}
{{- define "newrelic.compatibility.warningMessages" -}}
{{ if .Values.resources }}

You have specified the legacy "resources" option into values, however yada yada
{{ end }}
{{ if .Values.tolerations }}

You have specified the legacy "tolerations" yada yada
{{ end }}
{{ end -}} {{/* This goes to a NOTEX.TXT so we do not want to get rid of the last linebreak */}}
```

And then in the `NOTES.txt` something like:
```
{{- if include "newrelic.compatibility.warningMessages" . }}
#########################
# /!\ WARNING /!\
#########################

{{ include "newrelic.compatibility.warningMessages" . -}}
{{ end }}
```

kang-makes: If there are no warning it won't paint the warning message but it will paint only 1 warning for all the messages. I would also separate all the warning messages to other helper to separate the logic of the helper compability layer from the warning messages of the compatibility layer itself.

Just my two cents.

paologallinaharbur: yes, mine was a cheap implementation.
If you think that it worth invest in it the best would be to write a big condition to print the warning with a small introduction saying what is going one and the the list of issues

charts/newrelic-infrastructure-v3/values.yaml
```diff
@@ -115,6 +117,16 @@ ksm:
       cpu: 100m
       memory: 150M
   config: {}
+  # If staticUrl is specified autodiscovery is not performed and the specified URL is used
+  # staticUrl: "http://test.io:8080/metrics"
+  # If scheme is specified the scheme is not discovered and the specified one is used
+  # scheme: "http"
```
paologallinaharbur, line 123: once everything is approved, update readme and this with correct format

charts/newrelic-infrastructure-v3/templates/NOTES.txt
kang-makes: I don't care. Is a nit and this is a next version.

charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
roobre: I'd say we can still achieve this by merging:
```
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
  name: {{ template "newrelic.fullname" . }}-agent
data:
  newrelic-infra.yml: |
    {{- mustMerge .Values.common.agentConfig .Values.config (include "newrelic.compatibility.agentConfig" . | fromYaml) | toYaml | nindent 2 }}
```

The if would still be large, but I still think we can just get rid of it entirely.

charts/newrelic-infrastructure-v3/templates/NOTES.txt
```diff
@@ -4,13 +4,112 @@ Your deployment of the New Relic Infrastructure agent is complete if you have no
...
+ - images.integration.* to configure the image in charge of scraping k8s data
+{{- end }}
+
+{{- if (or .Values.enableWindows .Values.windowsOsList .Values.windowsSecurityContext .Values.windowsNodeSelector)}}
+
+-> WARNING LEGACY CONFIG <-
+Windows is currently supported by 2.x charts only and therefore "windowsOsList" "windowsSecurityContext" "windowsNodeSelector"
+are no longer supported.
+
+{{- end }}
```
roobre, line 65: If we cannot shim the values in an acceptable way I think we should fail so customer does not end up with unexpected config:

```suggestion
{{- if (or .Values.enableWindows .Values.windowsOsList .Values.windowsSecurityContext .Values.windowsNodeSelector)}}

-> WARNING LEGACY CONFIG <-
Windows is currently supported by 2.x charts only and therefore "windowsOsList" "windowsSecurityContext" "windowsNodeSelector"
are no longer supported.

{{- fail "Windows nodes are not supported in nri-kubernetes v3" }}
{{- end }}
```

charts/newrelic-infrastructure-v3/templates/NOTES.txt
```diff
@@ -4,13 +4,112 @@ Your deployment of the New Relic Infrastructure agent is complete if you have no
...
+#  autodiscover:
+#    - selector: "tier=control-plane,component=etcd"
+#      namespace: kube-system
+#      matchNode: true
+#      endpoints:
+#        - url: https://localhost:4001
+#          insecureSkipVerify: true
+#          auth:
+#            type: bearer
+
```
roobre, line 82: `{{- fail "Please migrate your control plane config to the nri-kubernetes v3 format " }}`

charts/newrelic-infrastructure-v3/templates/NOTES.txt
```diff
@@ -4,13 +4,112 @@ Your deployment of the New Relic Infrastructure agent is complete if you have no
...
+file. Example:
+# - endpoints:
+#     - url: https://localhost:9979
+#       insecureSkipVerify: true
+#       auth:
+#         type: mTLS
+#         mtls:
+#           secretName: secret-name
+#           secretNamespace: secret-namespace
+{{- end }}
```
roobre, line 98: `{{- fail "Please migrate your control plane config to the nri-kubernetes v3 format " }}`

charts/newrelic-infrastructure-v3/templates/NOTES.txt
```diff
@@ -4,13 +4,112 @@ Your deployment of the New Relic Infrastructure agent is complete if you have no
...
+The "logFile" value is no longer supported, please specify the option "log_file" directly in "common.agentConfig".
+
+Notice that you will need as well to set securityContext.readOnlyRootFilesystem to false or add extraVolumes to allow
+the container to write to such location
+{{- end }}
+
+{{- if .Values.apiServerSecurePort }}
 
+-> WARNING LEGACY CONFIG <-
+The "apiServerSecurePort" value is no longer supported, please specify it in the section "apiServer.autodiscover[].endpoints".
```
roobre, line 112: `{{- fail "Please migrate your control plane config to the nri-kubernetes v3 format " }}`

charts/newrelic-infrastructure-v3/templates/agent-configmap.yaml
roobre: Either way is fine for me, if you think getting rid of the `if` is not the best approach then definitely we cannot overlay it.

charts/newrelic-infrastructure-v3/templates/NOTES.txt
kang-makes: That will not work. That you are suggesting will show `Windows nodes are not supported in nri-kubernetes v3` without the message above.

You have to
```
{{- fail "
Windows is currently supported by 2.x charts only and therefore "windowsOsList" "windowsSecurityContext" "windowsNodeSelector"
are no longer supported.
" }}

Comment (kang-makes) [reply charts/newrelic-infrastructure-v3/templates/NOTES.txt:65]: If we are going to fail, then the nit I've put about having a helper which renders all the errors at the same time is more useful to don't hammer the user with errors showing them one by one while he is solving them.

Comment (roobre) [reply charts/newrelic-infrastructure-v3/templates/NOTES.txt:65]: Fuck, you're totally right. For some reason I though it would have been rendered anyway :(

Comment (paologallinaharbur) [reply charts/newrelic-infrastructure-v3/templates/NOTES.txt:65]: We could do some helm magic:
 - Save a variable `errors` and append there one by one all the errors and merely at the end print them at once with the fail
 
 It is a bit hackish and honestly I prefer the long message explaining what is going on, what do you think?

Comment (paologallinaharbur) [reply charts/newrelic-infrastructure-v3/templates/NOTES.txt:65]: This works:
```
{{ $errors:= "" }}

{{- if .Values.logFile }}
{{ $errors = printf "%s%s" $errors (include "newrelic.compatibility.message.log" . | nindent 2) }}
{{- end }}

{{- if .Values.apiServerSecurePort }}
{{- if .Values.logFile }}
{{ $errors = printf "%s%s" $errors (include "newrelic.compatibility.message.apiServerSecurePort" . | nindent 2) }}
{{- end }}
{{- end }}

{{- if $errors }}
{{- fail $errors }}
{{- end }}
```

Makes sense to implement it?

Comment (paologallinaharbur) [reply charts/newrelic-infrastructure-v3/templates/NOTES.txt]: I gave a shot, let me know if it makes sense. I am not failing for every condition and for some just printing a warning

Comment (roobre) [reply charts/newrelic-infrastructure-v3/templates/NOTES.txt]: Looking pretty nice to me!

**Comment by roobre** [comment]
I'll make a few style changes to make the NOTES match the ones we have in other places and we can merge this ✨

File: charts/newrelic-infrastructure-v3/README.md [hunk 34]
```diff
@@ -44,7 +44,7 @@ Kubernetes: `>=1.16.0-0`
 | images.forwarder.repository | string | `"newrelic/k8s-events-forwarder"` | Image for the agent sidecar. |
 | images.forwarder.tag | string | `"1.20.5"` | Tag for the agent sidecar. |
 | images.integration.repository | string | `"newrelic/nri-kubernetes"` | Image for the kubernetes integration. |
-| images.integration.tag | string | `"0.1.2"` | Tag for the kubernetes integration. |
+| images.integration.tag | string | `Matching <Chart.appVersion>` | Tag for the kubernetes integration. |
```
roobre, line 47: Is this true in the end? From what I can see, we have `default | .Chart.AppVersion` but that will never evaluate as we populate a tag by default.

roobre (approved):
LGTM, please remember to run `helm-docs` to update the readme before merging :)

charts/newrelic-infrastructure-v3/README.md
paologallinaharbur: I was waiting for helm-docs to fix it

PR #647, opened 2022-01-12
newrelic-infrastructure-v3: add a separate hostNetwork toggle for control plane monitoring

roobre:
Previously, if `privileged` was set to `false`, we deployed the control plane DaemonSet with `hostNetwork: false`. However, this will almost certainly cause it not to work as intended.

This PR:
- Adds a way to deploy the DaemonSet with `hostNetwork: true` even when `privileged: false` (disabled by default)
- Warns the user and encourages them to set this flag if control plane monitoring is enabled

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure-v3/README.md
+++ b/charts/newrelic-infrastructure-v3/README.md
@@ -23,8 +23,8 @@ Kubernetes: `>=1.16.0-0`
 |-----|------|---------|-------------|
 | common | object | See `values.yaml` | Config that applies to all instances of the solution: kubelet, ksm, control plane and sidecars. |
 | common.agentConfig | object | `{}` | Config for the Infrastructure agent. Will be used by the forwarder sidecars and the agent running integrations. See: https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings/ |
-| common.config.interval | duration | `15s` if `lowDataMode == false`, `30s` otherwise. | Intervals larger than 40s are not supported and will cause the NR UI to not behave properly. |
-| controlPlane | object | See `values.yaml` | Configuration for the DaemonSet that collects metrics from the control plane. |
+| common.config.interval | duration | `15s` if `lowDataMode == false`, `30s` otherwise. | Intervals larger than 40s are not supported and will cause the NR UI to not behave properly. Any non-nil value will override the `lowDataMode` default. |
+| controlPlane | object | See `values.yaml` | Configuration for the control plane scraper. |
 | controlPlane.affinity | object | Deployed only in master nodes. | Affinity for the control plane DaemonSet. |
 | controlPlane.config.apiServer | object | Common settings for most K8s distributions. | API Server monitoring configuration |
 | controlPlane.config.apiServer.enabled | bool | `true` | Enable API Server monitoring |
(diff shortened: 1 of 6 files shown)


charts/newrelic-infrastructure-v3/README.md
```diff
@@ -50,8 +48,22 @@ Kubernetes: `>=1.16.0-0`
...
+| kubelet.extraVolumeMounts | list | `[]` |  |
+| kubelet.extraVolumes | list | `[]` |  |
+| kubelet.nodeSelector | object | `{}` |  |
+| kubelet.resources.limits.memory | string | `"300M"` |  |
+| kubelet.resources.requests.cpu | string | `"100m"` |  |
+| kubelet.resources.requests.memory | string | `"150M"` |  |
+| kubelet.tolerations[0].effect | string | `"NoSchedule"` |  |
+| kubelet.tolerations[0].operator | string | `"Exists"` |  |
+| kubelet.tolerations[1].effect | string | `"NoExecute"` |  |
+| kubelet.tolerations[1].operator | string | `"Exists"` |  |
```
roobre, line 66: Someone forgot to update the README here :(

kang-makes: Don't be passive-agressive when you know it was me... Fixing that was out of scope of that PR.

There is it still has the issue that commented variables are not templated so you have eliminated `lowDataMode` and `common.config.interval`. That was the reason I did the README manually.

roobre: Let me see if I can whisper to `helm-docs` so it works nicely.

roobre: Restored, please take a look @kang-makes :)

charts/newrelic-infrastructure-v3/README.md
```diff
@@ -21,11 +21,9 @@ Kubernetes: `>=1.16.0-0`
...
 |-----|------|---------|-------------|
-| lowDataMode | bool | `nil` | Enables low data mode sending less data by incrementing the interval from "15s" (the default when `lowDataMode` is `false` or `nil`) to "30s". Can be superseded by `.common.config.interval` |
 | common | object | See `values.yaml` | Config that applies to all instances of the solution: kubelet, ksm, control plane and sidecars. |
 | common.agentConfig | object | `{}` | Config for the Infrastructure agent. Will be used by the forwarder sidecars and the agent running integrations. See: https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings/ |
-| common.config.interval | string | `nil` | How often the integration should collect and report data. Intervals larger than 40s are not supported and will cause the NR UI to not behave properly. This value takes precedence over `lowDataMode` but if this value is undefined its default from 15s (without low data mode) to 30s (with low data mode enabled) |
-| common.config.timeout | string | `"30s"` | Timeout for the different APIs contacted by the integration: Kubernetes, KSM, Kubelet, etc. |
```
roobre, line 28: Getting rid of `timeout` is expected as I don't think it is relevant enough to be exposed in the README.

paologallinaharbur: also because that config is currently not leveraged 😄  https://github.com/newrelic/nri-kubernetes/issues/322

charts/newrelic-infrastructure-v3/templates/control-plane/_helpers.tpl
```diff
@@ -0,0 +1,8 @@
+{{/* Returns whether the controlPlane component needs needsHostNetwork to work, that is, if they do not have a staticEndpoint configured */}}
```
paologallinaharbur, line 1: "if they do not have a staticEndpoint configured"

A nit, the description seems to suggest what we are going to check, but is it not what we check.

Moreover They should not have "a" staticEndpoint, but all static endpoints for the enabled components

charts/newrelic-infrastructure-v3/values.yaml
```diff
@@ -116,11 +117,14 @@ ksm:
       memory: 150M
   config: {}
 
-# -- Configuration for the DaemonSet that collects metrics from the control plane.
+# controlPlane -- Configuration for the DaemonSet that collects metrics from the control plane.
```
paologallinaharbur, line 120: theoretically it could be also a deployment since we expose the type

roobre: Fixed!

charts/newrelic-infrastructure-v3/values.yaml
```diff
@@ -116,11 +117,14 @@ ksm:
...
-# -- Configuration for the DaemonSet that collects metrics from the control plane.
+# controlPlane -- Configuration for the control plane scraper.
 # @default -- See `values.yaml`
 controlPlane:
-  # -- Deploy control plane monitoring DaemonSet. Requires `privileged` mode to be enabled.
+  # -- Deploy control plane monitoring component.
   enabled: true
+  # -- Run Control Plane scraper with `hostNetwork` even if `privileged` is set to false.
+  # `hostNetwork` is required for most control plane configurations, as they only accept connections from localhost.
+  unprivilegedHostNetwork: false
```
kang-makes, line 127: Why call it `unprivilegedHostNetwork` and not `hostNetwork` directly?

charts/newrelic-infrastructure-v3/templates/control-plane/_helpers.tpl
```diff
@@ -0,0 +1,8 @@
+{{/* Returns whether the controlPlane scraper should run with hostNetwork: true based on the user configuration. */}}
+{{- define "newrelic.controlPlane.hostNetwork" -}}
+{{- if .Values.privileged -}}
+true
+{{- else if .Values.controlPlane.unprivilegedHostNetwork -}}
+true
```
kang-makes, line 6: I would change the order, we what to honor `hostNetwork` value before the `privileged`.

Could this be another issue like the `lowDataMode` where we want to differentiate between `false` and "falsiness"?

kang-makes (commented):
I think we have another "discussion"... Again...

PR #675, opened 2022-01-25
[newrelic-infrastructure-v3] Adding needed variable for autodiscovery

paologallinaharbur:
#### Is this a new chart
no
Since hostnetwork is false now NRIA_HOST is needed. (read by the the discovery binary)
We removed the Cluster_name env var and the passthrough to the integration but it was still needed  

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure-v3/templates/kubelet/daemonset.yaml
+++ b/charts/newrelic-infrastructure-v3/templates/kubelet/daemonset.yaml
@@ -161,6 +161,17 @@ spec:
             {{- with .Values.kubelet.extraEnv }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
+            {{- /* Needed to populate clustername in integration metrics */}}
+            - name: "CLUSTER_NAME"
+              value: {{ include "newrelic.cluster" . | required "cluster name must be set" }}
+            - name: "NRIA_PASSTHROUGH_ENVIRONMENT"
+              value: "CLUSTER_NAME"
+            {{- /* Needed for autodiscovery since hostNetwork=false */}}
+            - name: "NRIA_HOST"
+              valueFrom:
+                fieldRef:
+                  apiVersion: "v1"
+                  fieldPath: "status.hostIP"
           {{- with .Values.kubelet.extraEnvFrom }}
           envFrom: {{ toYaml . | nindent 12 }}
           {{- end }}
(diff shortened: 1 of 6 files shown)


paologallinaharbur:
Tests are broken for the `upgrade` since the previous chart introduced the timeout breaking change and it is not capable of installing the old chart

charts/newrelic-infrastructure-v3/templates/kubelet/daemonset.yaml
```diff
@@ -161,6 +161,17 @@ spec:
             {{- with .Values.kubelet.extraEnv }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
+            {{- /* Needed to populate clustername in integration metrics */}}
+            - name: "CLUSTER_NAME"
+              value: {{ include "newrelic.cluster" . | required "cluster name must be set" }}
+            - name: "NRIA_PASSTHROUGH_ENVIRONMENT"
+              value: "CLUSTER_NAME"
```
roobre, line 168: This could be dangerous as some integrations might have a `CLUSTER_NAME` argument that is not the K8s cluster name, but the integration cluster name: E.g. the Kafka cluster name, haproxy cluster name, etc.

charts/newrelic-infrastructure-v3/templates/kubelet/daemonset.yaml
```diff
@@ -161,6 +161,17 @@ spec:
...
+            {{- /* Needed for autodiscovery since hostNetwork=false */}}
+            - name: "NRIA_HOST"
+              valueFrom:
+                fieldRef:
+                  apiVersion: "v1"
+                  fieldPath: "status.hostIP"
```
roobre, line 174: Do we have control on the name for this env var? If so, i think we should use the same prefix we have in other places, like `NRI_KUBERNETES_NODE_IP`.

paologallinaharbur: Not really and I dislike it as well. The prefix is set [here](https://github.com/newrelic/nri-discovery-kubernetes/blob/f0e731a9bdd657c050d21f33b9ce97309eddb57c/internal/config/config.go#L25).

Changing it is easy but could lead to a breaking change :( 

Moreover I would chose something NRI_DISCOVERY

roobre: The fact that it uses the same prefix as the agent worries me a bit... Are we sure the agent is not using the same variable for anything else?

charts/newrelic-infrastructure-v3/templates/kubelet/daemonset.yaml
roobre: Are we sure that is not the intended behavior, and what we are doing right now a bug?

Reminds me of this: https://github.com/newrelic/nri-haproxy/issues/28

roobre (approved):

PR #676, opened 2022-01-25
Modify k8s object names to be shorter

alvarocabanas:
- Modify k8s object names to be shorter
- control-plane now is controlplane
- the default name is RELEASENAME-nrk8s-object


charts/newrelic-infrastructure-v3/tests/podName_test.yaml
```diff
@@ -0,0 +1,46 @@
+suite: test object names
+templates:
+  - templates/controlplane/daemonset.yaml
```
paologallinaharbur, line 3: ❤️

roobre (commented):
Looking great, thanks a lot!
Seems like we need to bump the version, but other than that we should be rtm!

roobre (approved):
Looking rad, thanks!

PR #678, opened 2022-01-26
newrelic-infrastructure-v3: render integrations and agent configmaps unconditionally

roobre:

Previously, we completely skipped config maps and their respective mounts if the agent config or the integrations config was empty. This was done to reduce clutter and unnecessary objects in the clusters where the chart is deployed.

However, while working on the docs (https://github.com/newrelic/docs-website/pull/5742), I realized that for installations using the generated manifests (which we support) we point users to modify or add items to these configmaps. If the configmaps do not exist, it would be hard for users  to create them and mount them in the correct places.

For this reason, this PR makes these manifests to render unconditionally, and adds some comments and examples on them.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure-v3/templates/kubelet/integrations-configmap.yaml
+++ b/charts/newrelic-infrastructure-v3/templates/kubelet/integrations-configmap.yaml
@@ -1,4 +1,3 @@
-{{ if or .Values.integrations (include "newrelic.compatibility.integrations" .)}}
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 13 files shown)


alvarocabanas (approved):
LGTM, it will be much clearer for the users having that example.

charts/newrelic-infrastructure-v3/templates/kubelet/integrations-configmap.yaml
```diff
@@ -7,11 +6,32 @@ metadata:
     {{- include "newrelic.labels" . | nindent 4 }}
   name: {{ template "newrelic.fullname" . }}-integrations-cfg
 data:
+  # This ConfigMap holds config files for integrations. They should have the following format:
+  #redis-config.yml: |
+  #  # Run auto discovery to find pods with label "app=redis"
+  #  discovery:
+  #    command:
+  #      # Run discovery for Kubernetes. Use the following optional arguments:
```
paologallinaharbur, line 14: why is this useful? Do we point here from docs?

paologallinaharbur: isn't better somewhere else? Templates with those info are weird

paologallinaharbur (commented):
tests still failing due to 
```
  - templates/kubelet/integrations-configmap.yaml
```

Moreover, I am not sure how users are going to see that comment you added, are you going to point it from docs?

Everything else looks good

charts/newrelic-infrastructure-v3/templates/kubelet/integrations-configmap.yaml
roobre: Yes, in de docs we (now) say:

> In order for the snippet above to be recognized by the integration, it must be placed inside the designated ConfigMap. If you are using our Kubernetes Integration v3, a ConfigMap should already be generated, with a name ending in `-integrations-cfg`. Locate the config map and add the modified snippet to it, so it ends up looking like this:

I thought that having a commented-out example in there would make extra clear to the user that's the ConfigMap they need to edit.

paologallinaharbur: mmm 👍  I understand we do not follow normal flow since we point users not using helm to check on helm charts template.

I am fine with it

paologallinaharbur (approved):

PR #680, opened 2022-01-26
newrelic-infrastructure-v3: restrict interval to [10, 45]s

roobre:

Causes the chart to fail to render if `common.config.interval` is not in the `[10, 45]s` range.

This check can be overridden by setting the undocumented and unsupported flag `forceUnsupportedInterval` to `true`.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure-v3/tests/interval_override_test.yaml
+++ b/charts/newrelic-infrastructure-v3/tests/interval_override_test.yaml
@@ -1,0 +1,18 @@
+suite: test interval
+templates:
+  - templates/controlplane/configmap.yaml
+  - templates/ksm/configmap.yaml
+  - templates/kubelet/configmap.yaml
+tests:
+  - it: Does not fail with override
+    set:
+      licenseKey: test
+      cluster: test
+      common:
+        config:
+          interval: 1s
+      forceUnsupportedInterval: true
+    asserts:
+      - matchRegex:
+          path: data.nri-kubernetes\.yml
+          pattern: 'interval: 1s'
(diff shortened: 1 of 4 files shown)


paologallinaharbur (approved):
LGTM

charts/newrelic-infrastructure-v3/tests/interval_test.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - failedTemplate:
+          errorMessage: "raw: Intervals smaller than 10s are not supported"
+  - it: Does not fail with override
+    set:
+      licenseKey: test
+      cluster: test
+      common:
+        config:
+          interval: 1s
+      forceUnsupportedInterval: true
```
paologallinaharbur, line 32: just a doupt, is it asserting/doing anything? this test?

paologallinaharbur: maybe it is better just to check the interval was correctly set

roobre: Right now it's checking that the chart rendering does not fail, but that's a good point: I can check for the interval as well!

charts/newrelic-infrastructure-v3/tests/interval_override_test.yaml
```diff
@@ -0,0 +1,18 @@
...
+      licenseKey: test
+      cluster: test
+      common:
+        config:
+          interval: 1s
+      forceUnsupportedInterval: true
+    asserts:
+      - matchRegex:
+          path: data.nri-kubernetes\.yml
+          pattern: 'interval: 1s'
```
paologallinaharbur, line 18: nice! better that adding the whole config!

paologallinaharbur (approved):
thanks for addressing the corner case!

paologallinaharbur:
you have two PR pointing to the same version

PR #689, opened 2022-01-29
[synthetics-minion] add priorityClassName

keegoid-nr:
Add `priorityClassName` for those who want to assign a higher `priorityclass` to the minion pod.

[x] bumped chart version
[x] added variable to readme
[x] set pr name to appropriate chart


charts/synthetics-minion/templates/statefulset.yaml
```diff
@@ -74,6 +74,14 @@ spec:
             claimName: {{ .Values.persistence.claimName | quote }}
       {{ end }}
 
+      {{- if $.Values.priorityClassName }}
```
marcusperezNR, line 77: Seeing duplicates here, can we fix this?

marcusperezNR (changes requested):
Just one fix, and then its GTG

marcusperezNR (approved):
LGTM

PR #697, opened 2022-02-07
newrelic-infrastructure-v3: default to recreate strategy for controlPlane Deployment

roobre:
#### Is this a new chart

No

Changes the default `strategy` for the control plane component to `type: Recreate` strategy when `kind` is `Deployment`. This is done to prevent new versions of the deployment from getting stuck in `Pending` state, as the previous version is occupying the same host port in use.

This strategy can be overridden from `values.yaml`.

Opting out of `hostNetwork` when privileged is set to true is not possible at the moment, and is out of the scope of this PR.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure-v3/templates/controlplane/daemonset.yaml
+++ b/charts/newrelic-infrastructure-v3/templates/controlplane/daemonset.yaml
@@ -11,9 +11,13 @@ metadata:
   annotations: {{ . | nindent 4 }}
   {{- end }}
 spec:
-  {{- if eq .Values.controlPlane.kind "DaemonSet"}}
+  {{- if eq .Values.controlPlane.kind "DaemonSet" }}
   {{- with .Values.updateStrategy }}
   updateStrategy: {{ toYaml . | nindent 4 }}
+  {{- end }}
+  {{- else if eq .Values.controlPlane.kind "Deployment" }}
+  {{- with .Values.controlPlane.strategy }}
+  strategy: {{ toYaml . | nindent 4 }}
   {{- end }}
   {{- end }}
   selector:
--- a/charts/newrelic-infrastructure-v3/templates/controlplane/daemonset.yaml
+++ b/charts/newrelic-infrastructure-v3/templates/controlplane/daemonset.yaml
@@ -16,7 +16,7 @@ spec:
   updateStrategy: {{ toYaml . | nindent 4 }}
   {{- end }}
   {{- else if eq .Values.controlPlane.kind "Deployment" }}
-  {{- with .Values.controlPlane.strategy }}
+  {{- with .Values.strategy }}
   strategy: {{ toYaml . | nindent 4 }}
   {{- end }}
   {{- end }}
--- a/charts/newrelic-infrastructure-v3/tests/controlplane_strategy_test.yaml
+++ b/charts/newrelic-infrastructure-v3/tests/controlplane_strategy_test.yaml
@@ -1,0 +1,88 @@
+suite: Control plane strategy
+templates:
+  - templates/controlplane/daemonset.yaml
+  - templates/controlplane/configmap.yaml
+  - templates/secret.yaml
+  - templates/agent-configmap.yaml
+  - templates/_helpers.tpl
+  - templates/_helpers_compatibility.tpl
+tests:
+- it: DaemonSet defaults to global updateStrategy
+  set:
+    licenseKey: test
+    cluster: test
+  asserts:
+    - equal:
+        path: spec.updateStrategy
+        value:
+          type: RollingUpdate
+          rollingUpdate:
+            maxUnavailable: 1
+      template: templates/controlplane/daemonset.yaml
+- it: DaemonSet updateStrategy can be overridden
+  set:
+    licenseKey: test
+    cluster: test
+    controlPlane:
+      kind: Deployment
+  asserts:
+    - equal:
+        path: spec.strategy
+        value:
+          type: Recreate
+      template: templates/controlplane/daemonset.yaml
+    - isNull:
+        path: spec.updateStrategy
+      template: templates/controlplane/daemonset.yaml
+- it: Deployment defaults to recreate
+  set:
+    licenseKey: test
+    cluster: test
+    controlPlane:
+      kind: Deployment
+  asserts:
+    - equal:
+        path: spec.strategy
+        value:
+          type: Recreate
+      template: templates/controlplane/daemonset.yaml
+- it: Deployment strategy can be overridden
+  set:
+    licenseKey: test
+    cluster: test
+    controlPlane:
+      kind: Deployment
+      strategy:
+        type: Foobar
+  asserts:
+    - equal:
+        path: spec.strategy
+        value:
+          type: Foobar
+      template: templates/controlplane/daemonset.yaml
+    - isNull:
+        path: spec.updateStrategy
+      template: templates/controlplane/daemonset.yaml
+- it: Deployment strategy is not rendered on DaemonSet
+  set:
+    licenseKey: test
+    cluster: test
+    controlPlane:
+      strategy:
+        type: Foobar
+  asserts:
+    - isNull:
+        path: spec.strategy
+      template: templates/controlplane/daemonset.yaml
+- it: DaemonSet strategy is not rendered on Deployment
+  set:
+    licenseKey: test
+    cluster: test
+    controlPlane:
+      kind: Deployment
+      updateStrategy:
+        type: Foobar
+  asserts:
+    - isNull:
+        path: spec.updateStrategy
+      template: templates/controlplane/daemonset.yaml
--- a/charts/newrelic-infrastructure-v3/tests/controlplane_strategy_test.yaml
+++ b/charts/newrelic-infrastructure-v3/tests/controlplane_strategy_test.yaml
@@ -50,10 +50,10 @@ tests:
   set:
     licenseKey: test
     cluster: test
+    strategy:
+      type: Foobar
     controlPlane:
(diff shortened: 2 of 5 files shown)


charts/newrelic-infrastructure-v3/tests/controlplane_strategy_test.yaml
```diff
@@ -0,0 +1,90 @@
...
+      cluster: test
+    asserts:
+      - equal:
+          path: spec.updateStrategy
+          value:
+            type: RollingUpdate
+            rollingUpdate:
+              maxUnavailable: 1
+        template: templates/controlplane/daemonset.yaml
+  - it: updateStrategy can be overridden
```
kang-makes, line 22:
```suggestion
  - it: updateStrategy can be overridden in DaemonSet kind
```

kang-makes (commented):
I don't know if I am comfortable with this approach.

Let me think this afternoon and I'll do another PR.

kang-makes:
#698 is my approach. Let's discuss and I take more time to write the tests for it.

roobre:
@kang-makes Simplified the approach a bit in https://github.com/newrelic/helm-charts/pull/697/commits/8229892d576f08c184bcf31318a65ea7b73bdc72

kang-makes (commented):
Let's start a discussion about if we have to change the `updateStragety` or the logic of the `hostNetwork` helper...

charts/newrelic-infrastructure-v3/templates/controlplane/daemonset.yaml
```diff
@@ -11,9 +11,15 @@ metadata:
...
+  {{- if eq .Values.controlPlane.kind "DaemonSet" }}
   {{- with .Values.updateStrategy }}
-  updateStrategy: {{ toYaml . | nindent 4 }}
+  updateStrategy:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- else if eq .Values.controlPlane.kind "Deployment" }}
+  {{- with .Values.controlPlane.strategy }}
+  strategy:
+    {{- toYaml . | nindent 4 }}
```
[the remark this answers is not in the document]
roobre: I don't think there's a problem with the `hostNetwork` helper. It does what it needs to do: Checking if the user _allows_ (not sets) running with `hostNetwork` or not.

I think the issue here is that `hostNetwork` is just not compatible with deploying a Deployment (if you open a port as we do), so my approach is to change the strategy and _not_ hostNetwork. This way the user has full control over both the strategy and hostNetwork and can tweak them however they please, while keeping a default that works.

Why should we actively prevent the user from running `hostNetwork`? It might not be a common use case, but my motto here is to avoid as much as possible for the user to need to fork the chart, open issues, or the worst of all, deploy manifests modified by hand.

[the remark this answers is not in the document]
kang-makes: So what you want to do is the same as in `lowDataMode`: https://github.com/newrelic/helm-charts/pull/632/files#diff-ad90df5b5bd9a1c31f58e9ee28b3ad8ac425e6d720533d960f1495120e98cdb8R122-R139

Delete `unprivilegedHostNetwork` and `hostNetwork` from the chart's `values.yaml` (comment them) and change the default value depending on the `Kind` of the control plane. If it is `DaemonSet`, `hostNetwork` is `true`, else is `false`, unless the user have put a value in its `values.yaml` that overrides any default.

Could this be a better approach? Would you like that I do that logic in my PR for comparing it?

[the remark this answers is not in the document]
roobre: > Delete unprivilegedHostNetwork and hostNetwork from the chart's values.yaml (comment them) and change the default value depending on the Kind of the control plane.

On the contrary, I suggest to keep all of them and let users combine them however they want.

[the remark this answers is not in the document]
kang-makes: Delete as in "comment them", not as in "delete its functionality".

`lowDataMode` has it commented to make it overridable: https://github.com/newrelic/helm-charts/pull/632/files#diff-b5aab5e71b9779d8667e1a61c2f7ac342c2833e4c74ccafebdc77b8d43792be1R26

paologallinaharbur:
what I do not particularly like is that we take a "root" value if DaemonSet and a "controlPlane" one if Deployment.

Would make sense to have a root level strategy used as well for KSM? Or we want to avoid that since if multiple replicas are no really supported

roobre:
Remade the PR with a way simpler approach. Opting out of `hostNetwork` when `privileged` is set to true is not possible at the moment, and is out of the scope of this PR.

charts/newrelic-infrastructure-v3/templates/controlplane/daemonset.yaml
```diff
@@ -11,10 +11,14 @@ metadata:
...
 spec:
-  {{- if eq .Values.controlPlane.kind "DaemonSet"}}
+  {{- if eq .Values.controlPlane.kind "DaemonSet" }}
   {{- with .Values.updateStrategy }}
   updateStrategy: {{ toYaml . | nindent 4 }}
   {{- end }}
+  {{- else if eq .Values.controlPlane.kind "Deployment" }}
+  {{- with .Values.controlPlane.strategy }}
+  strategy: {{ toYaml . | nindent 4 }}
+  {{- end }}
```
kang-makes, line 21: I didn't see that coming...

charts/newrelic-infrastructure-v3/values.yaml
```diff
@@ -143,6 +143,10 @@ controlPlane:
   # -- Run Control Plane scraper with `hostNetwork` even if `privileged` is set to false.
   # `hostNetwork` is required for most control plane configurations, as they only accept connections from localhost.
   unprivilegedHostNetwork: false
+  # -- Deployment strategy when `kind` is `DaemonSet`.
+  # @default -- `type: Recreate`
+  strategy:
+    type: Recreate
```
kang-makes, line 149: I would delete the `strategy` and `updateStrategy` and leave them empty to the cluster to choose those values.

I could leave the `strategy` as `Recreate` is not the default strategy type...

kang-makes (commented):
Much simpler. I like it but I still don't like that a thing that needs not `hostNetwork` has it by default.

Also the `Affinity` issue where it will go to look for nonexistent "control plane" nodes.

charts/newrelic-infrastructure-v3/values.yaml
roobre: As we have discussed, defaulting to anything other than `Recreate` will result in a broken experience if `hostNetwork` is in use.

charts/newrelic-infrastructure-v3/templates/controlplane/daemonset.yaml
roobre: https://www.youtube.com/watch?v=Ec4YbVP9R-A&t=46s

roobre:
> I still don't like that a thing that needs not `hostNetwork` has it by default.

There are use cases where you want `hostNetwork` even for `Deployment`s, like using complex CNIs, service meshes, or funny network rules defining baremetal. A good example of this is https://github.com/newrelic/helm-charts/pull/709.

But perhaps more importantly, I deeply dislike changing defaults. If I change the `kind` I'd expect only the `kind` to change.

> Also the Affinity issue where it will go to look for nonexistent "control plane" nodes.

Once again, I think changing defaults behind the curtains is evil.

charts/newrelic-infrastructure-v3/templates/controlplane/daemonset.yaml
kang-makes: Neither this.

paologallinaharbur: at some point we are going to have:
```
charts/newrelic-infrastructure-v3/templates/controlplane/daemonset.yaml
charts/newrelic-infrastructure-v3/templates/controlplane/deployment.yaml
```

charts/newrelic-infrastructure-v3/templates/controlplane/daemonset.yaml
```diff
@@ -11,10 +11,14 @@ metadata:
   annotations: {{ . | nindent 4 }}
   {{- end }}
 spec:
-  {{- if eq .Values.controlPlane.kind "DaemonSet"}}
+  {{- if eq .Values.controlPlane.kind "DaemonSet" }}
   {{- with .Values.updateStrategy }}
   updateStrategy: {{ toYaml . | nindent 4 }}
   {{- end }}
+  {{- else if eq .Values.controlPlane.kind "Deployment" }}
+  {{- with .Values.controlPlane.strategy }}
```
paologallinaharbur, line 19: I still think that if we have ".Values.updateStrategy" we should have ".Values.strategy"

roobre: Good point, I can change that :)

paologallinaharbur:
I would try to merge this or the other PR (or at least decide the approach).

We are moving repo and we risk to forget about this

kang-makes:
I am not forgetting this because I will have to adapt `common.hostNetwork` to `nri-kubernetes` but we will raise this discussion again. If we cannot have a solution async then we will meet sync someday.

gsanchezgavier (approved):
I like this approach a bit more, it has less nested logic and It solves the issue of ports

paologallinaharbur:
Closing this, we have moved the chart to newrelic/nri-kubernetes

PR #700, opened 2022-02-08
[newrelic-logging] fix empty messages with CRI

### LINKED ISSUE #699: [newrelic-logging] Empty log messages with lowDataMode and containerd
**Bug description**

When using New Relic logging in lowDataMode and containerd (`criEnabled: true`) the message field is unpopulated due to fluentbit not preserving the 'message' field.

v3n:
When running containerd, the log message was being dropped due 'message'
not being in the record_modifier allowlist. The cri parser now
unmarshals the field as 'log' instead of 'message'. 'log' is the
preferred field name as the Kubernetes filter checks that key for JSON
when Merge_Log is set to true.

#### Is this a new chart

No.

Fixes lowDataMode when using `criEnabled: true`. See #699 

Fixes #699 

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/newrelic-logging/values.yaml
+++ b/charts/newrelic-logging/values.yaml
@@ -125,7 +125,7 @@ fluentBit:
       [PARSER]
           Name cri
           Format regex
-          Regex ^(?<time>[^ ]+) (?<stream>stdout|stderr) (?<logtag>[^ ]*) (?<message>.*)$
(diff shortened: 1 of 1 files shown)


jsubirat (changes requested):
Hi @v3n, thank you for your submission. Thanks for making this point, indeed, being able to use `Merge_Log` with the `cri` parser makes total sense. Could you please increase the Chart version [here](https://github.com/newrelic/helm-charts/blob/master/charts/newrelic-logging/Chart.yaml#L4), please? With this, we should be ready to merge. Thanks!

PR #707, opened 2022-02-10
[newrelic-infrastructure] Adding nri-flex config for Pixie cloud connector health and status checks

bpschmitt:
#### Is this a new chart

No

Adds the `flex/nri-pixie-check.yaml` flex config to the `newrelic-infra-integrations-cfg` configMap.  This provides users the ability to query the status of Pixie health check endpoints using `pixieHealthCheck` custom events.

N/A

#### Checklist
- [X] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure/flex/nri-pixie-check.yaml
+++ b/charts/newrelic-infrastructure/flex/nri-pixie-check.yaml
@@ -1,0 +1,22 @@
+---
+discovery:
+  command:
+    exec: /var/db/newrelic-infra/nri-discovery-kubernetes
+    match:
+      label.name: vizier-cloud-connector
+integrations:
+  - name: nri-flex
+    interval: 60s
+    config:
+      name: pixie-health-check
+      apis:
+        - event_type: pixieHealth
+          commands:
+            - run: curl --insecure -s https://${discovery.ip}:50800/healthz | xargs | awk '{print "cloud_connector_health:"$1}'
+              split_by: ":"
+          merge: pixieHealthCheck
+        - event_type: pixieStatus
+          commands:
+            - run: curl --insecure -s https://${discovery.ip}:50800/statusz | awk '{if($1 == ""){ print "cloud_connector_status:OK" } else { print "cloud_connector_status:"$1 }}'
+              split_by: ":"
+          merge: pixieHealthCheck
--- a/charts/newrelic-infrastructure/flex/nri-pixie-check.yaml
+++ b/charts/newrelic-infrastructure/flex/nri-pixie-check.yaml
@@ -1,9 +1,10 @@
 ---
 # This Flex config performs periodic checks of the Pixie
-# healthz and statusz endpoints.  A status for each endpoint
-# is sent to New Relic in a pixieHealthCheck event.
+# /healthz and /statusz endpoints exposed by the Pixie Cloud Connector.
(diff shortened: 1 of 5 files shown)


charts/newrelic-infrastructure/flex/nri-pixie-check.yaml
```diff
@@ -0,0 +1,27 @@
...
+      name: pixie-health-check
+      apis:
+        - event_type: pixieHealth
+          commands:
+            - run: curl --insecure -s https://${discovery.ip}:50800/healthz | xargs | awk '{print "cloud_connector_health:"$1}'
+              split_by: ":"
+          merge: pixieHealthCheck
+        - event_type: pixieStatus
+          commands:
+            - run: curl --insecure -s https://${discovery.ip}:50800/statusz | awk '{if($1 == ""){ print "cloud_connector_status:OK" } else { print "cloud_connector_status:"$1 }}'
```
roobre, line 25: I remember some weirdness could happen with unscaped `$` symbols (like in `$1` here) in pixie/agent configs, but unfortunately not the details of it. Feel free to ignore if you have tested this to be working.

charts/newrelic-infrastructure/templates/configmap.yaml
```diff
@@ -18,6 +17,8 @@ metadata:
   labels: {{ include "newrelic.labels" . | indent 4 }}
   name: {{ template "newrelic.fullname" . }}-integrations-cfg
 data:
+{{ (.Files.Glob "flex/**.yaml").AsConfig | indent 2 }} # include Flex configs
```
roobre, line 20: I think this is a clever approach, but I'm not 100% sure we need a generic way of including files if we don't foresee needing more, especially since users will not be able to add files to this folder.

roobre (commented):
Many thanks for your PR @bpschmitt!

Left some thoughts inline, mostly minor. On more general lines, I think we should add a toggle so customers can disable this if they are not comfortable with it. This toggle can be opt-out.

A rough idea of how would that look could be this, which would be an snippet of `values.yml`

```yaml
# -- Control reporting very lightweight metrics about the health of New Relic components.
# @default -- Enabled
selfMonitoring:
  pixie:
    enabled: true
```

We could then include the pixie flex config if this is set to true. I don't think there's a need to guard the whole config map with complex `if or`s, rendering an empty one should be fine.

charts/newrelic-infrastructure/flex/nri-pixie-check.yaml
```diff
@@ -0,0 +1,27 @@
...
+              split_by: ":"
+          merge: pixieHealthCheck
```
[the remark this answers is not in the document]
bpschmitt: Fixed

charts/newrelic-infrastructure/templates/configmap.yaml
bpschmitt: @roobre I made this more specific [here](https://github.com/newrelic/helm-charts/pull/707/files#diff-a6b2b5ef195e7a5534a265575a2bfdc196c15e03b15f4d27af7ffb1e76bbf413R20-R23).

roobre (approved):
Looking alright from my side @bpschmitt, thanks for looking into this!

You might need to rebase `master` and re-bump the version, but everything else looks nice to me.

PR #711, opened 2022-02-15
newrelic-infrastructure-v3: rename `nrk8s-kubelet` DS to `nrk8s-node`

roobre:
#### Is this a new chart

No

Changes the name of the `nrk8s-kubelet` DaemonSet (and its pods) to `nrk8s-node`.

While the most important task of the DaemonSet is pulling metrics from the Kubelet, it can also pull more detailed node metrics, as well as run OHIs. For this reason, the `-node` suffix might be more accurate.

Moreover, the `-kubelet` suffix could be slightly confusing for users that are not extremely familiar with the Kubernetes components, while the word `node` is more universally understandable.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure-v3/templates/kubelet/daemonset.yaml
+++ b/charts/newrelic-infrastructure-v3/templates/kubelet/daemonset.yaml
@@ -224,7 +224,7 @@ spec:
           emptyDir: {}
         - name: nri-kubernetes-config
           configMap:
-            name: {{ template "newrelic.fullname" . }}-node
+            name: {{ template "newrelic.fullname" . }}-kubelet
             items:
               - key: nri-kubernetes.yml
                 path: nri-kubernetes.yml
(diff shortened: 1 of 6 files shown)


paologallinaharbur:
I agree that the name improves, but I would try to be "uniform" and therefore having that also in the values.yaml.
If we have "node" instead of kubelet in pod names, I would expect to have "node" as well in values. 
Would the change be too big ?

charts/newrelic-infrastructure-v3/templates/kubelet/daemonset.yaml
```diff
@@ -224,7 +224,7 @@ spec:
           emptyDir: {}
         - name: nri-kubernetes-config
           configMap:
-            name: {{ template "newrelic.fullname" . }}-kubelet
```
paologallinaharbur, line 227: still old name is needed here as well

paologallinaharbur (approved):
🚀

paologallinaharbur:
Stale, closing for now

PR #712, opened 2022-02-16
[newrelic-infrastructure-v3] improved error messages

paologallinaharbur:
#### Is this a new chart
no
 - `--reuse-values` is not allowed when migrating from V2 to V3. It is specified into the docs, however it was not super clear at update time.

 - Moreover I refactor a bit the errors adding a link to docs:

```
Error: UPGRADE FAILED: execution error at (nri-bundle/charts/newrelic-infrastructure/templates/NOTES.txt:133:4):

######
The chart cannot be rendered since the values listed below are not supported. Please replace those with the new ones compatible with newrelic-infrastructure V3.

Keep in mind that the flag "--reuse-values" is not supported when migrating from V2 to V3.
Further information can be found in the official docs https://docs.newrelic.com/docs/kubernetes-pixie/kubernetes-integration/get-started/changes-since-v3#migration-guide"
######


You have specified the legacy 'tolerations' option in your values, which is not fully compatible with the v3 version.
This version deploys three different components and therefore you'll need to specify tolerations for each of them.
Please use
 - ksm.tolerations,
 - controlPlane.tolerations,
 - kubelet.tolerations.

------

nri-kubernetes v3 does not support deploying into windows Nodes.
Please use the latest 2.x version of the chart.

------
```

 - enableWindows  compatibility rule was too strict

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
+++ b/charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
@@ -85,35 +85,52 @@ Returns legacy integrations_config configmap data
 {{- end -}}
 
 {{- define "newrelic.compatibility.message.logFile" -}}
-The `logFile` option is no longer supported and has been replaced by common.agentConfig.log_file.
+The 'logFile' option is no longer supported and has been replaced by:
+ - common.agentConfig.log_file.
+
+------
 {{- end -}}
 
 {{- define "newrelic.compatibility.message.resources" -}}
-You have specified the legacy `resources` option in your values, which is not fully compatible with the v3 version.
+You have specified the legacy 'resources' option in your values, which is not fully compatible with the v3 version.
 This version deploys three different components and therefore you'll need to specify resources for each of them.
-Please use `ksm.resources`, `controlPlane.resources` and `kubelet.resources`.
+Please use
+ - ksm.resources,
+ - controlPlane.resources,
+ - kubelet.resources.
+
+------
 {{- end -}}
 
 {{- define "newrelic.compatibility.message.tolerations" -}}
-You have specified the legacy `tolerations` option in your values, which is not fully compatible with the v3 version.
+You have specified the legacy 'tolerations' option in your values, which is not fully compatible with the v3 version.
 This version deploys three different components and therefore you'll need to specify tolerations for each of them.
-Please use `ksm.tolerations`, `controlPlane.tolerations` and `kubelet.tolerations`.
+Please use
+ - ksm.tolerations,
+ - controlPlane.tolerations,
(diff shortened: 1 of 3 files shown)


charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
```diff
@@ -86,29 +86,45 @@ Returns legacy integrations_config configmap data
 
 {{- define "newrelic.compatibility.message.logFile" -}}
 The `logFile` option is no longer supported and has been replaced by common.agentConfig.log_file.
+
```
gsanchezgavier, line 89: perhaps list this one too , just to have the same format as others

gsanchezgavier (approved):
LGTM just added a small nit

paologallinaharbur:
Fixing the PR, I forgot we cannot specify any message in the "errors" variable

charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
```diff
@@ -86,29 +86,45 @@ Returns legacy integrations_config configmap data
...
 {{- end -}}
 
 {{- define "newrelic.compatibility.message.resources" -}}
 You have specified the legacy `resources` option in your values, which is not fully compatible with the v3 version.
 This version deploys three different components and therefore you'll need to specify resources for each of them.
-Please use `ksm.resources`, `controlPlane.resources` and `kubelet.resources`.
+Please use
+ - `ksm.resources`,
+ - `controlPlane.resources`
+ - `kubelet.resources`.
```
kang-makes, line 99: These are commands or subshells. Terminals do not render Markdown... YET.

Weird things could happen in case somebody pipe/script this and do thing as "saving the output of `helm upgrade --yadayada` as an environment variable to print it or something like that.

A inexperienced user could see it also as a copy and paste command like the ones on `info command`. Here you have an example with `info diff`:
Screenshot 2022-02-16 at 10 27 08

They put a ` with a ' to avoid copy and pasting or escaping problems.

kang-makes (commented):
LGTM. Only left a nit.

charts/newrelic-infrastructure-v3/templates/_helpers_compatibility.tpl
paologallinaharbur: Thanks! I'll replace those

roobre:
Why have the ``` quotes been replaced with `'`?

paologallinaharbur:
Following @kang-makes andive just above!

kang-makes (approved):
I don't know why I didn't approve it the first time.

PR #714, opened 2022-02-16
[newrelic-infrastructure-v3] init container

paologallinaharbur:
#### Is this a new chart
no
adding init containers needed in some scenarios

related to https://github.com/newrelic/newrelic-coreint/issues/291

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure-v3/templates/_helpers.tpl
+++ b/charts/newrelic-infrastructure-v3/templates/_helpers.tpl
@@ -240,4 +240,3 @@ interval: 30s
 interval: 15s
 {{- end -}}
 {{- end -}}
-
--- a/charts/newrelic-infrastructure-v3/Chart.yaml
+++ b/charts/newrelic-infrastructure-v3/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: newrelic-infrastructure-v3
 description: A Helm chart to deploy the New Relic Kubernetes monitoring solution
-version: 3.0.18
+version: 3.0.21
 appVersion: 3.0.0
 kubeVersion: ">=1.16.0-0"
 home: https://docs.newrelic.com/docs/kubernetes-pixie/kubernetes-integration/get-started/introduction-kubernetes-integration/
--- a/charts/newrelic-infrastructure-v3/README.md
+++ b/charts/newrelic-infrastructure-v3/README.md
@@ -2,7 +2,7 @@
 
 # newrelic-infrastructure-v3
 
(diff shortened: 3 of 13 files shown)


charts/newrelic-infrastructure-v3/templates/_helpers.tpl
```diff
@@ -241,3 +241,13 @@ interval: 15s
 {{- end -}}
 {{- end -}}
 
+{{/*
+Renders a value that contains template.
+*/}}
+{{- define "newrelic.tplvalues.render" -}}
```
paologallinaharbur, line 247: still deciding about this

kang-makes: Use the helpers in the init containers to have access to options like license, staging and such.

I think it could be useful.

kang-makes: I don't see the point of limiting the initContainer `tpl` to `.context`.

Why wouldn't I want to have access to all the `.Values` and templates?

roobre: Everything is passed, see here: https://github.com/newrelic/helm-charts/pull/714/files#diff-b2298e732a892ff75958aab46b0835f139eb0ed344eede9b00569040056a4d18R47

paologallinaharbur: I took it from here https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_tplvalues.tpl while looking around possible ways to do so.

I am not sure it worth the additional complexity. Should we just have `{{- tpl (.value | toYaml) .context }}`. I am not sure we have any use-case for witch we need to support strings (apart from custom attributes)

kang-makes: If you are giving the ability to render things this makes things difficult. For example, if you want to `template "newrelic.fullname" .` you have to `template "newrelic.fullname" .context` or `template "newrelic.fullname" $` to properly render.

That is the thing that worries me the most: Having a `value` and `context` instead of having directly `.Values` and that's it.

I would not have in mind that the initContainers could be strings. That is a really rare use case that I would not like to support. On the other hand, I could find edge cases where I would want to have access to the values and the templating in the init containers so templating the YAML in `.Values.initContainers` sounds cool.

roobre (approved):
Looking great, thanks for taking care of this!

PR #724, opened 2022-03-03
[common-library] added globals for the image values

paologallinaharbur:
#### Is this a new chart
no

It includes in the common library two globals to deal with image.registry and image.pullSecrets

# Why the signature for the tpl is that complex?
The main reason is that image structure is not the very same across charts (sometimes is "image", sometimes "images")

Also pull secrets are at different levels, therefore each chart might need call the function pointing to a different value structure:
 - images.forwarder.registry (newrelic-infrastructure-v3)
vs
 - image.registry (in newrelic-k8s-metrics-adapter)
 

#### Checklist
- [x] Chart Version bumped


library/common-library/templates/_images.tpl
```diff
@@ -0,0 +1,53 @@
...
+
+{{- if .context.Values.global }}
+    {{- if .context.Values.global.image.registry }}
+     {{- $registryName = .context.Values.global.image.registry  -}}
+    {{- end -}}
+{{- end -}}
+
+{{- with .imageRoot.registry }}
+    {{- $registryName = . -}}
+{{ end }}
```
kang-makes, line 18: This logic that allows superseding the global value with a local one is already done in the `lowDataMode` helper: https://github.com/newrelic/helm-charts/blob/feat/imageGlobals/library/common-library/templates/_low-data-mode.tpl

Using that logic allows us in the future to simply put `{{ include "common.images.image" . }}` instead of having to create a `dict` for each helper.

It could sound that this is not a big deal now because we have 3 or 4 helpers but when we have 20 we are going to start finding issues like which dictionary does this helper needs to render correctly.

Is a pattern that was useful in the compatibility layer but I do not see that is going to work here.

kang-makes, line 18: Create more than one helper. For registry, image and tag:
 - The registry needs a global but image and tag don't.
 - Assume that the image is going to be in `.Values.image.repository`. We know that we are going to have sidecars but in that case we can override the helper that returns the image and add other helper for the sidecar.
 - Do not use the same helper for all the `registry/image:tag` so it can over used/overriden in parts.
 - You can then have other helper that use the three written above to create the full image path so it is readable in the deployment.yaml

library/common-library/templates/_images.tpl
```diff
@@ -0,0 +1,53 @@
...
+    {{- if not (empty .) }}
+      {{- $ps = append $ps . -}}
+    {{- end -}}
+  {{- end -}}
+
+  {{- if gt (len $ps) 0 -}}
+    {{- range $ps }}
+{{ toYaml . }}
+    {{- end -}}
+  {{- end -}}
```
kang-makes, line 52: There is no issue in `mergeOverwrite` global and local `pullImagesecrets`. WDYT?

kang-makes (commented):
Almost there but I would split the image helper to share the function for the registry with the sidecars...

library/common-library/templates/_images.tpl
paologallinaharbur: I don't think we can assume that the image is going to be in `.Values.image.repository`. Most of charts uses more than one image, either for the integration, the webhook, the agent.

paologallinaharbur: The issue is that with the images helpers is that we cannot assume where fetching the image, therefore calling the function we need to pass the root of the image

kang-makes: I think this approach is overcomplicating because of the sidecars.

I would create a helper for the registry, for the repository and the tag and a fourth that concatenates them.

In the Helm charts where we use sidecars we could opt to not use the common function that concatenated them and use the `.Values.images.agent.repository` if we need to.

kang-makes: Then we should do common helpers for what is common: registry. The rest should be done in each chart.

library/common-library/templates/_images.tpl
kang-makes: It is also an array like
```yaml
- name: test
- name: test2
```
so we cannot `mergeOverride`.

Let me recheck it.

library/common-library/templates/_images.tpl
paologallinaharbur: We discussed about this PR offline, we agreed that:
 - yes  image and tag do not need global values
 - we cannot assume image is in the same root in different repos -> most of charts uses different images
 - we need to split the helpers
 - sadly the complex function signature is needed. If too complex we can decide to drop that "global"

kang-makes (approved):
It could look ugly that it is a limitation of Helm and we should cover charts with more than one image very often.

LGTM.

PR #725, opened 2022-03-03
[common-library] global for security context

paologallinaharbur:
#### Is this a new chart
no

Add to globals and library the security context

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/CHART-TEMPLATE/templates/deployment.yaml
+++ b/library/CHART-TEMPLATE/templates/deployment.yaml
@@ -24,8 +24,10 @@ spec:
         {{- . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "common.serviceAccount.name" . }}
+      {{- with include "common.securityContext.pod" . }}
       securityContext:
-        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+        {{- . | nindent 8 }}
+      {{- end }}
       {{- with include "common.priorityClassName" . }}
       priorityClassName: {{ . }}
       {{- end }}
@@ -35,8 +37,10 @@ spec:
       {{- end }}
       containers:
         - name: {{ .Chart.Name }}
+          {{- with include "common.securityContext.container" . }}
           securityContext:
-            {{- toYaml .Values.securityContext | nindent 12 }}
+            {{- . | nindent 12 }}
+          {{- end }}
           image: {{ include "common.images.image" ( dict "imageRoot" .Values.image "context" .) }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           ports:
--- a/library/common-library/templates/_security-context.tpl
+++ b/library/common-library/templates/_security-context.tpl
@@ -12,7 +12,7 @@
 {{- end -}}
 
 
-{{- /* Allows to change defaults for container security context either staic of dinamically */ -}}
+{{- /* Allows to change defaults for container security context either statically or dynamically */ -}}
 {{- define "common.securityContext.containerDefaults" -}}
 {{- end -}}
 
@@ -31,6 +31,6 @@
 {{- end -}}
 
 
-{{- /* Allows to change defaults for pod security context either staic of dinamically */ -}}
+{{- /* Allows to change defaults for pod security context either statically or dynamically */ -}}
 {{- define "common.securityContext.podDefaults" -}}
 {{- end -}}
--- a/library/common-library/templates/_security-context.tpl
+++ b/library/common-library/templates/_security-context.tpl
@@ -1,0 +1,22 @@
+{{- /* Defines the container securityContext context */ -}}
+{{- define "common.securityContext.container" -}}
+{{- $global := index .Values "global" | default dict -}}
+
+{{- if .Values.containerSecurityContext -}}
+    {{- toYaml .Values.containerSecurityContext -}}
+{{- else if $global.containerSecurityContext -}}
+    {{- toYaml $global.containerSecurityContext -}}
+{{- end -}}
+{{- end -}}
+
+
+{{- /* Defines the pod securityContext context */ -}}
+{{- define "common.securityContext.pod" -}}
+{{- $global := index .Values "global" | default dict -}}
+
+{{- if .Values.podSecurityContext -}}
+    {{- toYaml .Values.podSecurityContext -}}
+{{- else if $global.podSecurityContext -}}
+    {{- toYaml $global.podSecurityContext -}}
+{{- end -}}
+{{- end -}}
--- a/library/common-library/templates/_security-context.tpl
+++ b/library/common-library/templates/_security-context.tpl
@@ -6,7 +6,14 @@
     {{- toYaml .Values.containerSecurityContext -}}
 {{- else if $global.containerSecurityContext -}}
     {{- toYaml $global.containerSecurityContext -}}
+{{- else -}}
+    {{- include "common.securityContext.containerDefaults" . -}}
 {{- end -}}
+{{- end -}}
+
+
+{{- /* Allows to change defaults for container security context either staic of dinamically */ -}}
+{{- define "common.securityContext.containerDefaults" -}}
 {{- end -}}
 
 
@@ -18,5 +25,12 @@
     {{- toYaml .Values.podSecurityContext -}}
 {{- else if $global.podSecurityContext -}}
     {{- toYaml $global.podSecurityContext -}}
+{{- else -}}
+    {{- include "common.securityContext.podDefaults" . -}}
 {{- end -}}
+{{- end -}}
+
+
+{{- /* Allows to change defaults for pod security context either staic of dinamically */ -}}
+{{- define "common.securityContext.podDefaults" -}}
 {{- end -}}
--- a/library/CHART-TEMPLATE/Chart.lock
+++ b/library/CHART-TEMPLATE/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common-library
   repository: file://../common-library
-  version: 0.10.0
-digest: sha256:81087781c9d1d118efc31743f7d7b4cbeb790dbb1d0d3a2a8144ae0e427af4c8
-generated: "2022-03-17T16:32:44.730673+01:00"
+  version: 0.11.0
+digest: sha256:83edc5b9643b33c3420a2d6a19bea14b534d45d6ffb3284519a133904298f3a3
+generated: "2022-03-18T11:33:06.922088+01:00"
--- a/library/CHART-TEMPLATE/Chart.yaml
+++ b/library/CHART-TEMPLATE/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.10.0
+version: 0.11.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
@@ -25,7 +25,7 @@ appVersion: "1.16.0"
 
 dependencies:
   - name: common-library
-    version: 0.10.0
+    version: 0.11.0
     repository: file://../common-library
 
 keywords:
--- a/library/CHART-TEMPLATE/tests/common_images_security_context_container_test.yaml
+++ b/library/CHART-TEMPLATE/tests/common_images_security_context_container_test.yaml
@@ -1,0 +1,76 @@
+suite: test securityContainer at container level
+templates:
+  - templates/deployment.yaml
+release:
+  name: my-release
+  namespace: my-namespace
+set:
+tests:
+  - it: does not set securityContext if not specified
+    set:
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[0].securityContext
+          value: null
+
+  - it: does set securityContext if specified in values
+    set:
+      containerSecurityContext:
+        capabilities:
+          drop:
+          - ALL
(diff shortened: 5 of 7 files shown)


library/CHART-TEMPLATE/templates/deployment.yaml
```diff
@@ -23,12 +23,16 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "common.serviceAccount.name" . }}
+      {{- with include "common.securityContext.pod" . }}
       securityContext:
-        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+        {{- . | nindent 8 }}
```
gsanchezgavier, line 28: would make sense to be possible to add the `securityContext:` key inside the helper ? so in the actual template you just need to add the helper entry?

gsanchezgavier (approved):
LGTM

library/common-library/templates/_security-context.tpl
```diff
@@ -0,0 +1,25 @@
+{{- /* Defines the container securityContext context */ -}}
+{{- define "common.securityContext.container" -}}
+    {{- if .Values.global.securityContext -}}
```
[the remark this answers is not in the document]
gsanchezgavier: 💥

kang-makes, line 3: Here

library/common-library/templates/_security-context.tpl
```diff
@@ -0,0 +1,25 @@
...
+            {{- toYaml .Values.global.securityContext -}}
+        {{- end -}}
+    {{- else if .Values.securityContext -}}
+        {{- toYaml .Values.securityContext -}}
+    {{- end -}}
+{{- end -}}
+
+{{- /* Defines the pod securityContext context */ -}}
+{{- define "common.securityContext.pod" -}}
+    {{- if .Values.global.podSecurityContext -}}
```
kang-makes, line 16: And here.

kang-makes (commented):
Same as we talked. We cannot trust that `global` exists.

kang-makes:
I didn't refresh and I didn't see that you reviewed yourself...

I would like @roobre to review this because this is an issue that we talked about and I don't know if he is going to be happy about this.

I am owning this branch from now it you wish :)

library/common-library/templates/_security-context.tpl
```diff
@@ -0,0 +1,33 @@
+{{- /* Defines the container securityContext context */ -}}
+{{- define "common.securityContext.container" -}}
```
roobre, line 2: I like the idea of `.container` and `.pod` but I think the standard goes more towards the `podSecurityContext` and `containerSecurityContext` direction ([e.g.](https://github.com/kubernetes/ingress-nginx/blob/fb72fcd81772fb3eca923897aec2d92fa5bdff41/charts/ingress-nginx/values.yaml#L811)).

Do we have a strong reason for breaking the "standard"?

kang-makes: No. I'm on it.

library/common-library/templates/_security-context.tpl
```diff
@@ -0,0 +1,33 @@
...
+    {{- if .Values.global -}}
+        {{- if .Values.global.securityContext -}}
```
paologallinaharbur, line 4: logic is currently bit cumbersome due to the nested cases. I lend it of over to @kang-makes to be improved 😄

library/common-library/templates/_security-context.tpl
paologallinaharbur: the variable were not the same, it was just the function definition named like that to group them, but yeah better to stick to standards

library/CHART-TEMPLATE/templates/deployment.yaml
```diff
@@ -23,15 +23,19 @@ spec:
         {{- . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "common.serviceAccount.name" . }}
+      {{- with include "common.securityContext.pod" . }}
       securityContext:
```
roobre, line 27: Something thorny here is what will we do with the `privileged` flag, as it _needs_ to override some things in the security context (e.g. the `securityContext.privileged` flag). However I'm not sure if we want to discuss that in _this_ PR.

library/common-library/templates/_security-context.tpl
kang-makes: Solved in [#f27f4e9](https://github.com/newrelic/helm-charts/pull/725/commits/f27f4e99b33395141d632585fcd4c869f41c9ae1)

library/CHART-TEMPLATE/templates/deployment.yaml
kang-makes: We could talk about defaults changing if defaults "depending-on-things" in the `nri-kubernetes` chart.

kang-makes:
We [talked about replacing instead of merging](https://github.com/newrelic/helm-charts/pull/735#discussion_r821469445) so I changed how `securityContext` is replaced to be the same as `affinity` and `nodeSelector`.

A re-review is required. Sorry.

library/common-library/templates/_security-context.tpl
```diff
@@ -0,0 +1,40 @@
...
+{{- $localSecurityContext := dict -}}
+{{- if get .Values "podSecurityContext" | kindIs "map" -}}
+    {{- $localSecurityContext = .Values.podSecurityContext -}}
+{{- end -}}
+
+{{- $globalSecurityContext := dict -}}
+{{- $global := index .Values "global" | default dict -}}
+{{- if get $global "podSecurityContext" | kindIs "map" -}}
+    {{- $globalSecurityContext = $global.podSecurityContext -}}
+{{- end -}}
```
gsanchezgavier, line 33: to me is a bit verbose just to avoid a nesting level below (and perhaps because I'm not used to program in tpl), but i'm fine since we have the test now 😃

kang-makes: It tests that exists, is a map, and does not fail in case of being a null (which you reminded me that there is no test for that.

So I prefer that verbosity than having to see if `sprig` does a `default` left to right or right to left...

gsanchezgavier: In this case if not a map it will no fail during installation but the parameters are not going to be added (silently), but in case we just copy the yaml under the parameter and is not a map it will fail during installation. Am I right?

kang-makes:
As @gsanchezgavier pointed out, I was testing too much in the helper so I tested if the user put a map or not.

That means that if the user put an (incorrect) array it is silently discarded and will render without it, making the user think that it has the correct configuration when it has not.

That being said, I rebased (again) and squashed everything to have a better history.

Re-re-review is required. I am really sorry.

library/common-library/templates/_security-context.tpl
```diff
@@ -0,0 +1,22 @@
+{{- /* Defines the container securityContext context */ -}}
+{{- define "common.securityContext.container" -}}
+{{- $global := index .Values "global" | default dict -}}
+
+{{- if .Values.containerSecurityContext -}}
```
paologallinaharbur, line 5: How can we provide default in charts without discarding alway the global one?

Example:
 - http://github.com/newrelic/helm-charts/blob/ecf47c05c7aee6c5b04500a57c643109b36fce7a/charts/nri-metadata-injection/values.yaml#L58-L58
 - http://github.com/newrelic/helm-charts/blob/ecf47c05c7aee6c5b04500a57c643109b36fce7a/charts/nri-metadata-injection/templates/admission-webhooks/job-patch/job-patchWebhook.yaml#L47-L47`

Even with the merge I imagine that with default we would always overwrite globals

gsanchezgavier (approved):
LGTM , but looks that are some pending topics

library/common-library/templates/_security-context.tpl
roobre: This is an interesting topic. I see three solutions, and I like neither of them:

1. Provide the chart "defaults" in the template rather than in the values. So Global values will override the template defaults, and local values will override globals.
2. Make global override local for this particular case
3. Add a boolean "helper" somewhere, like `global.securityContext.override` (defaulting to `false`) that will cause the global SC to override the local one when `true`.

Of all this I think the one I dislike less is 1️⃣ .

roobre (changes requested):
Flagging changes requested so we don't merge this by accident until we solve the defaults problem

library/common-library/templates/_security-context.tpl
paologallinaharbur: I agree, I do not like 3 since it is not easy to be applied to other similar cases in the future.

I really dislike 2 since it would be absolutely counterintuitive and difficult to document.

I would go for the 1st as well

roobre: @newrelic/coreint Any last thought on this?
I think we can go for 1️⃣ tomorrow if no one has a better approach.

gsanchezgavier: i can think of some but non less ugly than the 1

roobre: I think the helper returning defaults is needed for booleans (like `hostNetwork`) as we cannot tell if the helper is returning empty or false, but in this case I think it might be more complex. We can simply do:

```yaml
{{- if include "common.securityContext.pod" . }}
  {{- include "common.securityContext.pod" . | toYaml | nindent 2 }}
{{- else }}
  runAsUser: 1000
  runAsGroup: 3000
  fsGroup: 2000
{{- end }}
```

Which is functionally the same as the `common.securityContext.podDefaults`, but might be more readable. WDYT?

kang-makes: I think the helper would be useful because of ugly things like the discussion we have to raise again:
- [newrelic-infrastructure-v3] Change hostNetwork depending if their kind is DaemonSet or Deployment [#698 ](https://github.com/newrelic/helm-charts/pull/698)
- newrelic-infrastructure-v3: default to recreate strategy for controlPlane Deployment [#697](https://github.com/newrelic/helm-charts/pull/697)

If we need something like "change defaults depending on something external" is better to have a fallback helper that can be overridable and it can allow me to change defaults in case of other flags changing (like `privileged` which is the next PR I would have to do)

Having so many moving parts is the reason I prefer the 3️⃣. Because changing defaults are evil but we are doing it in all the Helm charts because of UX.

kang-makes: The [latest commit](https://github.com/newrelic/helm-charts/pull/725/commits/0f4c594ecf9d3fcde4d859e4c41aedbd1fa4ca87) adds the ability to set defaults using a helper function because it is easier to revert it after finishing the discussion than do it later and have to wait for approvals.

Give me your thoughts, please.

library/common-library/templates/_security-context.tpl
```diff
@@ -18,5 +25,12 @@
     {{- toYaml .Values.podSecurityContext -}}
 {{- else if $global.podSecurityContext -}}
     {{- toYaml $global.podSecurityContext -}}
+{{- else -}}
+    {{- include "common.securityContext.podDefaults" . -}}
```
alvarocabanas, line 29: Oks, so in the values there will be all the block in podDefaults and only honoured if others not set, I like this solution. I think it's clear for the users looking at the values and covers the need we had.

library/common-library/templates/_security-context.tpl
```diff
@@ -0,0 +1,36 @@
...
+    {{- toYaml .Values.containerSecurityContext -}}
+{{- else if $global.containerSecurityContext -}}
+    {{- toYaml $global.containerSecurityContext -}}
+{{- else -}}
+    {{- include "common.securityContext.containerDefaults" . -}}
+{{- end -}}
+{{- end -}}
+
+
+{{- /* Allows to change defaults for container security context either staic of dinamically */ -}}
```
marcsanmi, line 15:
```suggestion
{{- /* Allows to change defaults for container security context either statically or dynamically */ -}}
```

library/common-library/templates/_security-context.tpl
```diff
@@ -0,0 +1,36 @@
...
+    {{- toYaml .Values.podSecurityContext -}}
+{{- else if $global.podSecurityContext -}}
+    {{- toYaml $global.podSecurityContext -}}
+{{- else -}}
+    {{- include "common.securityContext.podDefaults" . -}}
+{{- end -}}
+{{- end -}}
+
+
+{{- /* Allows to change defaults for pod security context either staic of dinamically */ -}}
```
marcsanmi, line 34:
```suggestion
{{- /* Allows to change defaults for pod security context either statically or dynamically */ -}}
```

library/common-library/templates/_security-context.tpl
kang-makes: Don't judge me, it is being a really hard Friday.

Solved in [#35fc122](https://github.com/newrelic/helm-charts/pull/725/commits/35fc12267613cb1f15d2e067f72262d70ec5f9dc)

library/common-library/templates/_security-context.tpl
kang-makes: Don't judge me, it is being a really hard Friday.

Solved in [#35fc122](https://github.com/newrelic/helm-charts/pull/725/commits/35fc12267613cb1f15d2e067f72262d70ec5f9dc)

library/common-library/templates/_security-context.tpl
kang-makes: `values.yaml` should be empty for the `.global` to work but the default can change using other helpers like the `privileged` which I am waiting for this to be merged to start working on it.

roobre (approved):
Looking great, I really like the final approach. Good job!

PR #726, opened 2022-03-04
[common-library] priorityClassName

paologallinaharbur:
#### Is this a new chart
no
adding priorityClassName to common library

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/common-library/templates/_priority-class-name.tpl
+++ b/library/common-library/templates/_priority-class-name.tpl
@@ -1,0 +1,10 @@
+{{- /* Defines the pod priorityClassName */ -}}
+{{- define "common.priorityClassName" -}}
+    {{- if .Values.priorityClassName -}}
+        {{- .Values.priorityClassName -}}
+    {{- else if .Values.global -}}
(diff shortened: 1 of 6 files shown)


gsanchezgavier (approved):
lgtm

library/common-library/templates/_priority-class-name.tpl
```diff
@@ -0,0 +1,8 @@
+{{- /* Defines the pod priorityClassName */ -}}
+{{- define "common.priorityClassName" -}}
+    {{- if .Values.priorityClassName -}}
+        {{- .Values.priorityClassName -}}
+    {{- else if .Values.global.priorityClassName -}}
```
kang-makes, line 5: Test here if global exists.

As we are in a library the values of the chart that are using it are unreliable. In fact, this is the thing why the idea of Guillermo on having a unit-test chart liked me so much because we could have an "empty values file" to test these kinds of things.

kang-makes (approved):
LGTM.

PR #728, opened 2022-03-04
[common-library] labels

paologallinaharbur:
#### Is this a new chart
no
add support in common library for podLabels and labels

#### Checklist
- [x] Chart Version bumped
- [] Variables are documented in the README.md

### DIFF
--- a/library/CHART-TEMPLATE/tests/common_library_labels_test.yaml
+++ b/library/CHART-TEMPLATE/tests/common_library_labels_test.yaml
@@ -1,6 +1,7 @@
 suite: test labels helpers
 templates:
-  - templates/deployment.yaml # we could check the same for service and serviceAccount as well
+  - templates/deployment.yaml
+  - templates/serviceaccount.yaml
 release:
   name: my-release
   namespace: my-namespace
@@ -11,11 +12,16 @@ tests:
   - it: does not fail if globals=nil
     set:
       global: null
+      serviceAccount:
+        create: true
     asserts:
       - hasDocuments:
           count: 1
+
   - it: sets labels according to values
     set:
+      serviceAccount:
+        create: true
       labels:
         helm.sh/chart: fixed
     asserts:
@@ -26,9 +32,12 @@ tests:
             app.kubernetes.io/managed-by: Helm
             app.kubernetes.io/name: CHART-TEMPLATE
             app.kubernetes.io/version: 1.16.0
-            helm.sh/chart: fixed   # We cannot test behaviour without overwriting this since the version chart often changes
+            helm.sh/chart: fixed  # We cannot test behaviour without overwriting this since the version chart often changes
+
   - it: sets labels with globals giving priority to values
     set:
+      serviceAccount:
+        create: true
       labels:
         helm.sh/chart: fixed
         one: overwriting
--- a/library/CHART-TEMPLATE/tests/common_library_labels_test.yaml
+++ b/library/CHART-TEMPLATE/tests/common_library_labels_test.yaml
@@ -1,0 +1,49 @@
+suite: test labels helpers
+templates:
+  - templates/deployment.yaml # we could check the same for service and serviceAccount as well
+release:
+  name: my-release
+  namespace: my-namespace
+set:
+  licenseKey: test-license-key
+  cluster: test-cluster
+tests:
+  - it: does not fail if globals=nil
+    set:
+      global: null
+    asserts:
+      - hasDocuments:
+          count: 1
+  - it: sets labels according to values
+    set:
+      labels:
+        helm.sh/chart: fixed
+    asserts:
+      - equal:
+          path: metadata.labels
+          value:
+            app.kubernetes.io/instance: my-release
+            app.kubernetes.io/managed-by: Helm
+            app.kubernetes.io/name: CHART-TEMPLATE
+            app.kubernetes.io/version: 1.16.0
+            helm.sh/chart: fixed   # We cannot test behaviour without overwriting this since the version chart often changes
+  - it: sets labels with globals giving priority to values
+    set:
+      labels:
+        helm.sh/chart: fixed
+        one: overwriting
+      global:
+        labels:
+          one: overwritten
+          two: second
+    asserts:
+      - equal:
+          path: metadata.labels
+          value:
+            app.kubernetes.io/instance: my-release
+            app.kubernetes.io/managed-by: Helm
+            app.kubernetes.io/name: CHART-TEMPLATE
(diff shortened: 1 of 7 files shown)


library/common-library/templates/_labels.tpl
```diff
@@ -16,6 +16,12 @@ helm.sh/chart: {{ include "common.naming.chart" . }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- with .Values.global.labels }}
```
kang-makes, line 19: It wouldn't be better if we `merge` them "just in case"?

library/common-library/templates/_labels.tpl
```diff
@@ -16,6 +16,12 @@ helm.sh/chart: {{ include "common.naming.chart" . }}
...
+{{ toYaml . }}
+{{- end -}}
+{{- with .Values.labels }}
+{{ toYaml . }}
+{{- end -}}
```
kang-makes, line 24: It wouldn't be better if we merge them "just in case"?

library/CHART-TEMPLATE/tests/common_library_labels_test.yaml
```diff
@@ -0,0 +1,44 @@
...
+  namespace: my-namespace
+set:
+  licenseKey: test-license-key
+  cluster: test-cluster
+tests:
+  # We cannot test behaviour without overwriting this since the version chart often changes
+  - it: sets labels according to values
+    set:
+      labels:
+        helm.sh/chart: fixed
```
kang-makes, line 15:
```suggestion
  - it: sets labels according to values
    set:
      labels:
        helm.sh/chart: fixed  # We cannot test behavior without overwriting this since the version chart often changes
```

library/common-library/templates/_labels.tpl
kang-makes: Well, we would have to do more magic creating a dict with `app.kubernetes.io/version`, `app.kubernetes.io/managed-by`,... so we can merge that too.

Maybe it is a bit overkill. WDYT?

kang-makes (approved):
LGTM. I left a couple of nits that are only style-related.

library/common-library/templates/_labels.tpl
kang-makes, line 19: After visiting the `priorityClass` PR I have to reject this PR because I overlooked this.

We cannot trust that global exists.

kang-makes (changes requested):
I overlooked an issue. Sorry :(

library/common-library/templates/_labels.tpl
paologallinaharbur: Do you mean that Values.global can be nil, right?

library/CHART-TEMPLATE/tests/common_library_labels_test.yaml
```diff
@@ -0,0 +1,44 @@
...
+    asserts:
+      - equal:
+          path: metadata.labels
+          value:
+            app.kubernetes.io/instance: my-release
+            app.kubernetes.io/managed-by: Helm
+            app.kubernetes.io/name: CHART-TEMPLATE
+            app.kubernetes.io/version: 1.16.0
+            helm.sh/chart: fixed
```
gsanchezgavier, line 24: it would be nice just to check if this label exist, just to have an independent test, but i guess that it cannot be done

library/common-library/templates/_labels.tpl
paologallinaharbur: - yes, I feel it would be a bit obscure to create a dict with `app.kubernetes.io/version, app.kubernetes.io/managed-by`. If needed anyone can still overriding them

 - regarding the merge I followed the approach of:
```
{{- if include "common.labels.overrides.addLabels" . }}
{{ include "common.labels.overrides.addLabels" . }}
{{- end }}
```
since it was simpler. We would need to merge 4 sources (standard, user, global, override) and it could be a bit confusing. But If you think it is better I am fine with changing it

library/common-library/templates/_labels.tpl
paologallinaharbur: I'll fix that and check why tests were not covering it

library/CHART-TEMPLATE/tests/common_library_labels_test.yaml
paologallinaharbur: we can, but it has the chart version in it, therefore we would need to update the test each time 🤯  
```
{{- define "common.naming.chart" -}}
{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
{{- end }}
```

library/CHART-TEMPLATE/tests/common_library_labels_test.yaml
```diff
@@ -0,0 +1,43 @@
...
+set:
+  licenseKey: test-license-key
+  cluster: test-cluster
+tests:
+  - it: sets labels according to values
+    set:
+      labels:
+        helm.sh/chart: fixed
+    asserts:
+      - equal:
```
gsanchezgavier, line 16: what about using [contains](https://github.com/vbehar/helm3-unittest/blob/master/unittest/assertion.go#L125) . So we just check what the test says. ( I never tried this `contains` so it might be not the thing we want)
```suggestion
      - contains:
           path: metadata.labels
           value:
             helm.sh/chart: fixed
```

paologallinaharbur: this would work!

paologallinaharbur: actually no, contains does not work as I expected 😖

kang-makes: As the documentation is almost inexistent I use this example as a guide: https://github.com/quintush/helm-unittest/blob/master/test/data/v3/basic/tests_failed/deployment_test.yaml

`contains` is for arrays.

kang-makes (approved):
I am owning this PR to free Paolo during this week.

As I rejected this PR now I have to approve it but I would love to have other people to review it again.

roobre (commented):
Looking nice to me, but seeing this many label-related functions makes me think that, as a chart writer, I'd have to spend quite some time figuring how they work or how am I supposed to use them.
Do you think it would be possible to have this documented somewhere?

kang-makes:
I rebased and added more comments to the helpers but I don't know of any tool that creates documentation from the helper's comments.

I think we are on our own on this...

PR #730, opened 2022-03-05
[nri-bundle] Update README.md

maxlemieux:
#### Is this a new chart
no

This change fixes the link in the readme to point to the new location for the `kube-state-metrics` helm chart.

The kube-state-metrics chart moved to the prometheus-community org as of v2.13: https://github.com/kubernetes/kube-state-metrics/releases/tag/kube-state-metrics-helm-chart-2.13.3

Old link was 404

#### Checklist

### DIFF
--- a/charts/nri-bundle/README.md
+++ b/charts/nri-bundle/README.md
@@ -166,7 +166,7 @@ helm install newrelic/nri-bundle \
 ```
 
 [1]: https://docs.newrelic.com/docs/accounts/install-new-relic/account-setup/license-key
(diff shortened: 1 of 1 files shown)


kang-makes:
Hi @maxlemieux,

Thanks for reaching out with this. but we are not using the Prometheus community helm chart, yet because it is based on KSM v2 we still do not support.

We have this in our roadmap but there is no ETA yet. If you want to reopen this issue with a different link that points to the v1 chart that was in that repo I would be happy to merge it 😄

Regards.

PR #734, opened 2022-03-07
[newrelic-infrastructure-v3] bump app version

paologallinaharbur:
#### Is this a new chart
no

bump app version

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-infrastructure-v3/Chart.yaml
+++ b/charts/newrelic-infrastructure-v3/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: newrelic-infrastructure-v3
 description: A Helm chart to deploy the New Relic Kubernetes monitoring solution
-version: 3.0.21
-appVersion: 3.0.0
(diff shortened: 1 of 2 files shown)


charts/newrelic-infrastructure-v3/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: newrelic-infrastructure-v3
 description: A Helm chart to deploy the New Relic Kubernetes monitoring solution
-version: 3.0.21
```
roobre, line 4: Should we bump the minor if the app minor was updated?

roobre (approved):

charts/newrelic-infrastructure-v3/Chart.yaml
paologallinaharbur: so far till is beta we were bumping patch only, but you are right, better the minor

PR #735, opened 2022-03-07
[common-library] add global.nodeSelector 

gsanchezgavier:
#### Is this a new chart
No
Adds `common.nodeSelector` helper to common library
  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/CHART-TEMPLATE/tests/common_library_nodeselector_test.yaml
+++ b/library/CHART-TEMPLATE/tests/common_library_nodeselector_test.yaml
@@ -1,0 +1,36 @@
+suite: test nodeSelector helpers
+templates:
+  - templates/deployment.yaml
+release:
+  name: my-release
+  namespace: my-namespace
+set:
+  licenseKey: test-license-key
+  cluster: test-cluster
+tests:
+  - it: does not set nodeSelector if not specified
+    set:
+    asserts:
+      - isNull:
+          path: spec.template.spec.nodeSelector
+  - it: does set nodeSelector if specified in values
+    set:
+      nodeSelector:
+        key: value
+    asserts:
+      - equal:
+          path: spec.template.spec.nodeSelector
+          value:
+            key: value
+  - it: overwrites nodeSelector from globals when specified in values
+    set:
+      global:
+        nodeSelector:
+          keyOnGlobal: foo
+      nodeSelector:
+        keyOnValues: bar
+    asserts:
+      - equal:
+          path: spec.template.spec.nodeSelector
+          value:
+            keyOnValues: bar
--- a/library/CHART-TEMPLATE/tests/common_library_nodeselector_test.yaml
+++ b/library/CHART-TEMPLATE/tests/common_library_nodeselector_test.yaml
@@ -10,13 +10,27 @@ set:
 tests:
   - it: does not set nodeSelector if not specified
     set:
+      global: null
+      nodeSelector: null
     asserts:
       - isNull:
           path: spec.template.spec.nodeSelector
   - it: does set nodeSelector if specified in values
     set:
+      global: null
       nodeSelector:
         key: value
+    asserts:
+      - equal:
+          path: spec.template.spec.nodeSelector
+          value:
+            key: value
+  - it: does set nodeSelector if specified in global
+    set:
+      global:
+        nodeSelector:
+          key: value
+      nodeSelector: null
     asserts:
       - equal:
           path: spec.template.spec.nodeSelector
--- a/library/common-library/templates/_nodeselector.tpl
+++ b/library/common-library/templates/_nodeselector.tpl
@@ -1,0 +1,10 @@
+{{- /* Defines the Pod nodeSelector */ -}}
+{{- define "common.nodeSelector" -}}
+    {{- if .Values.nodeSelector -}}
+        {{- toYaml .Values.nodeSelector -}}
+    {{- else if .Values.global -}}
+        {{- if .Values.global.nodeSelector -}}
+            {{- toYaml .Values.global.nodeSelector -}}
+        {{- end -}}
+    {{- end -}}
+{{- end -}}
--- a/library/CHART-TEMPLATE/Chart.lock
+++ b/library/CHART-TEMPLATE/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common-library
   repository: file://../common-library
-  version: 0.2.1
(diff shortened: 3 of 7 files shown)


roobre (approved):

gsanchezgavier:
@roobre  just pushed a commit with an empty values set in the test to check for nil pointers. Not sure if this is the best approach or just comment the `global` in the CHART-TEMPLATE `values.yaml` . wdyt? also @kang-makes

library/CHART-TEMPLATE/tests/common_library_nodeselector_test.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - it: does not set nodeSelector if not specified
+    values:
+      - ./empty-values.yaml
+    set:
+    asserts:
+      - isNull:
+          path: spec.template.spec.nodeSelector
+  - it: does set nodeSelector if specified in values
+    values:
+      - ./empty-values.yaml
```
paologallinaharbur, line 20: you can set as well:
```
 - it: does not break if global=null
    set:
      global: null
    asserts:
      - equal:
[...]
```
This approach is better merely in the case we need ot test "weird" cases: for example "global.image" being null

gsanchezgavier: yes, i just changed it to that. 👍

library/common-library/templates/_nodeselector.tpl
```diff
@@ -0,0 +1,10 @@
+{{- /* Defines the Pod nodeSelector */ -}}
+{{- define "common.nodeSelector" -}}
+    {{- if .Values.nodeSelector -}}
+        {{- toYaml .Values.nodeSelector -}}
+    {{- else if .Values.global -}}
+        {{- if .Values.global.nodeSelector -}}
+            {{- toYaml .Values.global.nodeSelector -}}
+        {{- end -}}
+    {{- end -}}
```
kang-makes, line 9: This could be merged. Take a look at this: https://github.com/newrelic/helm-charts/pull/725/files#diff-77ece6d92f1f208ef9ef4af7ab318c4ba282722510480e737f32dd0bccb10794R24-R39

kang-makes (commented):
`nodeSelector` is a map and we could merge them :)

library/common-library/templates/_nodeselector.tpl
gsanchezgavier: I did not merge it to allow a re-write from the values when a custom node selector is required for a particular component, but actually both use cases might be valid.

kang-makes: I thought about doing that not only in `nodeSelector` but everywhere, like in `securityContext` but I realized that it would be evil because that overrides the way Helm works.

If you use `helm upgrade something . -f values-1. yaml -f values-2. yaml -f values-3. yaml` the values are always merged.

This is sugar for customers to ease the installation but I would prefer to don't do magic under the hood. WDYT?

kang-makes: Maybe @roobre who approved this has something to say...

gsanchezgavier: But if you replace a values in helm you will expect that is not merged right. I think this is a matter of precedence and is not changing the behavior of Helm. On the opposite doing merge it might be the magic in this case.

roobre: I think Helm is a lot about merging, but mostly in layers of values and not that often whithin values themselves. Do we know if there is a community precedent where `globals` are merged with non-globals? At least I haven't seen it before and I'd rather not break users expectations on this. "A little copy is better than a little unexpected behavior".

kang-makes: No, if it is a map it merge them:
```
% helm create test
Creating test
% cd test
% vim f1.yaml
% sed -e s/1/2/g f1.yaml > f2.yaml
% bat f*
───────┬──────────────
       │ File: f1.yaml
───────┼──────────────
   1   │ nodeSelector:
   2   │   test-01: is-01
   3   │
───────┴──────────────
───────┬──────────────
       │ File: f2.yaml
───────┼──────────────
   1   │ nodeSelector:
   2   │   test-02: is-02
   3   │
───────┴──────────────
```
This is how it renders with on1 value file:
```
% helm template test . -s templates/deployment.yaml -f f1.yaml | tail -5
              port: http
          resources:
            {}
      nodeSelector:
        test-01: is-01
```
This is how it renders with two:
```
% helm template test . -s templates/deployment.yaml -f f1.yaml -f f2.yaml| tail -5
          resources:
            {}
      nodeSelector:
        test-01: is-01
        test-02: is-02
```

kang-makes: Answering Roberto no. I don't know anything that uses this kind of global.

That is because normally projects do not have more than 2 Helm charts but this is a really complex software to deploy. These kinds of globals are not seen frequently. At least I do not know more projects that use them. But I also down known any project that has 6 helm charts on an umbrella...

As I said, this is to ease user's configuration. I would expect helm to merge so I thought that it would be better to merge this too.

If you say that you would not merge this maybe we have to revisit `securityContext` PR. WDYT?

gsanchezgavier: oh i understand know our different point of view. I see globals values like a "parent" and not like a values at the same level.
I'm fine with both, i think both covers different use cases and if we follow one some of the use cases of the other are not going to be covered and viceversa (and probably mix them doesn't make much sense). But we need to document the behavior since it could be miss-understanding.

roobre: My main worry would be that if we set those to be merged, it might become hard to the user to "erase" the globals and revert to the defaults. Helm does provide _some_ erasing capabilities with `null` but it's [known to have limitations](https://github.com/helm/helm/issues/9136). I'd lean towards KISS here.

paologallinaharbur: I prefer as well the "if-else" behaviour as we discussed in the first PR merely because it is easier to debug and to document. But I see the point of having the merge behaviour and guillermo said. 

So far we went for the merge (labels, security context, annotations) and the append for image pull secrets. Therefore we can follow the same approach.
 - The only thing we should make sure is that with the merge we are able to overwrite completely the global, even unsetting it

kang-makes (approved):
We have a consensus then. We would have to follow an if-else approach.

This LGTM and I would change the PRs I have open that makes a merge.

roobre (approved):

PR #739, opened 2022-03-08
[common-library] Fix local service account name

alvarocabanas:
When setting a .values service account name should take precedence to the default one

### DIFF
--- a/library/CHART-TEMPLATE/tests/common_library_serviceaccount_name_test.yaml
+++ b/library/CHART-TEMPLATE/tests/common_library_serviceaccount_name_test.yaml
@@ -44,3 +44,13 @@ tests:
       - equal:
           path: spec.template.spec.serviceAccountName
           value: default
+
+  - it: allow to override the serviceacocunt name with the local one
+    set:
+      serviceAccount:
+        name: "my-local-service-account-name"
+        create: true
+    asserts:
+      - equal:
+          path: spec.template.spec.serviceAccountName
+          value: my-local-service-account-name
--- a/library/CHART-TEMPLATE/tests/common_library_serviceaccount_name_test.yaml
+++ b/library/CHART-TEMPLATE/tests/common_library_serviceaccount_name_test.yaml
@@ -49,6 +49,16 @@ tests:
     set:
       serviceAccount:
         name: "my-local-service-account-name"
+        create: false
+    asserts:
+      - equal:
+          path: spec.template.spec.serviceAccountName
+          value: my-local-service-account-name
+
+  - it: allow to override the serviceacocunt name with the local one
+    set:
+      serviceAccount:
+        name: "my-local-service-account-name"
         create: true
     asserts:
       - equal:
--- a/library/CHART-TEMPLATE/tests/common_library_serviceaccount_name_test.yaml
+++ b/library/CHART-TEMPLATE/tests/common_library_serviceaccount_name_test.yaml
@@ -45,8 +45,34 @@ tests:
           path: spec.template.spec.serviceAccountName
           value: default
 
+  - it: allow to override the serviceacocunt name with the global one
+    set:
+      global:
+        serviceAccount:
+          name: "my-global-service-account-name"
+          create: false
+    asserts:
+      - equal:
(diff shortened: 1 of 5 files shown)


library/common-library/templates/_serviceaccount.tpl
```diff
@@ -47,9 +47,9 @@
 {{- end -}}
 
 {{- if (include "common.serviceAccount.create" .) -}}
-    {{- default (include "common.naming.fullname" .) $globalServiceAccount $localServiceAccount -}}
+    {{- default (include "common.naming.fullname" .) $localServiceAccount $globalServiceAccount -}}
```
paologallinaharbur, line 50: helm function are always confusing since` {{param1 | func param2 }}` and `{{ func param2 param1 }}` follows different order (at least in my head) 

Isn't crearer using pipes with multiple parameters?

library/CHART-TEMPLATE/tests/common_library_serviceaccount_name_test.yaml
```diff
@@ -44,3 +44,13 @@ tests:
...
+
+  - it: allow to override the serviceacocunt name with the local one
+    set:
+      serviceAccount:
+        name: "my-local-service-account-name"
+        create: true
+    asserts:
+      - equal:
+          path: spec.template.spec.serviceAccountName
+          value: my-local-service-account-name
```
kang-makes, line 56: This test is cool and is the test that is missing but now I miss one more test.

Given these values:
```yaml
serviceAccount:
  create: false
  name: test
```
We should also test that the service account is not created and that the deployment is using "test" as a service account.

Could you take care of this in this PR to fix the `serviceAccount` once and for all?

library/common-library/templates/_serviceaccount.tpl
kang-makes: I also see this as a good measure to increase readability.

This is the reason I failed and this issue you are fixing exists.

kang-makes (commented):
Now I see another missing test I hope you have time to create :)

library/common-library/templates/_serviceaccount.tpl
alvarocabanas: sure I can go for it

alvarocabanas: {{- $globalServiceAccount | default ($localServiceAccount | default (include "common.naming.fullname" .)) -}}
or
{{- $globalServiceAccount | default ("common.naming.fullname" $localServiceAccount) -}}

alvarocabanas: `{{- $globalServiceAccount | default $localServiceAccount | default (include "common.naming.fullname" .) -}}`

alvarocabanas: That last one works.. it's the best option.

paologallinaharbur: yes I think the same, I believe it should be `{{- $localServiceAccount | default $globalServiceAccount | default (include "common.naming.fullname" .) -}}`, unless helm tricks me again 😄

alvarocabanas: I am putting the global as the most important one, then the local and then the default. it's intended, and it's what I did in the first change is it incorrect? does the local have to be more important than the global?

alvarocabanas: Oks I changed the order and added extra tests

library/CHART-TEMPLATE/tests/common_library_serviceaccount_name_test.yaml
alvarocabanas: done

library/common-library/templates/_serviceaccount.tpl
kang-makes: The most important is the local. In case local does not exist then the global should default. Then default to fullname/`default`.

alvarocabanas: Yes, it's fixed now

kang-makes (approved):
LGTM. Merge before you have to rebase it again ;)

PR #747, opened 2022-03-14
[common-library] Add common.fedramp

gsanchezgavier:
#### Is this a new chart
No
Adds fedramp helper to library
  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/CHART-TEMPLATE/templates/deployment.yaml
+++ b/library/CHART-TEMPLATE/templates/deployment.yaml
@@ -57,6 +57,9 @@ spec:
             - name: MY_APP_PROXY_URL
               value: {{ . | quote }}
           {{- end }}  
+          {{- if include "common.fedramp.enabled" . }}
+            - name: MY_APP_CUSTOM_ENDPOINT
+          {{- end }}
       {{- with include "common.nodeSelector" . }}
       nodeSelector:
         {{- . | nindent 8 }}
--- a/library/CHART-TEMPLATE/tests/common_library_fedramp_test.yaml
+++ b/library/CHART-TEMPLATE/tests/common_library_fedramp_test.yaml
@@ -1,0 +1,52 @@
+suite: test fedRAMP helpers
+templates:
+  - templates/deployment.yaml
+release:
+  name: my-release
+  namespace: my-namespace
+set:
+  licenseKey: test-license-key
+  cluster: test-cluster
+tests:
+  - it: does not set fedRAMP if not specified
+    set:
+      global: null
+      fedramp: null
+    asserts:
+      - notContains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: MY_APP_CUSTOM_ENDPOINT
+  - it: does set fedRAMP if enabled in values
+    set:
+      global: null
+      fedramp:
+        enabled: true
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: MY_APP_CUSTOM_ENDPOINT
+  - it: does set fedRAMP if enabled in global
+    set:
+      global: 
+        fedramp:
+          enabled: true
+      fedramp: null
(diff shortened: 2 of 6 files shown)


library/CHART-TEMPLATE/tests/common_library_fedramp_test.yaml
```diff
@@ -0,0 +1,52 @@
...
+  name: my-release
+  namespace: my-namespace
+set:
+  licenseKey: test-license-key
+  cluster: test-cluster
+tests:
+  - it: does not set fedRAMP if not specified
+    set:
+      global: null
+      proxy: null
```
paologallinaharbur, line 14: typo

paologallinaharbur, line 14: typo

library/CHART-TEMPLATE/templates/deployment.yaml
```diff
@@ -49,6 +49,11 @@ spec:
               port: http
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+          env:
+            - name: MY_APP_NON_OPTIONAL_ENV_VAR
```
paologallinaharbur, line 53: why no value for the env? 


in nri-prometheus for example the approach is different:

https://github.com/newrelic/helm-charts/blob/afd566c86e2d336c08d52edc7ab5a60fae6859b7/charts/nri-prometheus/values.yaml#L73



The user is required to add the url, it is not enough just to enable a flag

paologallinaharbur, line 53: why no value for the env? 


in nri-prometheus for example the approach is different:

https://github.com/newrelic/helm-charts/blob/afd566c86e2d336c08d52edc7ab5a60fae6859b7/charts/nri-prometheus/values.yaml#L73



The user is required to add the url, it is not enough just to enable a flag

paologallinaharbur: other chart deal with it in a different way?

gsanchezgavier: but in some other apps there are more than 1 endpoint. for instance the agent has 3 endpoints. And even if they would have just one endpoint to configure they are usually different (meaning DM endpoint, Otel endpoint , ... )
This is why i followed the boolean approach

gsanchezgavier: answered above

gsanchezgavier: I also though about something like this 
```yaml
fedramp:
  enabled: true
  metrics_api_url:
  events_api_url:
  ...
```
but it would required some internals knowledge of each component that we could save from users i think.

library/CHART-TEMPLATE/tests/common_library_fedramp_test.yaml
gsanchezgavier: thanks!

library/CHART-TEMPLATE/templates/deployment.yaml
paologallinaharbur: So the idea would be to have in global just the enable and in each chart the url value?

gsanchezgavier: yes so instead of configuring each endpoint manually in each subchart , just enabled this flag and all subchart could set this for you with the default values or [flags](https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings#fedramp) depending on the sub chart .

paologallinaharbur: ok, makes sense!

paologallinaharbur (approved):

PR #748, opened 2022-03-14
[common library] Add common library cluster helper

alvarocabanas:
Add common library cluster helper

### DIFF
--- a/library/CHART-TEMPLATE/templates/deployment.yaml
+++ b/library/CHART-TEMPLATE/templates/deployment.yaml
@@ -35,9 +35,6 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: {{ include "common.images.image" ( dict "imageRoot" .Values.image "context" .) }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          env:
-            - name: CLUSTER_NAME
-              value: {{ include "common.cluster" . }}
           ports:
             - name: http
               containerPort: 80
@@ -53,6 +50,8 @@ spec:
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           env:
+            - name: CLUSTER_NAME
+              value: {{ include "common.cluster" . }}
             - name: MY_APP_NON_OPTIONAL_ENV_VAR
           {{- with include "common.proxy" . }}
             - name: MY_APP_PROXY_URL


library/common-library/templates/_cluster.tpl
```diff
@@ -0,0 +1,17 @@
...
+Return the cluster
+*/}}
+{{- define "common.cluster" -}}
+    {{- if .Values.cluster -}}
+        {{- .Values.cluster -}}
+    {{- else if .Values.global -}}
+        {{- if .Values.global.cluster -}}
+            {{- .Values.global.cluster -}}
+        {{ else }}
+            {{- "" -}}
```
roobre, line 11: Do we need these `else ""` constructs? In my past experience it was fine to simply omit them, Helm will render nothing by default which should be evaluated to `false`.

library/CHART-TEMPLATE/tests/common_images_pull_secrets_test.yaml
```diff
@@ -4,17 +4,18 @@ templates:
 release:
   name: my-release
   namespace: my-namespace
-set:
 tests:
   - it: does not set imagePullSecrets if not specified
     set:
+      cluster: test-cluster
```
roobre, line 10: Why do we need to set `cluster: test-cluster` for _all_ test cases?

library/common-library/templates/_cluster.tpl
kang-makes: The helper will return an empty string if you print nothing. This is redundant. No helper has those so I would remove this.

kang-makes (commented):
I just pointed a few nits :)

library/common-library/templates/_cluster.tpl
alvarocabanas: oks, I will remove it

library/CHART-TEMPLATE/tests/common_images_pull_secrets_test.yaml
alvarocabanas: removed

library/common-library/templates/_cluster.tpl
alvarocabanas: done

library/CHART-TEMPLATE/tests/common_cluster_test.yaml
```diff
@@ -0,0 +1,30 @@
...
+release:
+  name: my-release
+  namespace: my-namespace
+set:
+tests:
+  - it: sets null as cluster name if not specified
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[0].env[0].value
+          value: null
```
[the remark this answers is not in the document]
alvarocabanas: done

library/common-library/templates/_cluster.tpl
```diff
@@ -0,0 +1,13 @@
...
+Return the cluster
+*/}}
+{{- define "common.cluster" -}}
+    {{- if .Values.cluster -}}
+        {{- .Values.cluster -}}
+    {{- else if .Values.global -}}
+        {{- if .Values.global.cluster -}}
+            {{- .Values.global.cluster -}}
+        {{- end -}}
+    {{- end -}}
```
paologallinaharbur, line 11: I believe we discussed this already, therefore I am just double-checking.
We decided not to fail here and to leave the decision to each chart, rigt?

paologallinaharbur (approved):
Just a small doubt, everything else looks good!

library/common-library/templates/_cluster.tpl
alvarocabanas: yes

PR #749, opened 2022-03-15
common-library: add licenseKey related helpers

roobre:
The following helper functions have been added to the common library:
- `common.license.secretName`, returing the name of the secret holding the license key.
- `common.license.secretKeyName`, returing name of the key inside the secret containing the license key.
- `common.license.secret`, which renders a secret if and only if user supplies the license key ad-hoc in the values.

All the logic of handling both globals _and_ custom secret names for the keys are already implemented in the helpers above, so chart writers only have to do two things:

1, include the `common.licenseKey.secret` template:
```
───────┬───────────────────────────────────────────────────────────────────────
       │ File: example-secret.yaml
       │ Size: 44 B
───────┼───────────────────────────────────────────────────────────────────────
   1   │ {{- include "common.licenseKey.secret" . }}
───────┴───────────────────────────────────────────────────────────────────────
```

2, reference it using the helpers:
```yaml
  env:
    - name: NRIA_LICENSE_KEY
      valueFrom:
        secretKeyRef:
          name: {{ include "common.licenseKey.secretName" . }}
          key: {{ include "common.licenseKey.secretKeyName" . }}
```

The library takes care of:

1. Not rendering anything in the secret if the user is providing their own.
2. Returning either the user-provided, or the generated names in `common.licenseKey.secretName` and `common.licenseKey.secretKeyName`, looking for globals.
3. Validating a license key has been provided.

Moreover, an overridable `common.licenseKey.overrides.allowEmpty` helper is defined, which by default returns false. Charts using the library may override the helpr to return true, in which case, the library will not error if a `licenseKey` is not defined in the user values.

### DIFF
--- a/library/CHART-TEMPLATE/templates/deployment.yaml
+++ b/library/CHART-TEMPLATE/templates/deployment.yaml
@@ -50,6 +50,12 @@ spec:
             - name: http
               containerPort: 80
               protocol: TCP
+          env:
+            - name: NRIA_LICENSE_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "common.license.secretName" . }}
+                  key: {{ include "common.license.secretKeyName" . }}
           livenessProbe:
             httpGet:
               path: /
--- a/library/CHART-TEMPLATE/templates/example-cm-licensekey.yaml
+++ b/library/CHART-TEMPLATE/templates/example-cm-licensekey.yaml
@@ -1,0 +1,9 @@
+# This is a dummy CM to test what licenseKey helpers of the common library return
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "common.naming.fullname" . }}-examples
+  namespace: {{ .Release.Namespace }}
+data:
+  licensekey-secret-name: {{ include "common.license.secretName" . }}
+  licensekey-secret-key-name: {{ include "common.license.secretKeyName" . }}
--- a/library/CHART-TEMPLATE/templates/example-secret.yaml
+++ b/library/CHART-TEMPLATE/templates/example-secret.yaml
@@ -1,0 +1,1 @@
+{{- include "common.license.secret" . }}
--- a/library/CHART-TEMPLATE/tests/common_library_license_helpers_test.yaml
+++ b/library/CHART-TEMPLATE/tests/common_library_license_helpers_test.yaml
@@ -1,0 +1,66 @@
+suite: test licenseKey helpers
+templates:
+  - templates/example-cm-licensekey.yaml
+release:
+  name: my-release
+  namespace: my-namespace
+tests:
+  - it: creates secret names for local licenseKey
+    set:
+      licenseKey: local
+    asserts:
+      - equal:
+          path: data.licensekey-secret-name
+          value: nri-my-release-CHART-TEMPLATE-license
+      - equal:
+          path: data.licensekey-secret-key-name
+          value: licenseKey
+  - it: creates secret names for global licenseKey
+    set:
+      global:
+        licenseKey: local
+    asserts:
+      - equal:
+          path: data.licensekey-secret-name
+          value: nri-my-release-CHART-TEMPLATE-license
+      - equal:
+          path: data.licensekey-secret-key-name
+          value: licenseKey
+
+  - it: returns local custom secret names
+    set:
+      customSecretName: local
+      customSecretLicenseKey: localkey
+    asserts:
+      - equal:
+          path: data.licensekey-secret-name
+          value: local
+      - equal:
+          path: data.licensekey-secret-key-name
+          value: localkey
+  - it: returns global custom secret names
+    set:
+      global:
+        customSecretName: global
+        customSecretLicenseKey: globalkey
+    asserts:
+      - equal:
+          path: data.licensekey-secret-name
+          value: global
+      - equal:
+          path: data.licensekey-secret-key-name
+          value: globalkey
+  - it: local secret names override global secret names
+    set:
+      customSecretName: local
+      customSecretLicenseKey: localkey
+      global:
+        customSecretName: global
+        customSecretLicenseKey: globalkey
+    asserts:
+      - equal:
+          path: data.licensekey-secret-name
+          value: local
+      - equal:
+          path: data.licensekey-secret-key-name
+          value: localkey
--- a/library/CHART-TEMPLATE/values.yaml
+++ b/library/CHART-TEMPLATE/values.yaml
@@ -16,6 +16,8 @@ global:
   tolerations: []
   affinity: {}
 
+licenseKey: foobar
+
 replicaCount: 1
 
 image:
--- a/library/common-library/templates/_license.tpl
+++ b/library/common-library/templates/_license.tpl
@@ -1,0 +1,62 @@
+{{/*
+Return the name of the secret holding the License Key.
+*/}}
+{{- define "common.license.secretName" -}}
+{{ include "common.license._customSecretName" . | default (printf "%s-license" (include "common.naming.fullname" . )) }}
+{{- end -}}
+
+{{/*
+Return the name key for the License Key inside the secret.
+*/}}
+{{- define "common.license.secretKeyName" -}}
+{{ include "common.license._customSecretKey" . | default "licenseKey" }}
+{{- end -}}
+
+{{/*
+Charts using this library can implement this function. If it returns a truthy value, common.license.secret will not
+fail if an empty licenseKey is provided.
+*/}}
+{{- define "common.license.overrides.allowEmpty" -}}
+{{- end }}
+
+{{/*
+Return local licenseKey if set, global otherwise.
+This helper is for internal use.
+*/}}
+{{- define "common.license._licenseKey" -}}
+{{- if .Values.licenseKey -}}
+  {{- .Values.licenseKey -}}
+{{- else if .Values.global -}}
+  {{- if .Values.global.licenseKey -}}
+    {{- .Values.global.licenseKey -}}
+  {{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return the name of the secret holding the License Key.
+This helper is for internal use.
+*/}}
+{{- define "common.license._customSecretName" -}}
+{{- if .Values.customSecretName -}}
+  {{- .Values.customSecretName -}}
+{{- else if .Values.global -}}
+  {{- if .Values.global.customSecretName -}}
+    {{- .Values.global.customSecretName -}}
+  {{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return the name key for the License Key inside the secret.
+This helper is for internal use.
+*/}}
+{{- define "common.license._customSecretKey" -}}
+{{- if .Values.customSecretLicenseKey -}}
+  {{- .Values.customSecretLicenseKey -}}
+{{- else if .Values.global -}}
+  {{- if .Values.global.customSecretLicenseKey }}
+    {{- .Values.global.customSecretLicenseKey -}}
+  {{- end -}}
+{{- end -}}
+{{- end -}}
--- a/library/common-library/templates/_license.tpl
+++ b/library/common-library/templates/_license.tpl
@@ -13,13 +13,6 @@ Return the name key for the License Key inside the secret.
 {{- end -}}
 
 {{/*
-Charts using this library can implement this function. If it returns a truthy value, common.license.secret will not
-fail if an empty licenseKey is provided.
-*/}}
-{{- define "common.license.overrides.allowEmpty" -}}
-{{- end }}
-
-{{/*
 Return local licenseKey if set, global otherwise.
 This helper is for internal use.
 */}}
--- a/library/common-library/templates/_license_secret.yaml.tpl
+++ b/library/common-library/templates/_license_secret.yaml.tpl
@@ -5,9 +5,7 @@ Renders the license key secret if user has not specified a custom secret.
 {{- if not (include "common.license._customSecretName" .) }}
 {{- /* Fail if licenseKey is empty and required: */ -}}
 {{- if not (include "common.license._licenseKey" .) }}
-    {{- if not (include "common.license.overrides.allowEmpty" .) }}
     {{- fail "You must specify a licenseKey or a customSecretName containing it" }}
-    {{- end }}
 {{- end }}
 ---
 apiVersion: v1
--- a/library/common-library/templates/_license_secret.yaml.tpl
+++ b/library/common-library/templates/_license_secret.yaml.tpl
@@ -1,0 +1,23 @@
+{{/*
+Renders the license key secret if user has not specified a custom secret.
+*/}}
+{{- define "common.license.secret" }}
+{{- if not (include "common.license._customSecretName" .) }}
+{{- /* Fail if licenseKey is empty and required: */ -}}
+{{- if not (include "common.license._licenseKey" .) }}
+    {{- if not (include "common.license.overrides.allowEmpty" .) }}
+    {{- fail "You must specify a licenseKey or a customSecretName containing it" }}
+    {{- end }}
+{{- end }}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "common.license.secretName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "common.labels" . | nindent 4 }}
+data:
+  {{ include "common.license.secretKeyName" . }}: {{ include "common.license._licenseKey" . | b64enc }}
+{{- end }}
+{{- end }}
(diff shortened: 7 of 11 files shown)


library/CHART-TEMPLATE/templates/example-secret.yaml
```diff
@@ -0,0 +1 @@
+{{- include "common.licenseKey.secret" . }}
```
paologallinaharbur, line 1: I have mixed feelings about this, I see the point of unifying the secret management, but I do not like especially to have a whole yaml in the helpers.

Have you checked if there is any different among the license secrets of the charts? (I believe there is not)

library/CHART-TEMPLATE/values.yaml
```diff
@@ -15,6 +15,8 @@ global:
   tolerations: []
   affinity: {}
 
+licenseKey: foobar
```
paologallinaharbur, line 18: I would not have this enabled by default in the chart example

library/CHART-TEMPLATE/templates/example-cm-licensekey.yaml
```diff
@@ -0,0 +1,9 @@
+# This is a dummy CM to test what licenseKey helpers of the common library return
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "common.naming.fullname" . }}-examples
+  namespace: {{ .Release.Namespace }}
+data:
+  licensekey-secret-name: {{ include "common.licenseKey.secretName" . }}
```
paologallinaharbur, line 8: I am not following, is it to avoid in unit test to do .spec.template.spec.env[0]...?

library/CHART-TEMPLATE/templates/deployment.yaml
```diff
@@ -39,6 +39,12 @@ spec:
             - name: http
               containerPort: 80
               protocol: TCP
+          env:
+            - name: NRIA_LICENSE_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "common.licenseKey.secretName" . }}
```
paologallinaharbur, line 46: I like the fact that this is custom secret or not it is totally transparent for the user

library/common-library/templates/_secret.yaml.tpl
```diff
@@ -0,0 +1,87 @@
+{{/*
+Return local licenseKey if set, global otherwise.
+This helper is for internal use.
+*/}}
+{{- define "common.licenseKey._licenseKey" -}}
```
paologallinaharbur, line 5: Having in the signature like `common.licenseKey._licenseKey` is common?

Could we call it something like`common.licenseKey.cusomLicenceKey`?

paologallinaharbur: I would move this down, closer to the section where it is actually used

roobre: I don't think it is common :(
Essentially I just wanted a way to very obviously tell chart writers that these helpers are internal, and not intended to be used. But there's probably a better way to do this.

What about `internal.licenseKey.licenseKey`, for example? So swapping `common` with `internal` for internal helpers.

library/CHART-TEMPLATE/values.yaml
roobre: Unfortunately, if we don't do it then the secret will fail to template.

I could get rid of it and implement the `allowEmpty` override though, if you think that's better.

library/common-library/templates/_secret.yaml.tpl
```diff
@@ -0,0 +1,87 @@
...
+
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "common.licenseKey.secretName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "common.labels" . | nindent 4 }}
+data:
+  {{ include "common.licenseKey.secretKeyName" . }}: {{ include "common.licenseKey._licenseKey" . | b64enc }}
```
paologallinaharbur, line 85: Therefore if someone is adding the `customSecretLicenseKey ` we render the secret using such value. 

This is something that would make sense? I would expect that either it is custom or it is not.

paologallinaharbur (commented):
overall I like the approach. The biggest doubt I have is whether we want to "hide" the template in the helper or not.

Naming is also a bit confusing 😕

library/CHART-TEMPLATE/templates/example-secret.yaml
roobre: > I do not like especially to have a whole yaml in the helpers

I also think this is a bit weird, but I think there is no other way to abstract the chart writers from the logic of whether or not create the secret, and which name it should have. This is a bit "darker" but also leaves very little room for mistake to chart writers.

> Have you checked if there is any different among the license secrets of the charts? (I believe there is not)

I think we have a couple of charts that use APIKeys and such. For those, we can either incorporate that logic to this secret, or just use other secret for that.

library/CHART-TEMPLATE/templates/example-cm-licensekey.yaml
roobre: After reading the comment again, yes, this also simplifies having to go deep to the deployment to test.

In fact, I intedned the code in `deployment.yaml` to be just an example... but after reading your comment, we might use it for tests as well. I can change it if you think it's clearer!

library/CHART-TEMPLATE/values.yaml
paologallinaharbur: somehow I read `customLicenseKey`. No need to change that 😄

library/common-library/templates/_secret.yaml.tpl
roobre: I use the `include` here not to allow users to override this key (as it does not make much sense), but to avoid having the hardcoded `default "licenseKey"` in two places.

library/common-library/templates/_secret.yaml.tpl
```diff
@@ -0,0 +1,87 @@
...
+
+{{/*
+Return the name key for the License Key inside the secret.
+*/}}
+{{- define "common.licenseKey.secretKeyName" -}}
+{{ include "common.licenseKey._customSecretKey" . | default "licenseKey" }}
+{{- end -}}
+
+{{/*
+Charts using this library can implement this function. If it returns a truty value, common.licenseKey.secret will not
```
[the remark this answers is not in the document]
roobre: I was thinking about https://github.com/newrelic/helm-charts/pull/543, where the licenseKey is optional :)

library/CHART-TEMPLATE/templates/example-cm-licensekey.yaml
paologallinaharbur: It is fine, I saw that there is also the comment
`# This is a dummy CM to test what licenseKey helpers of the common library return`
explaining what is going on

library/common-library/templates/_secret.yaml.tpl
paologallinaharbur: ok, therefore we allow it, but it is not an issue. Everything should be working in the exact same way

library/common-library/templates/_secret.yaml.tpl
paologallinaharbur: what I found confusing was the double "licenseKey.licenseKey". the underscore was actually fine to show that it is internal.

It would be nice to have "unexported" function with with `_`

library/CHART-TEMPLATE/templates/example-secret.yaml
paologallinaharbur: I see the point of having it and we can mark is as solved, I was just a bit worried that is quite different from all templates we own. 

Maybe @kang-makes has some ideas about this.

library/common-library/templates/_secret.yaml.tpl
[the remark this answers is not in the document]
paologallinaharbur: Therefore I always sponsor to add feature whenever we actually need them, otherwise there is the risk of having this there for ever and never actually using it

library/CHART-TEMPLATE/templates/example-secret.yaml
roobre: Oh no, you summoned the nitpicking beast... My pr is doomed /s

library/common-library/templates/_secret.yaml.tpl
roobre: `common.licenseKey._actualKey` 😁

roobre: Jokes aside, I've used the first `licenseKey` in `common.licenseKey` as a sort of "namespace" for everything related to the licenseKey, which includes the licenseKey itself. But I'm open to rename it to something else, like `common.licensing`?

paologallinaharbur: I cannot come up with a better naming, I would like to avoid having in the same object twice "key" used with two different meanings. The issue that it is licenseKey already everywhere 😢

library/common-library/templates/_secret.yaml.tpl
```diff
@@ -0,0 +1,87 @@
...
+{{- end -}}
+
+{{/*
+Return the name of the secret holding the License Key.
+This helper is for internal use.
+*/}}
+{{- define "common.licenseKey._customSecretName" -}}
+{{- if .Values.customSecretName -}}
+  {{- .Values.customSecretName -}}
+{{- else if and .Values.global -}}
```
kang-makes, line 22:
```suggestion
{{- else if .Values.global -}}
```

library/common-library/templates/_secret.yaml.tpl
kang-makes: I don't bother if we have a `common.common.common.common` if it is readable like [this](https://github.com/newrelic/helm-charts/pull/749/files#diff-02a7e2152f1e56c1ccb8b4026261ab4d059454ac82f6f3b8cfcc099d0c294142R33) but I don't like that all the helper functions have a `common.nameOfTheTplFile.TheNameYouNeedForYourHelper` and this helper breaks this.

kang-makes (commented):
Impressive. This is why I wanted a common library. The usage of other PRs in this makes it really readable.

I left some comments I would like to address before approving it.

library/CHART-TEMPLATE/tests/common_library_license_helpers_test.yaml
```diff
@@ -0,0 +1,66 @@
+suite: test licenseKey helpers
+templates:
+  - templates/example-cm-licensekey.yaml
+release:
+  name: my-release
+  namespace: my-namespace
+tests:
+  - it: creates secret names for local licenseKey
+    set:
+      licenseKey: local
```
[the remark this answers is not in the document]
roobre: Got that, I did not realize `global` is not `null` in this chart. I'll add that tomorrow!

library/common-library/templates/_secret.yaml.tpl
kang-makes: Don't you say... Then put a couple of `and ` more and create a document of why we do not have enough `and ` in our code.

library/common-library/templates/_secret.yaml.tpl
kang-makes: I would rename this file to `license` and helpers to `common.license.licenseKey` and `common.license.licenseSecretKey` for example so we have the two `license` but not two `licenseKey`. WDYT?

library/common-library/templates/_secret.yaml.tpl
```diff
@@ -0,0 +1,87 @@
...
+    {{- .Values.global.customSecretName -}}
+  {{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return the name of the secret holding the License Key.
+*/}}
+{{- define "common.licenseKey.secretName" -}}
+{{ include "common.licenseKey._customSecretName" . | default (printf "%s-license" (include "common.naming.fullname" . )) }}
```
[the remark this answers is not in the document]
roobre: Not a huge fan of `printf` but I think it's good enough ❤️

library/common-library/templates/_secret.yaml.tpl
```diff
@@ -0,0 +1,87 @@
...
+kind: Secret
+metadata:
+  name: {{ include "common.licenseKey.secretName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "common.labels" . | nindent 4 }}
+data:
+  {{ include "common.licenseKey.secretKeyName" . }}: {{ include "common.licenseKey._licenseKey" . | b64enc }}
+{{- end }}
+{{- end }}
```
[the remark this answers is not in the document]
roobre: Files are split now, LMK what you think!

library/common-library/templates/_secret.yaml.tpl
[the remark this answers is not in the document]
roobre: Fair point, I've removed the feature for now :)

library/CHART-TEMPLATE/values.yaml
roobre: @kang-makes We can create an unit testing chart with sub-charts for these very corner cases. But I'd vouch for doing that in another PR :)

kang-makes: Yeah, in case we have time to do it is going to be done after GA Helm v3 charts...

library/common-library/templates/_license_secret.yaml.tpl
```diff
@@ -0,0 +1,22 @@
+{{/*
+Renders the license key secret if user has not specified a custom secret.
+*/}}
+{{- define "common.license.secret" }}
+{{- if not (include "common.license._customSecretName" .) }}
+
+{{/* Fail if licenseKey is empty and required */}}
```
kang-makes, line 7: As this `if` has no `-}}` the comment should be like this:
```mustache
{{- if not (include "common.license._customSecretName" .) }}

{{- /* Fail if licenseKey is empty and required */}}
```
Although I prefer this:

```suggestion
{{- if not (include "common.license._customSecretName" .) -}}

{{/* Fail if licenseKey is empty and required */}}
```

library/common-library/templates/_license_secret.yaml.tpl
```diff
@@ -0,0 +1,22 @@
...
+{{- if not (include "common.license._licenseKey" .) }}
+    {{- fail "You must specify a licenseKey or a customSecretName containing it" }}
+{{- end }}
```
kang-makes, line 10: Get rid of the empty lines that we have "for logic" and/or add the YAML start token (`---`)
```suggestion
{{- end }}
---
```

library/common-library/templates/_license_secret.yaml.tpl
kang-makes: This is a nit. It only renders empty lines.

library/common-library/templates/_license_secret.yaml.tpl
kang-makes: This is a nit. It only renders empty lines.

library/common-library/templates/_license.tpl
```diff
@@ -0,0 +1,55 @@
...
+
+{{/*
+Return the name key for the License Key inside the secret.
+This helper is for internal use.
+*/}}
+{{- define "common.license._customSecretKey" -}}
+{{- if .Values.customSecretLicenseKey -}}
+  {{- .Values.customSecretLicenseKey -}}
+{{- else if .Values.global -}}
+  {{- if .Values.global.customSecretLicenseKey }}
```
kang-makes, line 51: NIT because the next line has `{{-` and will get rid of that end-of-line.

```suggestion
  {{- if .Values.global.customSecretLicenseKey -}}
```

kang-makes (approved):
Things that I overlooked in the last review are only nits.

LGTM. Thanks!

library/common-library/templates/_license_secret.yaml.tpl
roobre: I think we will get an empty line here anyway as comments cannot have line chompers :(

library/common-library/templates/_license_secret.yaml.tpl
roobre: Ok, looks like it's safe to do: https://go.dev/play/p/rX5kp9Pp2bx

library/common-library/templates/_license_secret.yaml.tpl
kang-makes: [They can](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_low-data-mode.tpl#L10-L14)

library/common-library/templates/_license_secret.yaml.tpl
kang-makes: Some YAML parsers (not the golang one and not Helm) will have issues having empty lines because they detect it as a two document YAML with the first one empty.

roobre: Good point, I'll go for that.

library/common-library/templates/_license_secret.yaml.tpl
roobre: Last time I tried this I got a syntax error... TIL.

PR #750, opened 2022-03-16
[common-library] Add host network helper

kang-makes:
#### Is this a new chart

No

This adds to chart template and common library the ability to set `hostNetwork` to the pods.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/common-library/templates/_hostnetwork.tpl
+++ b/library/common-library/templates/_hostnetwork.tpl
@@ -1,0 +1,34 @@
+{{- /*
+Abstraction of the hostNetwork toggle.
+This helper allows to override the global `.global.hostNetwork` with the value of `.hostNetwork`.
+Returns "true" if `hostNetwork` is enabled, otherwise fallbacks to a overridable function
+*/ -}}
+{{- define "common.hostNetwork" -}}
+{{- /* This allows us to use `$global` as an empty dict directly in case `Values.global` does not exists */ -}}
+{{- $global := index .Values "global" | default dict -}}
+
+{{- /* `get` will return "" (empty string) if value is not found, and the value otherwise, so we can type-assert with kindIs */ -}}
+{{- if get .Values "hostNetwork" | kindIs "bool" -}}
+    {{- if .Values.hostNetwork -}}
+        true
+    {{- else -}}
+        false
+    {{- end -}}
+{{- else if get $global "hostNetwork" | kindIs "bool" -}}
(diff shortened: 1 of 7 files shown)


roobre:
I am still having some doubts on the use case for this, especially for the `global` one. While I can perfectly see why you would want to define labels, secContext, or image registries globally, I'm not seeing this use case that clearly.

library/common-library/templates/_hostnetwork.tpl
```diff
@@ -0,0 +1,34 @@
...
+{{- /* This allows us to use `$global` as an empty dict directly in case `Values.global` does not exists */ -}}
+{{- $global := index .Values "global" | default dict -}}
+
+{{- /* `get` will return "" (empty string) if value is not found, and the value otherwise, so we can type-assert with kindIs */ -}}
+{{- if get .Values "hostNetwork" | kindIs "bool" -}}
+    {{- if .Values.hostNetwork -}}
+        true
+    {{- else -}}
+        false
+    {{- end -}}
```
roobre, line 16: Why not just
```suggestion
    {{- .Values.hostNetwork }}
```

In the end, we have already asserted it is bool so it does not make much sense to if-else on it.

library/common-library/templates/_hostnetwork.tpl
```diff
@@ -0,0 +1,34 @@
...
+{{- else if get $global "hostNetwork" | kindIs "bool" -}}
+    {{- if $global.hostNetwork -}}
+        true
+    {{- else -}}
+        false
+    {{- end -}}
```
roobre, line 22: Ditto

roobre (approved):

library/common-library/templates/_hostnetwork.tpl
kang-makes: Solved in [#76799db](https://github.com/newrelic/helm-charts/pull/750/commits/76799db5ccad9eef227078fea726a82d6d12ea35)

library/common-library/templates/_hostnetwork.tpl
kang-makes: I left it when I did not assert that it is a bool. And I am still in doubt if I should fail when it is not a bool or not.

Solved in [#76799db](https://github.com/newrelic/helm-charts/pull/750/commits/76799db5ccad9eef227078fea726a82d6d12ea35)

PR #752, opened 2022-03-17
[common-library] Add verbose flag

kang-makes:
#### Is this a new chart

No

This allows toggling the debug verbosity consistently in all the Helm charts that use this library 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/CHART-TEMPLATE/templates/example-cm-verbose.yaml
+++ b/library/CHART-TEMPLATE/templates/example-cm-verbose.yaml
@@ -6,6 +6,6 @@ metadata:
 data:
   {{- /* Be careful because this returns an empty string when false because "false" is evaluated as true in Helm
          Take a look to the tests suite "common_library_lowdatamode_test.yaml" to see who this behaves */}}
-  verbose-with-if: {{ if include "common.verbose" . }}enabled{{ else }}disabled{{ end }}
-  verbose-with-quote: {{ include "common.verbose" . | quote }}
-  verbose-with-default: {{ include "common.verbose" . | default "false" | quote }}
+  verboseLog-with-if: {{ if include "common.verboseLog" . }}enabled{{ else }}disabled{{ end }}
+  verboseLog-with-quote: {{ include "common.verboseLog" . | quote }}
+  verboseLog-with-default: {{ include "common.verboseLog" . | default "false" | quote }}
--- a/library/CHART-TEMPLATE/templates/example-cm-verbose.yaml
+++ b/library/CHART-TEMPLATE/templates/example-cm-verbose.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "common.naming.fullname" . }}-examples-verbose
+  namespace: {{ .Release.Namespace }}
+data:
+  {{- /* Be careful because this returns an empty string when false because "false" is evaluated as true in Helm
+         Take a look to the tests suite "common_library_lowdatamode_test.yaml" to see who this behaves */}}
+  verbose-with-if: {{ if include "common.verbose" . }}enabled{{ else }}disabled{{ end }}
+  verbose-with-quote: {{ include "common.verbose" . | quote }}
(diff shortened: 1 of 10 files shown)


library/common-library/templates/_verbose.tpl
```diff
@@ -0,0 +1,26 @@
+{{- /*
+Abstraction of the verbose toggle.
+This helper allows to override the global `.global.verbose` with the value of `.verbose`.
+Returns "true" if `verbose` is enabled, otherwise "" (empty string)
+*/ -}}
+{{- define "common.verbose" -}}
+{{- /* `get` will return "" (empty string) if value is not found, and the value otherwise, so we can type-assert with kindIs */ -}}
+{{- if (get .Values "verbose" | kindIs "bool") -}}
+    {{- if .Values.verbose -}}
```
alvarocabanas, line 9: In newerelic-infrastructure | -v3 the value is called verboseLog, should we treat it here or have a compatibility layer helper in that repo?

kang-makes: You are right. Solved in [#3c553eb](https://github.com/newrelic/helm-charts/pull/752/commits/3c553ebce4127a9f81e6e4c39fd5344df8a584b4)

library/common-library/templates/_verbose-log.tpl
```diff
@@ -0,0 +1,26 @@
...
+{{- /* `get` will return "" (empty string) if value is not found, and the value otherwise, so we can type-assert with kindIs */ -}}
+{{- if (get .Values "verboseLog" | kindIs "bool") -}}
+    {{- if .Values.verboseLog -}}
+        {{- /*
+            We want only to return when this is true, returning `false` here will template "false" (string) when doing
+            an `(include "common.verboseLog" .)`, which is not an "empty string" so it is `true` if it is used
+            as an evaluation somewhere else.
+        */ -}}
+        {{- .Values.verboseLog -}}
+    {{- end -}}
```
roobre, line 16: In other PRs, we print the `false` string rather than returning empty. This does not allow to `if` the result of this helper, but does allow to render it unconditionally which is allegedly a more common use case.

kang-makes: We are using `with` in the templates so it is perfect to print `true` or `""`. In all the cases we have faced until now the default is `false`. See `hostNetwork` [here](https://github.com/newrelic/helm-charts/blob/master/library/CHART-TEMPLATE/tests/common_library_hostnetwork_test.yaml), `nrStaging` [here](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_staging.tpl) or this helper for adding a verbose flag. In case of them, you need an explicit `false` you can use `default` as the templates do [here](https://github.com/newrelic/helm-charts/pull/752/files#diff-c4d3c722fb93e10d6c397bc0513bbff9d34b0b4337f30130c41c49f6eea99566R11).

This comment has raised the issue that `hostNetwork` is printing "false" and that `staging` has no tests but other toggles we have has exactly the same behaviour (and I know becuase I copy&pasted them myself 😅)

alvarocabanas (approved):
LGTM

library/CHART-TEMPLATE/templates/example-cm-verbose.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ConfigMap
+metadata:
+  name: {{ include "common.naming.fullname" . }}-examples-verbose
+  namespace: {{ .Release.Namespace }}
+data:
+  {{- /* Be careful because this returns an empty string when false because "false" is evaluated as true in Helm
+         Take a look to the tests suite "common_library_lowdatamode_test.yaml" to see who this behaves */}}
+  verboseLog-with-if: {{ if include "common.verboseLog" . }}enabled{{ else }}disabled{{ end }}
+  verboseLog-with-quote: {{ include "common.verboseLog" . | quote }}
+  verboseLog-with-default: {{ include "common.verboseLog" . | default "false" | quote }}
```
roobre, line 11: Should we have a `common.verboseLog.value` as we do with `hostNetwork`?

roobre (approved):

PR #753, opened 2022-03-17
[common-library] Add dnsConfig feature to the common library

### LINKED ISSUE #589: [nri-infrastructure] Reduce DNS load on the K8 clusters

On my EKS cluster I see newrelic to be the n#1 DNS request issuer. By default when doing a DNS query on "infra-api.newrelic.com", coreDNS will try to resolve it in different context :
* infra-api.newrelic.com.eu-west-1.compute.internal.
* infra-api.newrelic.com.cluster.local
* infra-api.newrelic.com.svc.cluster.local.
* infra-api.newrelic.com.newrelic.svc.cluster.local.

Finally it's able to understand that the pod is not asking for a local address but for api.newrelic.com public one. So each time a pod is doing a DNS query to newrelic, it goes through 5 errors in the DNS resolution before finding the good one.

Since newrelic is using multiple endpoints (intra, indentity, metric), multiplied by the number of hosts and the frequency of theses requests, you end up with hundread thousands of DNS request per day for nothing. 

One solution could be to use only fully qualified domain name (FQDN) like **infra-api.newrelic.com.** (mind the tailing dot), or add an option to override the default newrelic endpoint.

Another option would be to allow us setting the NDOT setting of the deamonset :
``` yaml 

dnsConfig:
    options:
      - name: ndots
        value: '1'
```

> ndots sets a threshold for the number of dots which must appear in a name before an initial absolute query will be made. The default for n is 1, meaning that if there are any dots in a name, the name will be tried first as an absolute name before any search list elements are appended to it.

**Additional context**

A good article about FQDN kube and DNS : https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html

kang-makes:
#### Is this a new chart
No

We still have to adapt all the helm charts to use this but this allows us to change `dnsConfig` through all our helm charts.

  - Fixes #589 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/CHART-TEMPLATE/Chart.lock
+++ b/library/CHART-TEMPLATE/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common-library
   repository: file://../common-library
-  version: 0.9.0
-digest: sha256:a4d77b4baea7aee359a160157aa2a0187285827a03fde1b6da4168ae61f70989
-generated: "2022-03-15T16:01:23.39701+01:00"
+  version: 0.10.0
+digest: sha256:81087781c9d1d118efc31743f7d7b4cbeb790dbb1d0d3a2a8144ae0e427af4c8
+generated: "2022-03-17T16:32:44.730673+01:00"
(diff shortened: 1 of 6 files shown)


alvarocabanas (approved):
LGTM

PR #754, opened 2022-03-18
common-library: implement privileged helper

roobre:
This PR implements the `common.privileged` helper.

As stated in the comments, all the logic of handling how this flag impacts settings such as `hostNetwork` or the pod and container `securityContext`s is deliberately left out from the common library, and instead Chart writers are expected to use `common.privileged` to define their default overrides for the mentioned settings.

I have two reasons to believe this is a better option than trying to implement this in the library:
1. We only have one chart (two, if you count logging, but its special so refer to point 2) that uses the privileged mode. I believe we should only implement common things in the common library and logic used by one chart is not common.
2. Charts that require the privileged flag require it because they do weird things. It is likely that different charts that use `common.privileged` understand the meaning of "privileged" in different ways, and therefore will require specific behavior to handle it. For this reason, I believe it's just simpler to offload this to the chart rather to try to find some common-yet-flexible approach.

### DIFF
--- a/library/common-library/templates/_privileged.tpl
+++ b/library/common-library/templates/_privileged.tpl
@@ -18,17 +18,18 @@ And then use the helpers this library provides to render those.
 {{- $global := index .Values "global" | default dict -}}
 {{- /* `get` will return "" (empty string) if value is not found, and the value otherwise, so we can type-assert with kindIs */ -}}
 {{- if get .Values "privileged" | kindIs "bool" -}}
-    {{- with .Values.privileged -}}
-        {{- . -}}
+    {{- if .Values.privileged -}}
+        {{- .Values.privileged -}}
     {{- end -}}
 {{- else if get $global "privileged" | kindIs "bool" -}}
-    {{- with $global.privileged -}}
-        {{- . -}}
+    {{- if $global.privileged -}}
+        {{- $global.privileged -}}
     {{- end -}}
 {{- else -}}
     {{- include "common.privileged.defaultOverride" . -}}
 {{- end -}}
 {{- end -}}
+
 
 {{- /*
 This allows to change the default user setting for `privileged`, by default it returns a falsy value ("").
--- a/library/common-library/templates/_privileged.tpl
+++ b/library/common-library/templates/_privileged.tpl
@@ -1,0 +1,37 @@
+{{- /*
+This file contains helpers to handle whether the chart should assume it is fine to deploy itself with elevated
+privileges, thus changing the defaults for hostNetwork, securityContexts, etc.
+Please note that either local or global definitions of the aforementioned settings will override the privileged flag.
+*/ -}}
+
+{{- /*
+common.privileged is a helper that returns whether the chart should assume the user is fine deploying privileged pods.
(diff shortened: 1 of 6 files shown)


library/common-library/templates/_privileged.tpl
```diff
@@ -0,0 +1,37 @@
...
+*/ -}}
+{{- /* Chart writers should not rely on this helper directly. */ -}}
+{{- define "common.privileged" -}}
+{{- /* This allows us to use `$global` as an empty dict directly in case `Values.global` does not exists. */ -}}
+{{- $global := index .Values "global" | default dict -}}
+{{- /* `get` will return "" (empty string) if value is not found, and the value otherwise, so we can type-assert with kindIs */ -}}
+{{- if get .Values "privileged" | kindIs "bool" -}}
+    {{- with .Values.privileged -}}
+        {{- . -}}
+    {{- end -}}
```
kang-makes, line 23: I think the `with` is cool while templating but it could somehow hide the things that it is doing.

This is the reason I chose in other PRs to use `if-else`. as we used in other helpers:
 - https://github.com/newrelic/helm-charts/blob/privileged/library/common-library/templates/_low-data-mode.tpl
 - https://github.com/newrelic/helm-charts/blob/privileged/library/common-library/templates/_staging.tpl
 - https://github.com/newrelic/helm-charts/blob/privileged/library/common-library/templates/_serviceaccount.tpl#L1-L38
 
There is no style guide yet but I would prefer to be the same as the other approved and merged. If the team prefers to switch to `with` instead of `if` I would change all at the same time after merging this properly to see al the changes of style at the same time.

```suggestion
    {{- if .Values.privileged -}}
        {{- .Values.privileged -}}
    {{- end -}}
```

library/common-library/templates/_privileged.tpl
```diff
@@ -0,0 +1,37 @@
...
+{{- else if get $global "privileged" | kindIs "bool" -}}
+    {{- with $global.privileged -}}
+        {{- . -}}
+    {{- end -}}
+{{- else -}}
+    {{- include "common.privileged.defaultOverride" . -}}
+{{- end -}}
+{{- end -}}
+
+{{- /*
```
kang-makes, line 33: Two line jumps between functions to see better the switch of function on large helpers. Seen [here](https://github.com/newrelic/helm-charts/blob/privileged/library/common-library/templates/_serviceaccount.tpl#L1-L38) (if you ignore that a one of the functions has three jumps 😅): 
```suggestion
{{- end -}}


{{- /*
```

kang-makes (approved):
For more context for the team when they come here to this PR (or if you are searching on a `git blame` or something):

@roobre and I were talking about whether this PR should also implement changing defaults like `hostNetwork` changes its defaults if it is in `privileged mode` or not.

The changes that a chart needs while in `privileged` are intrinsic to the chart, the technology and such. For `nri-kubernetes` it means having write access to the host and `hostNetwork` for the infra pod while it does not care for the rest of the pods it has to create or some chart simply drops a capability or changes the user that container runs as and can vary depending on which image each chart uses.

So the simplest approach is also de more flexible while it keeps the tooling consistent.

LGTM.

PR #756, opened 2022-03-21
[common-library] Fix hostNetwork default behavior

kang-makes:
#### Is this a new chart

No

`hostNetwork` was not correctly developed and printed `false` when being called. A new test, template and function have been added to solve this issue.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/CHART-TEMPLATE/templates/deployment.yaml
+++ b/library/CHART-TEMPLATE/templates/deployment.yaml
@@ -23,10 +23,6 @@ spec:
       imagePullSecrets:
         {{- . | nindent 8 }}
       {{- end }}
-      {{- with include "common.hostNetwork" . }}
-      hostNetwork: {{ . }}
-      {{- end }}
-      serviceAccountName: {{ include "common.serviceAccount.name" . }}
       {{- with include "common.securityContext.pod" . }}
       securityContext:
         {{- . | nindent 8 }}
@@ -38,6 +34,10 @@ spec:
       dnsConfig:
         {{- . | nindent 8 }}
       {{- end }}
+
+      hostNetwork: {{ include "common.hostNetwork.boolean" . }}
+      serviceAccountName: {{ include "common.serviceAccount.name" . }}
+
       containers:
         - name: {{ .Chart.Name }}
           {{- with include "common.securityContext.container" . }}
--- a/library/CHART-TEMPLATE/templates/deployment.yaml
+++ b/library/CHART-TEMPLATE/templates/deployment.yaml
@@ -35,7 +35,7 @@ spec:
         {{- . | nindent 8 }}
       {{- end }}
 
-      hostNetwork: {{ include "common.hostNetwork.boolean" . }}
+      hostNetwork: {{ include "common.hostNetwork.value" . }}
       serviceAccountName: {{ include "common.serviceAccount.name" . }}
 
       containers:
--- a/library/common-library/templates/_hostnetwork.tpl
+++ b/library/common-library/templates/_hostnetwork.tpl
@@ -7,11 +7,21 @@ Returns "true" if `hostNetwork` is enabled, otherwise fallbacks to a overridable
 {{- /* This allows us to use `$global` as an empty dict directly in case `Values.global` does not exists */ -}}
 {{- $global := index .Values "global" | default dict -}}
 
-{{- /* `get` will return "" (empty string) if value is not found, and the value otherwise, so we can type-assert with kindIs */ -}}
+{{- /*
+`get` will return "" (empty string) if value is not found, and the value otherwise, so we can type-assert with kindIs
+
+We also want only to return when this is true, returning `false` here will template "false" (string) when doing
+an `(include "common.hostNetwork" .)`, which is not an "empty string" so it is `true` if it is used
+as an evaluation somewhere else.
+*/ -}}
 {{- if get .Values "hostNetwork" | kindIs "bool" -}}
-    {{- .Values.hostNetwork -}}
+    {{- if .Values.hostNetwork -}}
+        {{- .Values.hostNetwork -}}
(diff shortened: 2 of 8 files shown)


library/CHART-TEMPLATE/templates/deployment.yaml
```diff
@@ -38,6 +34,10 @@ spec:
       dnsConfig:
         {{- . | nindent 8 }}
       {{- end }}
+
+      hostNetwork: {{ include "common.hostNetwork.boolean" . }}
```
roobre, line 38: I like the name to resolve the ambiguity!

As a nit, I'd rather use `bool` as it's the Go type name.

library/common-library/templates/_hostnetwork.tpl
```diff
@@ -7,20 +7,42 @@ Returns "true" if `hostNetwork` is enabled, otherwise fallbacks to a overridable
...
 {{- end -}}
 {{- end -}}
 
 
+{{- /*
+Abstraction of the hostNetwork toggle.
+This helper allows to override the global `.global.hostNetwork` with the value of `.hostNetwork`.
+Returns "true" if `hostNetwork` is enabled, otherwise fallbacks to a overridable function
+*/ -}}
+{{- define "common.hostNetwork.boolean" -}}
```
roobre, line 36: ditto
```suggestion
{{- define "common.hostNetwork.bool" -}}
```

or maybe

```suggestion
{{- define "common.hostNetwork.value" -}}
```

kang-makes: I like `value`. Going for it.

library/CHART-TEMPLATE/templates/deployment.yaml
kang-makes: Solved in [#cd27a22](https://github.com/newrelic/helm-charts/pull/756/commits/cd27a2262b8b953a7e5f921de89f8c81f3ead7c4)

library/common-library/templates/_hostnetwork.tpl
kang-makes: Solved in [#cd27a22](https://github.com/newrelic/helm-charts/pull/756/commits/cd27a2262b8b953a7e5f921de89f8c81f3ead7c4)

library/common-library/templates/_hostnetwork.tpl
```diff
@@ -7,20 +7,41 @@ Returns "true" if `hostNetwork` is enabled, otherwise fallbacks to a overridable
...
+        {{- $global.hostNetwork -}}
+    {{- end -}}
 {{- else -}}
     {{- include "common.hostNetwork.defaultOverride" . -}}
 {{- end -}}
 {{- end -}}
 
 
+{{- /*
+Abstraction of the hostNetwork toggle.
```
paologallinaharbur, line 32: I do not understand why this is useful.. To get an explicit false instead of ""?

library/common-library/templates/_hostnetwork.tpl
```diff
@@ -7,20 +7,41 @@ Returns "true" if `hostNetwork` is enabled, otherwise fallbacks to a overridable
...
+    {{- if .Values.hostNetwork -}}
+        {{- .Values.hostNetwork -}}
+    {{- end -}}
 {{- else if get $global "hostNetwork" | kindIs "bool" -}}
-    {{- $global.hostNetwork -}}
+    {{- if $global.hostNetwork -}}
+        {{- $global.hostNetwork -}}
+    {{- end -}}
 {{- else -}}
     {{- include "common.hostNetwork.defaultOverride" . -}}
```
paologallinaharbur, line 26: why defaultOverride? isn't take into consideration merely if nothing else is set?

library/common-library/templates/_hostnetwork.tpl
kang-makes: Because if you use something like `{{ if include "common.hostNetwork" . }}` and it returns `"false"` (note it is a string) helm evaluates it as `true`. Because Helm falsiness does not include `"false"` but an empty string.

kang-makes: Should I add [this comment](https://github.com/newrelic/helm-charts/blob/fix/common-library/hostnetwork/library/common-library/templates/_low-data-mode.tpl#L10-L14) that is in `lowDataMode` and `staging` also here?

library/common-library/templates/_hostnetwork.tpl
kang-makes: The name in the `common-library` itself could be misleading but when you use the library as it is called "defaultOverride" it makes sense.

library/common-library/templates/_hostnetwork.tpl
paologallinaharbur: it it fine!

paologallinaharbur (approved):

roobre (approved):

library/common-library/templates/_hostnetwork.tpl
```diff
@@ -7,20 +7,41 @@ Returns "true" if `hostNetwork` is enabled, otherwise fallbacks to a overridable
...
+{{- /*
+Abstraction of the hostNetwork toggle.
+This helper abstracts the function "common.hostNetwork" to return true or false directly.
+*/ -}}
+{{- define "common.hostNetwork.value" -}}
+{{- if include "common.hostNetwork" . -}}
+true
+{{- else -}}
+false
+{{- end -}}
```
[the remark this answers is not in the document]
kang-makes: I thought about it and thought about fi this helpers makes sense and we can user `default` while templating and I finally decided that I would prefer to have `print "true"` like to clarify this use case.

This is sugar to make templates easier so I preferred to left it explicit.

PR #757, opened 2022-03-21
[common-library] Add agent configuration

kang-makes:
#### Is this a new chart

No

The approach of having agent configuration managed globally was too scary so we reduced it to a helper that has all the configurations managed by other helpers (like `staging` or `verbose`) and left it ready to be used in other agent configurations.

#### Checklist
- [x] Chart Version bumped


library/CHART-TEMPLATE/values.yaml
```diff
@@ -15,6 +15,15 @@ global:
   nodeSelector: {}
   tolerations: []
   affinity: {}
+  agent:  # This should not touch main/master because it has things that could break all the charts
+    config:
+      global: global
+      global_overriden: test
```
roobre, line 21: As a user, I would expect things that I put in `global` to be applied to all charts (unless overridden). This makes me have two concerns with having `global.agent`:
- Not all charts will have an agent, which might be counterintuitive
- Some charts might have internal options (e.g. env vars) that silently override things I can put here. As a user I don't have visibility on that and might create unexpected behavior.

I wouldn't consider any of this points to be a blocker, but I wonder what are the advantages of having a `global.agent` to offset the downsides.

roobre, line 21: As a user, I would expect things that I put in `global` to be applied to all charts (unless overridden). This makes me have two concerns with having `global.agent`:
- Not all charts will have an agent, which might be counterintuitive
- Some charts might have internal options (e.g. env vars) that silently override things I can put here. As a user I don't have visibility on that and might create unexpected behavior.

I wouldn't consider any of this points to be a blocker, but I wonder what are the advantages of having a `global.agent` to offset the downsides.

library/common-library/templates/_agent-config.tpl
```diff
@@ -0,0 +1,88 @@
...
+{{- define "common.agentConfig.defaults" -}}
+{{- /*if include "common.verboseLog" . }}
+verbose: 1
+{{- end */}}
+{{- if (include "common.nrStaging" .) }}
+staging: true
+{{- end }}
+{{- with include "common.proxy" . }}
+proxy: {{ . | quote }}
+{{- end }}
```
roobre, line 13: This is a very useful thing. I like it.

roobre, line 13: This is a very useful thing. I like it.

library/common-library/templates/_agent-config.tpl
```diff
@@ -0,0 +1,88 @@
...
+features: "docker_enabled:false"
+
+httpServerEnabled: true
+httpServerPort: 8003
+
+passthroughEnvironment: CLUSTER_NAME
+
+{{- if .Values.enableProcessMetrics | kindIs "bool" }}
+EnableProcessMetrics: {{ .Values.enableProcessMetrics | quote }}
+{{- end }}
```
roobre, line 40: If this is merely an example I would leave it out of the PR. At this point I find hard to tell if this is required for the architecture to work or not.

On top of this: Does `common.agentConfig.defaults.override` completely replace (i.e. is _not_ merge with) `common.agentConfig.defaults`? I think the latter has very nice things (like the proxy stuff) that would be a shame to have to copypaste.


What is the use case for `common.agentConfig.defaults.override`? Perhaps we could make it `common.agentConfig.extra`, and merge it with the library defaults?

roobre, line 40: If this is merely an example I would leave it out of the PR. At this point I find hard to tell if this is required for the architecture to work or not.

library/common-library/templates/_agent-config.tpl
```diff
@@ -0,0 +1,88 @@
...
+THIS IS AN EXAMPLE OF HOW THE CONFIG DEFAULTS SHOULD GO.
+This should not touch main/master because it has things that could break all the charts
+that use this library.
+nri-kubernetes has been taken as an example with the variables that it has hascoded.
+***************************************************************************************
+********************************************************************************** */}}
+overrideHostRoot: ""
+forwardOnly: true
+isSecureForwardOnly: true
+{{- end -}}
```
roobre, line 62: I am again not quite seeing why would we want to do this. If we need to override the helper logic entirely, what I'd do is just not use the helper and render a configMap by hand.

roobre, line 62: I am again not quite seeing why would we want to do this. If we need to override the helper logic entirely, what I'd do is just not use the helper and render a configMap by hand.

library/CHART-TEMPLATE/templates/sidecar-agent-configmap.yaml
```diff
@@ -0,0 +1,6 @@
+{{- include "common.agentConfig.configMap" (dict
+    "global"  .Values.global.agent.config
+    "local"   .Values.agent.config
+    "name"    (printf "%s-example-config" (include "common.naming.fullname" . ))
```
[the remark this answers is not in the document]
roobre: Ok, I see now: It's because we might want to have different agent `configMap`s. My question would be: Does it make sense then? All the other helpers are built around centralizing common things, while this one needs to do advanced logic to be able to be too flexible. Couldn't that fact be telling us that the agent config is not as `common` as we thought?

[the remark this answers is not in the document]
roobre: Ok, I see now: It's because we might want to have different agent `configMap`s. My question would be: Does it make sense then? All the other helpers are built around centralizing common things, while this one needs to do advanced logic to be able to be too flexible. Couldn't that fact be telling us that the agent config is not as `common` as we thought?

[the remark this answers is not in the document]
kang-makes: Images [has also this complex behavior](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_images.tpl#L17-L30) that [leads to complex usage](https://github.com/newrelic/helm-charts/blob/master/library/CHART-TEMPLATE/templates/deployment.yaml#L22-L25).

When there is "more than one" either image or agent it leads to this pattern because Helm is not flexible enough.

Although it could look complex, it uses the same pattern as Bitnami.

library/CHART-TEMPLATE/values.yaml
kang-makes: > Not all charts will have an agent, which might be counterintuitive

You are right. But as other helpers, they are here to help those charts that do have them. The same as `privileged` is effectively used by one chart but it is in the common-library to use the power of all the helpers(overrides we developed.

> Some charts might have internal options (e.g. env vars) that silently override things I can put here. As a user I don't have visibility on that and might create unexpected behavior.

This aims to add flexibility when we failed to add it to the charts so users don't get constraints because of our possible lack of vision of their use cases. This is a POC and if the merge order should be changed we could do it now if you wish. IIRC I added user's values on top of the defaults (and added a override as an example). We (and the user) can footgun ourselves in many ways like using `verboseLog` at the same time that we change the verbosity in the config. We should document that this is an advance feature, of course.

kang-makes: Duplicate of [this](https://github.com/newrelic/helm-charts/pull/757#discussion_r832149573). Resolving.

library/common-library/templates/_agent-config.tpl
kang-makes: Maybe this adds noise and you are right. I left it as an example of the way `nri-kubernetes` should use these helpers in the same way I used them [here](https://github.com/newrelic/helm-charts/pull/757#discussion_r832149970)

The use case of `common.agentConfig.defaults.override` is to put things that has to be done in that way always. Like `forwardOnly` for example. Is a future proof template that could allow us to use the image from CAOS without having to create new images just tweaking helpers.

It is an example (and it opened the conversation I wanted to open) and I will remove it before putting this PR as "ready to review".

kang-makes: Duplicated of [this](https://github.com/newrelic/helm-charts/pull/757#discussion_r832150858). Resolving.

library/common-library/templates/_agent-config.tpl
kang-makes: I answered this [here](https://github.com/newrelic/helm-charts/pull/757#discussion_r832176827) without noticing. We could have the conversation in the same thread if you would.

kang-makes: Duplicated of [this](https://github.com/newrelic/helm-charts/pull/757#discussion_r832153829). Resolving.

library/common-library/templates/_agent-config.tpl
```diff
@@ -0,0 +1,88 @@
...
+{{/* **********************************************************************************
+***************************************************************************************
+THIS IS AN EXAMPLE OF HOW THE CONFIG DEFAULTS SHOULD GO.
+This should not touch main/master because it has things that could break all the charts
+that use this library.
+nri-kubernetes has been taken as an example with the variables that it has hascoded.
+***************************************************************************************
+********************************************************************************** */}}
+overrideHostRoot: ""
+forwardOnly: true
```
paologallinaharbur, line 60: I find this a bit scary, there is a lot of logic already involving agent config. 

Aren't enough the helpers for the agent config (proxy, fedramp, verbose, etc) rather then having global config?  Which config are we trying to cover?

Most of agent deployed are used in different ways needing almost no parameters that are not specific to each installation this is why I wonder if it worth it.

kang-makes: I answered it [here](https://github.com/newrelic/helm-charts/pull/757#discussion_r832176827) and [here](https://github.com/newrelic/helm-charts/pull/757#discussion_r832171075)

> Is a future proof template that could allow us to use the image from CAOS without having to create new images just tweaking helpers.

> This aims to add flexibility when we failed to add it to the charts so users don't get constraints because of our possible lack of vision of their use cases. [...] We (and the user) can footgun ourselves in many ways like using `verboseLog` at the same time that we change `verbose` in the config. We should document that this is an advance feature, of course.

paologallinaharbur: Replacing different images with configuration can make sense. However, wouldn't be a bit obscure for GTS and users having all this logic in helpers?

The point is that I have the feeling that none of the options of the example (and also of the agent) needs to be setted across all charts at the same time with the same value:
```
  features: docker_enabled:false
  forwardOnly: true
  global: global
  global_overriden: test
  httpServerEnabled: true
  httpServerPort: 8003
  isSecureForwardOnly: true
  local: test
  overrideHostRoot: ""
  passthroughEnvironment: CLUSTER_NAME
```

 - Therefore I see this option useful to set the defaults of the agent config in the very same way across multiple charts.

Regarding checking if the config has Verbose or any other config that should be set in a different way we can warn about it

kang-makes: You are right. I wrote this to @roobre [here](https://github.com/newrelic/helm-charts/pull/757#discussion_r832176827):

> Maybe this adds noise and you are right. I left it as an example of the way `nri-kubernetes` should use these helpers [...]

I have the feeling that GTS would be more comfortable reading a configMap with the configuration and the `values.yaml` with the agent's configuration than the way we have it right now.

Would you prefer that I create a more finished/polished PR instead of this one and discuss over it?

paologallinaharbur: I am still not fully convinced that the use-case is strong enough to add and then maintain all this logic, but maybe I am missing some piece. @roobre what do you think?

library/common-library/templates/_agent-config.tpl
```diff
@@ -0,0 +1,17 @@
...
+{{- if include "common.verboseLog" . }}
+verbose: 1
+{{- end }}
+{{- if (include "common.nrStaging" . ) }}
+staging: true
+{{- end }}
+{{- with include "common.proxy" . }}
+proxy: {{ . | quote }}
+{{- end }}
+{{- with include "common.fedramp.enabled" . }}
```
paologallinaharbur, line 14: the fedramp enable I believe returns just true/false

we should populate the correct endpoint for the agent

library/common-library/templates/_agent-config.tpl
```diff
@@ -0,0 +1,17 @@
+{{/*
+This helper should return the defaults that all agents should have
+*/}}
+{{- define "common.agentConfig.defaults" -}}
+{{- if include "common.verboseLog" . }}
+verbose: 1
+{{- end }}
+{{- if (include "common.nrStaging" . ) }}
+staging: true
+{{- end }}
```
paologallinaharbur, line 10: could we add as well NRIA_CUSTOM_ATTRIBUTES?

kang-makes: I am in if we want custom attributes as a `global` to be honest...

library/common-library/templates/_agent-config.tpl
kang-makes: I only found [this about fedramp](https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings/#fedramp).

Which other variables do we need?

paologallinaharbur: oh, nothing, last time I checked I believe we needed to add manually each endpoint. Either I am wrong or they added it later on. Then it should be ok

library/CHART-TEMPLATE/tests/common_cluster_test.yaml
```diff
@@ -1,6 +1,7 @@
 suite: test cluster helper
 templates:
-  - templates/deployment.yaml
+  - templates/integration-deployment.yaml
+  - templates/sidecar-deployment.yaml
```
paologallinaharbur, line 4: why do we need to add a sidecar template and to test it? to show how to use it?

kang-makes: Because tests can change because of things like `.spec.template.spec.containers[?].*`

Also the thought that we had about having a style guide on how to write Helm charts applies in the Chart template.

library/common-library/templates/_agent-config.tpl
roobre: I think we will have the fattest global I've ever seen in a helm chart 😅 
But let's go for it, at this point we're already in.

kang-makes: This conversation will be solved with #767

roobre (approved):
LGTM, I'd only change the title and get rid of "global" from the title :)

library/common-library/templates/_agent-config.tpl
```diff
@@ -5,13 +5,26 @@ This helper should return the defaults that all agents should have
...
+
 {{- if (include "common.nrStaging" . ) }}
 staging: true
 {{- end }}
+
+{{- with include "common.license._licenseKey" . }}
+{{- /* We will mostly override this using a environment variable but we can have it here just in case */}}
+license: {{ . }}
+{{- end }}
+
```
roobre, line 17: Not a big fan of doing things in two different ways. Plus, this will cause the licenseKey to almost surely end up in a ConfigMap rather than a Secret. I'd stick with the env+secret approach for the LK.

library/common-library/templates/_agent-config.tpl
```diff
@@ -5,13 +5,26 @@ This helper should return the defaults that all agents should have
...
 {{- end }}
+
 {{- with include "common.fedramp.enabled" . }}
 fedramp: {{ . }}
 {{- end }}
+
+{{- with include "common.customAttributes" . }}
+custom_attributes:
+  {{- . | nindent 2 }}
+{{- end }}
```
roobre, line 29: Are customAttributes allowed to be formatted as YAML? I though they need to be a JSON-encoded string.

roobre (changes requested):

library/common-library/templates/_agent-config.tpl
```diff
@@ -5,13 +5,26 @@ This helper should return the defaults that all agents should have
...
 verbose: 1
 {{- end }}
+
 {{- if (include "common.nrStaging" . ) }}
 staging: true
 {{- end }}
+
+{{- with include "common.license._licenseKey" . }}
+{{- /* We will mostly override this using a environment variable but we can have it here just in case */}}
+license: {{ . }}
```
paologallinaharbur, line 15: agent configuration is usually added as a config-map. I would avoid having the license key here.

roobre: Jinx!

kang-makes: LGTM. Removing.

library/common-library/templates/_agent-config.tpl
kang-makes: LGTM. Removing.

library/common-library/templates/_agent-config.tpl
kang-makes: Removed in [#f629333](https://github.com/newrelic/helm-charts/pull/757/commits/f6293332c3ba99e6898c337a9b5ee93c24f67c12)

library/common-library/templates/_agent-config.tpl
kang-makes: Removed in [#f629333](https://github.com/newrelic/helm-charts/pull/757/commits/f6293332c3ba99e6898c337a9b5ee93c24f67c12)

library/CHART-TEMPLATE/templates/sidecar-agent-configmap.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "common.naming.fullname" . }}-agent-config
+data:
+  newrelic-infra.yml: |
+    {{- $agentDefaults := fromYaml (include "common.agentConfig.defaults" . ) }}
+    {{- $agentSettings := .Values.agent | default dict }}
+    {{- mustMergeOverwrite $agentDefaults $agentSettings | toYaml | nindent 4 }}
```
paologallinaharbur, line 10: do we have a test covering the order of priority of each config?

library/common-library/templates/_agent-config.tpl
```diff
@@ -0,0 +1,30 @@
...
+
+{{- with include "common.fedramp.enabled" . }}
+fedramp: {{ . }}
+{{- end }}
+
+{{- with include "common.customAttributes" . }}
+custom_attributes:
+  {{- . | nindent 2 }}
+{{- end }}
+{{- end -}}
```
paologallinaharbur, line 25: do we have a test covering the "happy path" of the config generation?

library/common-library/templates/_agent-config.tpl
kang-makes: It is needed as a string if you use it as `NRIA_CUSTOM_ATTRIBUTES` environment variable but you can put a map directly in the YAML: https://github.com/newrelic/infrastructure-agent/blob/master/pkg/config/config.go#L187

roobre: It seems like they can be YAML: https://github.com/newrelic/infrastructure-agent/blob/master/pkg/config/config.go#L182=

roobre (approved):
Unblocking, but I'd follow @paologallinaharbur's advice and add a simple tests for merging order.

library/common-library/templates/_agent-config.tpl
kang-makes: For the custom attributes we have it [here](https://github.com/newrelic/helm-charts/blob/common-library/add-agent-global/library/CHART-TEMPLATE/tests/common_library_custom_attributes_test.yaml)

But we don't have it for the config file as you stated above.

library/CHART-TEMPLATE/templates/sidecar-agent-configmap.yaml
kang-makes: I completely forgot during this strange week. Sorry.
 
 I'm on it.

kang-makes: There are two new commits: the first one with the tests and the second one fixing issues raised by the tests.

Everything should be solved with [#3bad27b](https://github.com/newrelic/helm-charts/pull/757/commits/3bad27b464c2a47b003ab924bc3b100c74ea30e3)

library/CHART-TEMPLATE/tests/common_library_agent_config_test.yaml
```diff
@@ -0,0 +1,204 @@
...
+  - it: sets custom_attributes according to values
+    set:
+      customAttributes:
+        cluster: name
+        test: test
+    asserts:
+      - equal:
+          path: data.newrelic-infra\.yml
+          value: |
+            custom_attributes:
```
paologallinaharbur, line 152: it is incredible that this was possible and we were not doing so

kang-makes: ¯\_(ツ)_/¯

paologallinaharbur (approved):

PR #758, opened 2022-03-22
[common-library] Fix template naming

kang-makes:
#### Is this a new chart

No

The function that overrides the chart name was implemented but not used in `common.naming.fullname` function. 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/common-library/templates/_naming.tpl
+++ b/library/common-library/templates/_naming.tpl
@@ -26,7 +26,7 @@ If release name contains chart name it will be used as a full name.
 {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
 {{- else }}
 
-{{- $name := default (include "common.naming.chartnameOverride" .) .Values.nameOverride }}
+{{- $name := .Values.nameOverride | default (include "common.naming.chartnameOverride" .) }}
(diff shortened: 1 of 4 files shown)


library/common-library/templates/_naming.tpl
```diff
@@ -26,7 +26,7 @@ If release name contains chart name it will be used as a full name.
 {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
 {{- else }}
 
-{{- $name := default .Chart.Name .Values.nameOverride }}
+{{- $name := default (include "common.naming.chartnameOverride" .) .Values.nameOverride }}
```
paologallinaharbur, line 29: Could we change this default to use the `|` so that it is obvious the order?

paologallinaharbur (approved):
just a nit

library/common-library/templates/_naming.tpl
kang-makes: Solved in [#5cd2662](https://github.com/newrelic/helm-charts/pull/758/commits/5cd26625c7c25f6ff377376eb450dbfbdb6bff6f)

PR #759, opened 2022-03-22
[common-library] Fail when no cluster name is given by the user

kang-makes:
#### Is this a new chart

No

`cluster` helper gets cluster correctly but does not fail when it is not defined. Charts using this library should rely on helpers who will always give the value they need. 

If you feel a bit overwhelmed by the number of changes go commit by commit as I also sanitized the tests in this PR removing useless things that they were not working.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/CHART-TEMPLATE/tests/common_cluster_test.yaml
+++ b/library/CHART-TEMPLATE/tests/common_cluster_test.yaml
@@ -11,14 +11,34 @@ release:
 #constantly to overwrite it.
 
 tests:
-  - it: helper works (and fails) when global is null and no local cluster value
+  - it: helper works when global is null
     set:
       global: null
+      cluster: test-cluster
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[0].env[0].value
+          value: test-cluster
+
+  - it: helper works when local is null
+    set:
+      global:
+        cluster: test-cluster
       cluster: null
     asserts:
       - equal:
           path: spec.template.spec.containers[0].env[0].value
-          value: null
+          value: test-cluster
+
+#I found no way of asserting failing configurations. This test fails with:
+# Error: execution error at (CHART-TEMPLATE/templates/deployment.yaml:71:24): There is not cluster name definition set neither in `.global.cluster' nor `.cluster' in your values.yaml. Cluster name is required.
+#  - it: helper works (and fails) when cluster is null everywhere
+#    set:
+#      global: null
+#      cluster: test
+#    asserts:
+#      - hasDocuments:
+#          count: 0
 
   - it: sets local cluster as cluster name even if global is set
     set:
--- a/library/CHART-TEMPLATE/tests/common_cluster_test.yaml
+++ b/library/CHART-TEMPLATE/tests/common_cluster_test.yaml
@@ -4,20 +4,22 @@ templates:
 release:
   name: my-release
   namespace: my-namespace
-set:
+
+# The name of the cluster is mandatory, to not make all the tests fail because they have no license
+# and not having to add license to all the tests I have added it to the `values.yaml`.
+# This change makes this tests a bit funky because we have to manually set local cluster to null
+#constantly to overwrite it.
+
 tests:
-  - it: sets null as cluster name if not specified
-    asserts:
-      - equal:
-          path: spec.template.spec.containers[0].env[0].value
-          value: null
-  - it: works when global is null and no local cluster value
+  - it: helper works (and fails) when global is null and no local cluster value
     set:
       global: null
+      cluster: null
     asserts:
       - equal:
           path: spec.template.spec.containers[0].env[0].value
           value: null
+
   - it: sets local cluster as cluster name even if global is set
     set:
       cluster: test-cluster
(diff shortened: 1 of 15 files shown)


library/CHART-TEMPLATE/tests/common_cluster_test.yaml
```diff
@@ -4,20 +4,40 @@ templates:
 release:
   name: my-release
   namespace: my-namespace
-set:
+
+# The name of the cluster is mandatory, to not make all the tests fail because they have no license
+# and not having to add license to all the tests I have added it to the `values.yaml`.
+# This change makes this tests a bit funky because we have to manually set local cluster to null
+#constantly to overwrite it.
```
paologallinaharbur, line 11: missing space after #

library/common-library/templates/_cluster.tpl
```diff
@@ -2,12 +2,15 @@
...
-    {{- end -}}
+{{- /* This allows us to use `$global` as an empty dict directly in case `Values.global` does not exists */ -}}
+{{- $global := index .Values "global" | default dict -}}
+
+{{- if .Values.cluster -}}
+    {{- .Values.cluster -}}
+{{- else if $global.cluster -}}
+    {{- $global.cluster -}}
+{{- else -}}
+    {{ fail "There is not cluster name definition set neither in `.global.cluster' nor `.cluster' in your values.yaml. Cluster name is required." }}
```
roobre, line 13: Not a huge fan of templates that `fail` (I did not like that about the licenseKey helper either).

Rather than this logic that is hard to override, why wouldn't we just:

```
clusterName: {{ include "common.cluster" . | required "You must supply a clusterName" }}
```

Whenever we need it?

library/CHART-TEMPLATE/tests/common_cluster_test.yaml
kang-makes: Solved in [#ddaebab](https://github.com/newrelic/helm-charts/pull/759/commits/ddaebabdcb252bc5da99b20d1bf96e3f2f172e0f)

library/CHART-TEMPLATE/tests/common_cluster_test.yaml
```diff
@@ -4,20 +4,40 @@ templates:
...
+      global: null
+      cluster: test-cluster
     asserts:
       - equal:
           path: spec.template.spec.containers[0].env[0].value
-          value: null
-  - it: works when global is null and no local cluster value
+          value: test-cluster
+
+  - it: helper works when local is null
```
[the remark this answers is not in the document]
kang-makes: You are right. I also change the order of the values set in one test to have consistency while reading it.

Solved in [#ddaebab](https://github.com/newrelic/helm-charts/pull/759/commits/ddaebabdcb252bc5da99b20d1bf96e3f2f172e0f)

library/common-library/templates/_cluster.tpl
kang-makes: We are not error-handling lots of things because we think that the user is clever enough to create good objects like in `affinity, `tolerations` or `hostNetwork: "SureBro"` instead of a boolean. But in that case we simply template it and Kubernetes will yell at the user. In this case is a dependency from us so it has to fail.

Here we have two approaches:
 - I trust what the common-library does and simply use it. I do not neither test it not ask myself what it does under the hood. I simply use it. So the common library has to do also the error handling.
 - I do not trust the common-library and retest what comes from it.

I think the first approach is the best because it *always* has to fail when no cluster name is provided so it leads to code repetition through all the charts.

Nevertheless, [In my last commit](https://github.com/newrelic/helm-charts/pull/759/commits/2304dc8cea873e705ee767b591bded35e5d013bc) I added a function to [change that behavior](https://github.com/newrelic/helm-charts/pull/759/commits/2304dc8cea873e705ee767b591bded35e5d013bc) if needed.

paologallinaharbur: I am fine with .1 and with having the fail explicit in the library.

(as always I would avoid having already the function to change the behaviour if we do not need it already)

kang-makes: We have a couple of defaults that now I see that we don't need and other functions that I did not think that I was going to need and are handy to use (like the `.value` ones).

Avoiding having the function means that when you need it you will have no access to it and you'll have to develop it, review, merge and update and test all the charts.

Having a default function is like the "de facto" style that we have everywhere and It does no harm.

(I also think you are right)

roobre: Like this
image

kang-makes: We have a deal.

kang-makes: I am going to leave this conversation unresolved so everybody can see it when doomscrolling.

roobre (approved):
Reluctantly approving 😁

library/common-library/templates/_cluster.tpl
```diff
@@ -9,8 +9,16 @@ Return the cluster
     {{- .Values.cluster -}}
 {{- else if $global.cluster -}}
     {{- $global.cluster -}}
-{{- else -}}
+{{- else if (include "common.cluster.failIfEmpty" . ) -}}
```
roobre, line 12: @paologallinaharbur made me remove this from the licenseKey helper please yell at @kang-makes as well

kang-makes: I will take a look at `nri-kubernetes` and metrics adapter for unused functions and I can/will remove them when I write/review the style guide.

kang-makes: The overriding system was removed so this reminder is no longer needed.

Solving.

PR #762, opened 2022-03-23
[newrelic-logging] Feature/761 logging pod security context

### LINKED ISSUE #761: [newrelic-logging] Allow securityContext editing in order to resolve fluent bug involving kernel limit issues

There is a bug in Fluentbit currently that results in logs not being processed when under heavy load, as  system limits imposed by the kernel are reached, more info here:
https://github.com/fluent/fluent-bit/issues/2661 

An overview of the behaviour can be seen in the above thread, and also in the NewRelic Forum issue linked in Additional Context

In order to workaround this isssue, I would like the ability to configure the pod securityPolicy via Helm installation.

We could maintain our own version of this chart, but would prefer to have the feature available

**Additional context**

Security Context Manifests: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
NewRelic Forum Issue: https://discuss.newrelic.com/t/new-relic-logging-bundle-fluentbit-failing-to-send-logs/180245

The way Helm usually handles this is by a large block in values. When generating a new template, we can see the following:
```
$ helm create demo
Creating demo
$ cat demo/templates/deployment.yaml 
apiVersion: apps/v1
kind: Deployment
...
      serviceAccountName: {{ include "demo.serviceAccountName" . }}
      securityContext:
        {{- toYaml .Values.podSecurityContext | nindent 8 }}
      containers:

$ cat demo/values.yaml 
# Default values for demo.
...

securityContext: {}
  # capabilities:
  #   drop:
  #   - ALL
  # readOnlyRootFilesystem: true
  # runAsNonRoot: true
  # runAsUser: 1000
...

### PR DESCRIPTION
**Pull request by TheQueenIsDead**
Adds in the ability to configure pod securityContext for the fluent logging pod in order to help alleviate issues described in #761, relating to an upstream issue with fluent and linux kernel system limits:
https://github.com/fluent/fluent-bit/issues/2661

  - fixes #761 

#### Checklist
- [x] Chart Version bumped

### CODE REVIEW

**Comment by roobre** [comment]
Hi @TheQueenIsDead,

Thanks a lot for your contribution!

Incidentally, we are in parallel working on adding a lot of missing options to several charts. To provide consistency among them, we have developed a library chart which already include this functionality, although it uses a different convention:
https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_security-context.tpl

I cannot give a timeline for when will this change land in the `newrelic-logging` chart, but as the convention we're moving towards is `podSecurityContext` and `containerSecurityContext`, I think it's best not to move forward with this change which uses a different convention (`securityContext`).

File: charts/newrelic-logging/templates/daemonset.yaml [hunk 1]
```diff
@@ -105,6 +105,9 @@ spec:
           resources:
 {{ toYaml .Values.resources | indent 12 }}
           {{- end }}
+          {{- if .Values.securityContext }}
+          securityContext: {{ toYaml .Values.securityContext | nindent 12 }}
+          {{- end }}
```
roobre, line 110: Chart implementations should use newrelic's common library:

https://github.com/newrelic/helm-charts/blob/d41364a102659445b03938f2a6f97d39deb38600/library/CHART-TEMPLATE/templates/deployment.yaml#L26-L29

roobre (changes requested):

TheQueenIsDead:
Thanks for showing me the _security-context.tpl as well as the CHART-TEMPLATE example @roobre, that makes the required changes incredibly clear :+1: 

Apologies for not spotting it before, I should have had a better look at the templates available instead of jumping to the Helm 3 standard project reference

TheQueenIsDead:
To clarify, would you accept the change to import the common-library from newrelic-logging, and the minimum changes to the daemonset.yaml?

I wouldn't want to change the entire daemonset prior to the the new common library being rolled out in a standardised and tested manner, but given the issue with fluentd currently we're manually maintaining a manifest for the daemonset ourselves, which is less than ideal!

roobre:
Hey @TheQueenIsDead,
No worries at all for not noticing, this is something we are developing right now and we haven't rolled it out yet :)

> To clarify, would you accept the change to import the common-library from newrelic-logging, and the minimum changes to the daemonset.yaml?

👍🏻 That would be OK from my side. Our plan is to make the finish touches to the library (which is almost complete right now) and then start rolling it out to the charts during the upcoming weeks. I would only object to changes that would change the interface the new library expects, to avoid changing the exposed values twice in a month :)

However, as this particular chart is shared with @newrelic/logging I'd like to hear their opinion as well though.

TheQueenIsDead:
Neat, thanks for explaining that! I don't want to rush the process, and definitely understand the desire to not change the requirements for the values.yaml twice in a short span, just trying to get away from maintaining our own manifests :-) 

I've added the codeblocks for securityContexts from the example and imported the common library chart into newrelic.

I noticed the last run had an issue with the dependencies, and it seems I neglected to update the requirements.lock for the bundle, so I've bumped that now as well, seems to have done the trick. Here's hoping the newrelic-logging is open to this interim advancement of the library inclusion :+1:

roobre (approved):
Thanks a lot for the change!
This is looking great. If you don't mind, I'll leave this approved but hold on merging for a bit more. The main reason for this is the `apiVersion: v1` -> `apiVersion: v2` change. We're in the process of bumping `apiVersion` for all charts, but we have a defined scheduled to do this to gracefully handle users that are still using Helm 2.

We're currently targeting to make this change during April, during which your PR will be included :)

Thanks again for your contribution!

charts/newrelic-logging/Chart.yaml
```diff
@@ -16,3 +16,7 @@ maintainers:
 keywords:
   - logging
   - newrelic
+dependencies:
+  - name: common-library
+    version: 0.14.3
+    repository: file://../../library/common-library
```
kang-makes, line 22: Missing newline at EOF

kang-makes (approved):
I am pinging the logging team for they to review and merge it :)

charts/newrelic-logging/Chart.yaml
kang-makes: @TheQueenIsDead, could you also update to the latest version of the common library, please?

TheQueenIsDead: Merged the latest master and bumped the dependency version in 6e00fbc660652762fe41b5fb25e7628e7c7fdde7

charts/newrelic-logging/Chart.yaml
```diff
@@ -18,5 +18,5 @@ keywords:
   - newrelic
 dependencies:
   - name: common-library
-    version: 0.14.3
+    version: 1.0.4
     repository: file://../../library/common-library
```
kang-makes, line 22: I did not realised previously that you were using a local path...

This has to be a URL like in other charts: https://github.com/newrelic/nri-kubernetes/blob/main/charts/newrelic-infrastructure/Chart.yaml#L14-L17

This is because, in case you need to migrate the source code somewhere else, you can still publish the chart here during the EOL time.

It removes also issues with possible differences between the local filesystem and the CI filesystem.

kang-makes (commented):
Last change. I am pinging (again) the logging team.

charts/nri-bundle-legacy/requirements.yaml
```diff
@@ -32,7 +32,7 @@ dependencies:
   - name: newrelic-logging
     repository: https://newrelic.github.io/helm-charts
     condition: logging.enabled
-    version: 1.10.9
+    version: 1.10.10
```
kang-makes, line 35: Tests are failing because of this change (also because you did not `helm dependency update`).

You don't have to update the dependencies of the bundle. We have configured renovatebot that will open PRs automatically with the update of the dependencies. Here you have an example: https://github.com/newrelic/helm-charts/pull/820

This change is a blocker, sorry.

tejunior (approved):

charts/newrelic-logging/Chart.yaml
TheQueenIsDead: No worries, I'll change this. When I opened the PR, common had not been published / used throughout the code yet, so I went with the example in the library folder. That line is certainly more recent than this commit.

I don't think the FS matters too much as it's a relative path, but I'm not arguing :-) 

I'll tidy up the last few bits when it's not so late in the evening! Neat to see the common ;ibrary chart being used, I'll certainly be taking a bit of inspiration from your setup in my own work, thanks for your time reviewing this @kang-makes 💯

kang-makes: If you even want to use that common library for your own work feel free to use ours and contribute to it. Open an issue before jumping to develop a feature in case we do an effort that is not going to be accepted so we can analyze it and we will be happy to merge it :)

Here you have an example: https://github.com/newrelic/helm-charts/pull/794#issuecomment-1118795491

You have other examples of charts using libraries in the Bitnami charts or the k8s-at-home project if you wish to take a look at them.

I am eager for you to solve these issues and merge it, we had an approval from the logging team :)

charts/nri-bundle-legacy/requirements.yaml
TheQueenIsDead: Reverted the change

charts/newrelic-logging/Chart.yaml
TheQueenIsDead: Changed URL and updated the chart lock

kang-makes:
Tests are failing, we would have to merge #831 before and merge `master` to this PR before merging it.

PR #767, opened 2022-03-29
[common-library] Add global/local custom attributes helper

kang-makes:
#### Is this a new chart

No

This allows having custom Attributes set as global/local

I wanted to copy&paster from `labels` helper but it was not merging properly so I fixed that as well. Sorry for derailing this PR but I wanted to have it solved once and for all.

#### Checklist
- [x] Chart Version bumped


library/common-library/templates/_labels.tpl
```diff
@@ -10,27 +10,30 @@ This function allows easily to add more labels to the function "common.labels"
...
-app.kubernetes.io/managed-by: {{ .Release.Service }}
-
 {{- $global := index .Values "global" | default dict -}}
 
-{{- with $global.labels }}
-{{ toYaml . }}
-{{- end }}
+{{- $labels := dict "helm.sh/chart" (include "common.naming.chart" . ) -}}
+{{- $labels = mustMergeOverwrite $labels (fromYaml (include "common.labels.selectorLabels" . )) -}}
+{{- $labels = mustMergeOverwrite $labels (dict "app.kubernetes.io/managed-by" .Release.Service) -}}
```
paologallinaharbur, line 17:
```suggestion
{{- $labels := dict "helm.sh/chart" (include "common.naming.chart" . ) -}}
{{- $selectorLabels = fromYaml (include "common.labels.selectorLabels" . ) -}}

{{- $labels = mustMergeOverwrite $labels $selectorLabels -}}
{{- $labels = mustMergeOverwrite $labels (dict "app.kubernetes.io/managed-by" .Release.Service) -}}
```

library/common-library/templates/_custom-attributes.tpl
```diff
@@ -0,0 +1,29 @@
+{{/*
+This function allows easily to overwrite custom attributes to the function "common.customAttributes"
+*/}}
+{{- define "common.customAttributes.overrideAttributes" -}}
+{{- end }}
```
roobre, line 5: I am guessing that things like the `clusterName` would be put here. Is this correct?

On the other hand, I wonder if we need this or would just be better to trust users to ` | fromYaml | mergeOverwrite` on their side.

kang-makes: Exactly that use case. This is for `nri-kubernetes` with the `clusterName` custom attribute.

library/common-library/templates/_labels.tpl
kang-makes: Done in [#ca7876b](https://github.com/newrelic/helm-charts/pull/767/commits/ca7876be0239ef6f9f93c71e9a88a0481efba545)

paologallinaharbur (approved):

library/common-library/templates/_labels.tpl
```diff
@@ -10,27 +10,31 @@ This function allows easily to add more labels to the function "common.labels"
...
-app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
-{{- end }}
-app.kubernetes.io/managed-by: {{ .Release.Service }}
-
 {{- $global := index .Values "global" | default dict -}}
 
-{{- with $global.labels }}
-{{ toYaml . }}
-{{- end }}
+{{- $labels := dict "helm.sh/chart" (include "common.naming.chart" . ) -}}
```
paologallinaharbur, line 15: NIT: would make sense to have something like this, avoiding a merge at line 19??
```suggestion
{{- $labels := dict "helm.sh/chart" (include "common.naming.chart" . ) "app.kubernetes.io/managed-by"  .Release.Service -}}
```

library/common-library/templates/_custom-attributes.tpl
roobre: Then I'd ask the question again: Why not expect chart writers to merge this on their side?

library/common-library/templates/_labels.tpl
paologallinaharbur: yeah, also this one works 👍

kang-makes: Solved in [#fd03c2a](https://github.com/newrelic/helm-charts/pull/767/commits/fd03c2a38e80ab629ed785d602d0be35479f7823)

library/common-library/templates/_custom-attributes.tpl
kang-makes: If there is in the `common-library` then it makes sense to have it in a separate helpers file like `nri-kubernetes` (example [here](https://github.com/newrelic/nri-kubernetes/pull/383/files#diff-3dc1d0b38b4813f0520e3397a987711789b05548c1fc55680d8bee2f3bc18699)) and makes it easier to follow up.

Also, we should/will test that override helper in a near future with some kind of chart for unit testing.

roobre: I'm not a huge fan of the approach being different from the `agentConfig` helpers, but as long as it's documented, I'm OK with it.

library/common-library/templates/_custom-attributes.tpl
```diff
@@ -0,0 +1,29 @@
...
+
+
+
+{{/*
+This will render custom attributes as a YAML ready to  be templated or be used with `fromYaml`.
```
roobre, line 10:
```suggestion
This will render custom attributes as a YAML ready to  be templated or be used with `fromYaml`.
Chart writers can override `common.customAttributes.overrideAttributes`, which will be included in the output of this helper.
```

PR #770, opened 2022-04-06
chore(gh-pages): updating index manually

paologallinaharbur:
Updating index manually from fork to include beta

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -4322,6 +4322,67 @@ entries:
   nri-bundle:
   - apiVersion: v1
     appVersion: "1.0"
+    created: "2022-04-06T08:12:46.492916989Z"
+    dependencies:
+      - alias: newrelic-infrastructure
+        condition: infrastructure.enabled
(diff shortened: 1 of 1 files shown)


index.yaml
```diff
@@ -4320,6 +4320,67 @@ entries:
     - https://github.com/newrelic/helm-charts/releases/download/newrelic-pixie-0.1.0-alpha.1/newrelic-pixie-0.1.0-alpha.1.tgz
     version: 0.1.0-alpha.1
   nri-bundle:
+  - apiVersion: v1
```
roobre, line 4323: Mostly out of curiosity, I wonder why its apiVersion 1 here. But it does not seem to be a problem.

roobre (approved):

index.yaml
paologallinaharbur: I believe the nri-bundle is V1, nri-kubernetes-v3 is V2

PR #772, opened 2022-04-07
chore: properly format function comments

sigilioso:

Properly format some functions comments in common-library

#### Checklist

### DIFF
--- a/library/common-library/templates/_images.tpl
+++ b/library/common-library/templates/_images.tpl
@@ -14,10 +14,10 @@ Return the proper image name
     {{- end -}}
 {{- end -}}
 
-{{/*
+{{- /*
 Return the proper image registry
 {{ include "common.images.registry" ( dict "imageRoot" .Values.path.to.the.image "context" .) }}
-*/}}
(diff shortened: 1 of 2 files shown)


library/common-library/templates/_images.tpl
```diff
@@ -14,6 +14,7 @@ Return the proper image name
     {{- end -}}
 {{- end -}}
 
+{{/*
 Return the proper image registry
 {{ include "common.images.registry" ( dict "imageRoot" .Values.path.to.the.image "context" .) }}
 */}}
```
kang-makes, line 20:
```suggestion
{{- /*
Return the proper image registry
{{ include "common.images.registry" ( dict "imageRoot" .Values.path.to.the.image "context" .) }}
*/ -}}
```

kang-makes (commented):
Comments can have also line chompers... Good to not print empty lines that are between functions.

Not needed here because it is a library that does not print/template anything but maybe it is good for you to know this ;)

Bump version and I'll approve it.

PR #774, opened 2022-04-08
[common-library] Fix podLabels to include the selector labels

kang-makes:
#### Is this a new chart

No

`common-library` is meant to be used in a way that a template includes everything needed without the user to have to merge them.

This way fix and add the ability to support `.global.podLabels` and `.podLabels` at the same time that we keep idiomatic for what `helm create` uses by default.

#### Checklist
- [x] Chart Version bumped


library/common-library/templates/_labels.tpl
```diff
@@ -65,17 +67,16 @@ This function allows easily to add more labels to the function "common.labels.po
...
 
-{{- with .Values.podLabels }}
-{{ toYaml . }}
-{{- end }}
+{{- $overrideAddPodLabels := fromYaml ( include "common.labels.overrides.addPodLabels" . ) | default dict -}}
 
-{{- with include "common.labels.overrides.addPodLabels" . }}
-{{ . }}
-{{- end }}
+{{- $podLabels := mustMergeOverwrite $selectorLabels $globalPodLabels $localPodLabels $overrideAddPodLabels -}}
```
paologallinaharbur, line 79: I have the feeling we fixed this already 🤔

kang-makes: Yes and no... When I fixed the `customAttributes` I wanted to copy and paste from the behavior or the labels and found that they not working properly so I fixed it: https://github.com/newrelic/helm-charts/commit/68dbc805c345d6959a912dc44f42645f33503c10

But I fixed `common.labels` and I didn't take time to fix the rest of the file (that I took today)

library/common-library/templates/_labels.tpl
```diff
@@ -44,11 +44,13 @@ This function allows easily to add more labels to the function "common.labels.se
 This will render the labels that should be used in deployments/daemonsets template pods as a selector.
 */}}
 {{- define "common.labels.selectorLabels" -}}
-app.kubernetes.io/name: {{ include "common.naming.name" . }}
-app.kubernetes.io/instance: {{ .Release.Name }}
-{{- if include "common.labels.overrides.addSelectorLabels" . }}
-{{ include "common.labels.overrides.addSelectorLabels" . }}
-{{- end }}
+{{- $name := dict "app.kubernetes.io/name" ( include "common.naming.name" . ) -}}
```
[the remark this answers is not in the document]
kang-makes: It was tested but the templates were not correct so tests passed. Solved [here](https://github.com/newrelic/helm-charts/pull/774/commits/52568600d4ea9474fe7c7d94ce594baea7315715).

library/CHART-TEMPLATE/templates/sidecar-deployment.yaml
```diff
@@ -16,7 +16,6 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       labels:
-        {{- include "common.labels.selectorLabels" . | nindent 8 }}
         {{- include "common.labels.podLabels" . | nindent 8 }}
```
alvarocabanas, line 19: Personal opinion, what do we win with having the selectorLabels inside the podLabels? For me it ofuscates the fact that the podLabels have the selectorLabels and will match the rule.

kang-makes: I agree that they are obfuscated and that even the user can footgun itself overriding labels that are used in the `selectorLabels` using `.global.podLabels` and `.podLabels`.

On the other hand, it is a matter of consistency, in how Yaml templating and Yaml parser work, both of the dictionaries should have to be merged or duplicated key-values could create issues depending on how the helpers are arranged.

gsanchezgavier (approved):
LGTM

PR #777, opened 2022-04-11
[common-library] Remove overriding system

### LINKED ISSUE #773: [common-library] The override system is broken
**Bug description**
The template system of Helm does support overriding a function but it is not atomically overridden in the helmchart that overrides it but in the whole deployment: https://github.com/newrelic-forks/helm/blob/b6f876a64ab6f898c8279f9b20597a0767a71cc0/pkg/engine/engine.go#L226-L227

This means that when a chart with dependencies overrides a function it is overridden on the children too, making the override system designed for the common library useless when you integrate the chart in the `nri-bundle`

**Which chart?**

`common-library` 0.x

**How to reproduce it?**
If you override a function that is meant to change the behavior of the common library is changed in all the charts.

Note that I override it at `umbrella` level and is propagated to all the helm charts that this also happens between children.

image
image

kang-makes:
#### Is this a new chart
No

  - fixes #773

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/common-library/templates/_images.tpl
+++ b/library/common-library/templates/_images.tpl
@@ -14,6 +14,8 @@ Return the proper image name
     {{- end -}}
 {{- end -}}
 
+
+
 {{- /*
 Return the proper image registry
 {{ include "common.images.registry" ( dict "imageRoot" .Values.path.to.the.image "context" .) }}
@@ -30,6 +32,8 @@ Return the proper image registry
     {{- end -}}
 {{- end -}}
 
+
+
 {{- /*
 Return the proper image repository
 {{ include "common.images.repository" .Values.path.to.the.image }}
@@ -38,6 +42,8 @@ Return the proper image repository
     {{- .repository -}}
 {{- end -}}
 
+
+
 {{- /*
 Return the proper image tag
 {{ include "common.images.tag" ( dict "imageRoot" .Values.path.to.the.image "context" .) }}
@@ -45,6 +51,8 @@ Return the proper image tag
 {{- define "common.images.tag" -}}
     {{- .imageRoot.tag | default .context.Chart.AppVersion | toString -}}
 {{- end -}}
+
+
 
 {{- /*
 Return the proper Image Pull Registry Secret Names evaluating values as templates
@@ -69,7 +77,7 @@ Return the proper Image Pull Registry Secret Names evaluating values as template
 
   {{- if gt (len $ps) 0 -}}
     {{- range $ps }}
-{{ toYaml . }}
+      {{ toYaml . | nindent 0 }}
     {{- end -}}
   {{- end -}}
 {{- end -}}
--- a/library/common-library/templates/_images.tpl
+++ b/library/common-library/templates/_images.tpl
@@ -2,7 +2,7 @@
 Return the proper image name
 {{ include "common.images.image" ( dict "imageRoot" .Values.path.to.the.image "context" .) }}
 */ -}}
-{{- define "common.images.image" -}}
+{{- define "newrelic.common.images.image" -}}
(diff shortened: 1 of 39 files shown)


library/common-library/templates/_images.tpl
```diff
@@ -69,7 +77,7 @@ Return the proper Image Pull Registry Secret Names evaluating values as template
 
   {{- if gt (len $ps) 0 -}}
     {{- range $ps }}
-{{ toYaml . }}
+      {{ toYaml . | nindent 0 }}
```
sigilioso, line 80:
```suggestion
      {{- toYaml . | nindent 0 }}
```
After including `nindent` there is a missing `{{-`, isn't there?

sigilioso:
Maybe the whole overwriting issue is a blessing in disguise. This required change may make some helpers less flexible than intended, however I think this new version of the common library will be simpler to use and it will force us to be more explicit in charts using it.

Great job 👍

I've merely pointed out what I think to be a typo before approving.

sigilioso (approved):

sigilioso (approved):
LGTM 👍

library/common-library/templates/_images.tpl
```diff
@@ -7,7 +7,7 @@ Return the proper image name
     {{- $repositoryName := include "common.images.repository" .imageRoot -}}
     {{- $tag := include "common.images.tag" ( dict "imageRoot" .imageRoot "context" .context) -}}
 
-    {{- if $registryName }}
+    {{- if $registryName -}}
```
gsanchezgavier, line 10: just out of curiosity. What is the effect of the this trim operator on conditional entries?

gsanchezgavier (approved):
LGTM

library/common-library/templates/_images.tpl
kang-makes: Exactly the same as other templates. Trim white spaces to the left. Here is an example:
```mustache
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  key: |
    1
    {{ if include "newrelic.common.verboseLog" . }}
    {{ end }}

    2
    {{ if (include "newrelic.common.nrStaging" . ) }}
    {{ end }}

    3
    {{ with include "newrelic.common.proxy" . }}
    {{ end }}
```
outputs:
```yaml
# Source: CHART-TEMPLATE/templates/test.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  key: |
    1


    2


    3
```
while with line trimmers
```mustache
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  key: |
    1
    {{- if include "newrelic.common.verboseLog" . }}
    {{- end }}

    2
    {{- if (include "newrelic.common.nrStaging" . ) }}
    {{- end }}

    3
    {{- with include "newrelic.common.proxy" . }}
    {{- end }}

```
outputs
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  key: |
    1

    2

    3
```
If you ask explicitly why to trim in the `if` is because the `if` one is going to create a new line that rarely be seen because it could be eaten by the `- end` but could be led in some cases to an empty line depending what is inside the if.

kang-makes:
```mustache
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  key: |
    1
    {{- if include "newrelic.common.verboseLog" . }}
    {{ end }}

    2
    {{- if (include "newrelic.common.nrStaging" . ) }}
    {{ end }}

    3
    {{- with include "newrelic.common.proxy" . }}
    {{ end }}
```
outputs also
```yaml
---
# Source: CHART-TEMPLATE/templates/test.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  key: |
    1

    2

    3
```

PR #785, opened 2022-04-14
common.library: include helper to handle dns-valid names with suffixes

sigilioso:

Include a naming helper to handle names with prefixes. See potential usage example [here](https://github.com/newrelic/k8s-metadata-injection/blob/bb45f8a0c1d49fd47fa96bb95cd14b8603ce327f/charts/nri-metadata-injection/templates/admission-webhooks/job-patch/job-createSecret.yaml#L5)

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/common-library/templates/_naming.tpl
+++ b/library/common-library/templates/_naming.tpl
@@ -11,19 +11,20 @@ This is an function to be called directly with a string just to truncate strings
 {{- /*
 Given a name and a suffix returns a 'DNS Valid' which always include the suffix, truncating the name if needed.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
-It fails if suffix is too long to fint in the limit set.
+If suffix is too long it gets truncated but it al`ays takes precedence over name, so a 63 chars suffix would suppress the name.
 Usage:
 {{ include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" "<my-name>" "suffix" "my-suffix" ) }}
 */ -}}
 {{- define "newrelic.common.naming.truncateToDNSWithSuffix" -}}
-{{- $maxLen := (sub 63 (len .suffix)) -}}
-
-{{- if (lt $maxLen 1) -}}
-    {{ fail (printf "The suffix chosen (%s) is hitting the Kubernetes limit of 63 chars for the object name. We cannot create %s-%s" .suffix .name .suffix) }}
-{{- end -}}
+{{- $suffix := (include "newrelic.common.naming.trucateToDNS" .suffix) -}}
+{{- $maxLen := (sub 63 (len $suffix)) -}}
 
 {{- $newName := .name | trunc ($maxLen | int) | trimSuffix "-"  -}}
-{{- printf "%s-%s" $newName .suffix -}}
+{{- if $newName -}}
+{{- printf "%s-%s" $newName $suffix -}}
+{{- else -}}
+{{ $suffix }}
+{{- end -}}
 
 {{- end -}}
 
--- a/library/common-library/templates/_naming.tpl
+++ b/library/common-library/templates/_naming.tpl
@@ -8,6 +8,27 @@ This is an function to be called directly with a string just to truncate strings
 
 
 
+{{- /*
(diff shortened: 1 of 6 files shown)


library/common-library/templates/_naming.tpl
```diff
@@ -8,6 +8,27 @@ This is an function to be called directly with a string just to truncate strings
...
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+It fails if suffix is too long to fint in the limit set.
+Usage:
+{{ include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" "<my-name>" "suffix" "my-suffix" ) }}
+*/ -}}
+{{- define "newrelic.common.naming.truncateToDNSWithSuffix" -}}
+{{- $maxLen := (sub 63 (len .suffix)) -}}
+
+{{- if (lt $maxLen 1) -}}
+    {{ fail "(newrelic.common.naming.truncateToDNSWithSuffix) suffix length is too long to compose a valid name" }}
```
kang-makes, line 22: NIT: This error is not specific enough and is going to be shown to customers... It should be more specific.
```suggestion
    {{ fail (printf "The release name you choose (%s) is hitting the Kubernetes limit of 63 chars for the object name. We cannot create %s-%s" .name .name .suffix) }}
```
I did not test my suggestion but you should get the idea ;)

kang-makes (approved):
Sounds cool. LGTM but I would rephrase the error of that helper.

library/common-library/templates/_naming.tpl
sigilioso: The problem would not be the release name but the suffix chosen instead, anyway I like the idea of being more specific, thanks!

sigilioso: Error improved here: [b4ecebf](https://github.com/newrelic/helm-charts/pull/785/commits/b4ecebf7ba33d0cdefd6f0d8708d1d4413248759)

library/common-library/templates/_naming.tpl
```diff
@@ -8,6 +8,27 @@ This is an function to be called directly with a string just to truncate strings
...
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+It fails if suffix is too long to fint in the limit set.
+Usage:
+{{ include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" "<my-name>" "suffix" "my-suffix" ) }}
+*/ -}}
+{{- define "newrelic.common.naming.truncateToDNSWithSuffix" -}}
+{{- $maxLen := (sub 63 (len .suffix)) -}}
+
+{{- if (lt $maxLen 1) -}}
+    {{ fail (printf "The suffix chosen (%s) is hitting the Kubernetes limit of 63 chars for the object name. We cannot create %s-%s" .suffix .name .suffix) }}
```
gsanchezgavier, line 22: would make sense to warn instead of fail? not sure if there are any implications on this being chopped more than "cosmetic"

sigilioso: I think there could be unexpected name collisions (it actually depends on how the helper is used). Normally it would be cosmetic but we cannot be sure...

sigilioso: As discussed, failing here is not coherent with truncating long names when there is no suffix, so I changed the helper to truncate not only the name but also the suffix before using it: [94867bc](https://github.com/newrelic/helm-charts/pull/785/commits/94867bcfb9e75d11e708c30a1836001a2d6cb47a)

library/common-library/templates/_naming.tpl
```diff
@@ -11,19 +11,20 @@ This is an function to be called directly with a string just to truncate strings
 {{- /*
 Given a name and a suffix returns a 'DNS Valid' which always include the suffix, truncating the name if needed.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
-It fails if suffix is too long to fint in the limit set.
+If suffix is too long it gets truncated but it al`ays takes precedence over name, so a 63 chars suffix would suppress the name.
```
gsanchezgavier, line 14: typo: always

sigilioso: thanks, good catch! 👍

PR #794, opened 2022-04-25
[nri-bundle] Adding annotations capabilities to deployment templates

vakaobr:
#### Is this a new chart
No

#### What this PR does / why we need it:
- Adds the possibility to set annotations in subcharts which uses deployments 

#### Checklist
- [ ] Chart Version bumped
- [X]  Variables are documented in the README.md
- [X]  Title of the PR starts with chart name (e.g. [mychartname])

PR #802, opened 2022-05-05
[common-library] Standardize the common library

kang-makes:
#### Is this a new chart
No

With all the PRs, the common library has a missing feature and some inconsistencies in the templates, testing, wrong naming, and a missing feature for the annotations.

As changing ani test will lead to a version bump I preferred to develop this missing feature in the same PR to leverage it and don't bump the version so dependabot don't open PRs to all the charts with a simple documentation change.

You can see this PR commit by commit to make the review easier.

#### Checklist
- [x] Chart Version bumped


library/common-library/DEVELOPERS.md
```diff
@@ -99,6 +99,37 @@ spec:
 
 
 
+## _annotations.tpl
+### `newrelic.common.annotations.deployment` and `newrelic.common.annotations.pod`
+These are functions that are used to annotate objects. They are configured by this `values.yaml`
+```yaml
+global:
+  deploymentAnnotations: {}  # included in all the deployments and daemonsets of all the charts
+  podAnnotations: {}  # included in all the pods of all the charts
```
paologallinaharbur, line 108: why for labels we have pod and generic labels and for annotations we have merely pod and deployment?

library/common-library/DEVELOPERS.md
```diff
@@ -99,6 +99,37 @@ spec:
 
 
 
+## _annotations.tpl
+### `newrelic.common.annotations.deployment` and `newrelic.common.annotations.pod`
+These are functions that are used to annotate objects. They are configured by this `values.yaml`
+```yaml
+global:
+  deploymentAnnotations: {}  # included in all the deployments and daemonsets of all the charts
```
paologallinaharbur, line 107: having deployment annotations added for daemonsets is counterintuitive. Could we name it in a different way?

library/common-library/DEVELOPERS.md
```diff
@@ -99,6 +99,37 @@ spec:
...
+  podAnnotations: {}  # included in all the pods of all the charts
+deploymentAnnotations: {}  # included in all the deployments and daemonsets of the chart that the chart writes is implementing
```
paologallinaharbur, line 109: the comment is a bit confusing

paologallinaharbur: "of the chart that the chart writes is implementing"?

library/common-library/README.md
```diff
@@ -32,8 +32,10 @@ Here is a list of global options:
 | global.licenseKey | licenseKey | `""` |  | This set this license key to use |
 | global.customSecretName | customSecretName | `""` |  | In case you don't want to have the license key in you values, this allows you to point to a user created secret to get the key from there |
 | global.customSecretLicenseKey | customSecretLicenseKey | `""` |  | In case you don't want to have the license key in you values, this allows you to point to which secret key is the license key located |
-| global.podLabels | podLabels | `{}` | yes | Additional labels for chart pods |
 | global.labels | labels | `{}` | yes | Additional labels for chart objects |
+| global.podLabels | podLabels | `{}` | yes | Additional labels for chart pods |
+| global.deploymentAnnotations | deploymentAnnotations | `{}` | yes | Annotations to add to deployments, statefulSets and daemonSets |
```
paologallinaharbur, line 37: here it states that we are also adding it to statefulSets, in the other docs those are not mentioned.

I would try to follow the same "UI" of labels

library/CHART-TEMPLATE/tests/cl-agent_config_test.yaml
```diff
@@ -39,6 +41,7 @@ tests:
       - equal:
           path: data.newrelic-infra\.yml
           value: |
+            # Leave a comment to the user here
```
paologallinaharbur, line 44: why adding this? To show how to create comments?

library/common-library/DEVELOPERS.md
kang-makes: Because `labels` are labeled for **all** objects while annotations are only for things that deploy pods. Would you want me to change it to something like `deployerAnnotations`?

library/common-library/DEVELOPERS.md
kang-makes: We can change it to `deployerAnnotations` as I suggested above.

library/common-library/DEVELOPERS.md
kang-makes: > of the chart that chart writer is implementing

I copied and pasted this comment from the labels so there could be an issue in that part too. I am rechecking.

library/common-library/DEVELOPERS.md
kang-makes: Let's focus on one thread, please because the issue is the same.

Let's do the follow-up in [this thread](https://github.com/newrelic/helm-charts/pull/802#discussion_r865917563).

library/common-library/DEVELOPERS.md
paologallinaharbur: Having both DeploymentAnnotations and DaemonSet annotations seems the obvious solution, but annoying logic would be needed in the control Plane component

roobre: I think this is a very convoluted use case 🤔 
If we're really sure we want to implement this i'd go with

```
workloads:
  annotaitons:
```

Just as we do with e.g. `serviceAccount`.

kang-makes: We have a `deployment.annotations` in `nri-kube-events`: https://github.com/newrelic/helm-charts/pull/794/files#diff-8de0ccfb5402d7bfb33702ff03f10bd570021abbfefb098c89c3f2932a75a729L118-L120

So we might want to follow this:
```yaml
deployment:
  annotations: {}
daemonset:
  annotations: {}
statefulset:
  annotations: {}
```

But, to be honest, there is no option that I like. I dislike less to simply put `deploymentAnnotations` and use it for Damonsets and StatefulSets...

kang-makes: As we decided offline, this feature is out of scope of the PR and the way that we like could take more time that the timebox we have to solve all the issues we have.

Reverted in [#5c4233b](https://github.com/newrelic/helm-charts/pull/802/commits/5c4233b1f6e9d72e2d18e475d5bb705fad8f4e88)

paologallinaharbur (approved):

library/common-library/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: common-library
 description: Provides helpers to provide consistency on all the charts
 
 type: library
-version: 1.0.3
```
paologallinaharbur, line 6: at the end it is just changing a comment, a patch bump can be enough

kang-makes: Sure. I am changing it before merging.

PR #806, opened 2022-05-09
[all] Chart cleanup

kang-makes:
#### Is this a new chart

No

Cleans all the DEPRECATED helm charts from this repo and adds notices in READMEs that point to where these charts live now.

#### Checklist
- [x] Chart Version bumped


charts/newrelic-infra-operator/templates/admission-webhooks/job-patch/clusterrolebinding.yaml
```diff
@@ -1,20 +0,0 @@
-{{- if (and (not .Values.customTLSCertificate) (not .Values.certManager.enabled)) }}
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
-metadata:
-  name: {{ template "newrelic-infra-operator.fullname" . }}-admission
```
paologallinaharbur, line 5: I wanted as well to delete all charts but I had mainly one doubt:
 - how can we keep releasing helmV2 security updates in the meanwhile?

paologallinaharbur: Or there is the need merely for infra V2?

kang-makes: It is a PR that enabled to release charts from the `v2` branch of `nri-kubernetes`: newrelic/nri-kubernetes#414
Here we will have the infrastructure agent chart for v2: newrelic/nri-kubernetes#415

In a follow-up PR I have to update all the charts that had been updated during the last week and I would change `nri-bundle-legacy` to point to that repository.

paologallinaharbur: yes, but I was wondering if we need to do the same for all V2 charts (eg: to bump nri-kube-events forwarder for helmV2 charts, no idea if that is needed)

kang-makes: Not from my point of view. But in case we could need that, we can discuss opening a branch in the repository or having a folder with a legacy chart in the repository where it lives now.

paologallinaharbur (approved):

PR #809, opened 2022-05-09
[nri-metadata-injection] greatly reduces default timeout for admission hooks

guhcampos:
#### Is this a new chart
No.

The Admission webhooks introduced by New Relic are a major cause of
significant performance problems in GKE private clusters, as identified
in https://github.com/newrelic/k8s-metadata-injection/issues/69. 

While the root cause is the lack of access to port 8443, the problem is
worsened because the default timeout for such webhooks is 30 seconds.

This commit reduces the default to 3 seconds, which is a lot better, but
potentially could be lowered to even less, the communication happens
mostly between K8s control plane and the in-cluster service.

I have also fixed the docs wich incorrectly stated the default to be 10
seconds in some cases, while it was in fact 30 seconds.

Note this has the pontential to further mask the problem in some
customers, reducing the performance problem to the point where people
might fail to notice it. The higher timeout is more prone to raising
people's attention. In my personal experience, however, my current
employer had this going for 6+ months before someone pinpointed the
scaling problems to New Relic.

In general, admission webooks must be fast. Very fast, so I don't think it makes sense to keep such high defaults anyway.

does not fix, but mitigates [k8s-metadata-injection#69](https://github.com/newrelic/k8s-metadata-injection/issues/69)

@alvarocabanas 
@carlossscastro 
@sigilioso 

#### Checklist
- [ ] Chart Version bumped


paologallinaharbur:
You are modifying the infra operator code that currently is deployed merely on fargate, that should be not an issue for private GKE clusters.

Moreover the source code of the two charts have been moved to:
 - https://github.com/newrelic/k8s-metadata-injection . We can carry on the discussion on that repo
 - https://github.com/newrelic/newrelic-infra-operator (no change here should be needed)

paologallinaharbur:
I saw that you already open a PR there -> https://github.com/newrelic/k8s-metadata-injection/pull/179

PR #812, opened 2022-05-10
[CI/CD] Make this repository forkable

kang-makes:
#### Is this a new chart
No

We want that contributors run pipelines inside this repo to be able to test their changes. They would have to patch their repository to make the pipeline work so that makes the contribution harder as they have to patch, run and unpatch.

This PR should solve that issue.


sigilioso (approved):
LGTM

PR #833, opened 2022-05-24
[nri-bundle] enable KSM and kube-events by default, use custom labels for bundled KSM

roobre:
KSM is a mandatory dependency of the infrastructure chart, which is enabled by default. Despite this, KSM was not. This PR makes two changes:
1. Enable KSM by default
2. Change the name of the bundled KSM to a `nrk8s` specific one. This will make autodiscovery to pick the KSM we bundle by default, rather than other KSMs in the cluster.

### KSM

Users who were using their own KSM will need to adjust the selector to the default KSM label:

```yaml
newrelic-infrastructure:
  ksm:
    config:
      selector: "app.kubernetes.io/name=kube-state-metrics"
```

### `nri-kube-events`

`nri-kube-events` is now enabled by default, following what the guided install was already doing.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/nri-bundle/values.yaml
+++ b/charts/nri-bundle/values.yaml
@@ -1,6 +1,10 @@
 newrelic-infrastructure:
   # newrelic-infrastructure.enabled -- Install the [`newrelic-infrastructure` chart](https://github.com/newrelic/nri-kubernetes/tree/main/charts/newrelic-infrastructure)
   enabled: true
+  ksm:
+    config:
+      # newrelic-infrastructure.ksm.config.selector -- Use the following selector to discover an instance of kube-state-metrics in the cluster.
+      selector: "app.kubernetes.io/name=nrk8s-kube-state-metrics"
 
 nri-prometheus:
   # nri-prometheus.enabled -- Install the [`nri-prometheus` chart](https://github.com/newrelic/nri-prometheus/tree/main/charts/nri-prometheus)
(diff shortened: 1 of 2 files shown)


charts/nri-bundle/values.yaml
```diff
@@ -11,9 +15,11 @@ nri-metadata-injection:
...
 
 kube-state-metrics:
-  # kube-state-metrics.enabled -- Install the [`kube-state-metrics` chart from the stable helm charts repository](https://github.com/kubernetes/kube-state-metrics/tree/master/charts/kube-state-metrics)
-  # This is mandatory if `infrastructure.enabled` is set to `true` and the user does not provide its own instance of KSM version >=1.8 and <=2.0
-  enabled: false
+  # kube-state-metrics.enabled -- Install the [`kube-state-metrics` chart](https://kubernetes.github.io/kube-state-metrics/). kube-state-metrics is a mandatory dependency of the integration.
+  # If kube-state-metrics >= v1.8.0 <= 1.9.8 is already installed in your cluster, you can set this to `false` and adjust `newrelic-infrastructure.ksm.config.selector` accordingly to match the installed KSM label.
+  enabled: true
+  # -- Use the following name and app label for NR-bundled KSM.
+  nameOverride: nrk8s-kube-state-metrics
```
gsanchezgavier, line 22: How about to add a new label like `newrelic.io/ksm-instance: true` just to avoid changing an existing label that it might be already used ?

roobre: I think the purpose of breaking the label is two-fold. The first one is to prevent our integration from autoselecting other KSM when ours is installed. The second one I see is to "claim ownership" of this KSM. After breaking this label, we can start tweaking it more heavily than before: Removing `collectors`, bumping its major without bumping ours, etc. I think this is desirable for us and will cause us to move faster with it, but would not work if we don't get rid of the original label.

paologallinaharbur:
@roobre Can we merge this?

charts/nri-bundle/values.yaml
```diff
@@ -1,6 +1,10 @@
 newrelic-infrastructure:
   # newrelic-infrastructure.enabled -- Install the [`newrelic-infrastructure` chart](https://github.com/newrelic/nri-kubernetes/tree/main/charts/newrelic-infrastructure)
   enabled: true
+  ksm:
+    config:
+      # newrelic-infrastructure.ksm.config.selector -- Use the following selector to discover an instance of kube-state-metrics in the cluster.
+      selector: "app.kubernetes.io/name=nrk8s-kube-state-metrics"
```
paologallinaharbur, line 7: why overriding it at bundle lavel and not at nri-kubernetes one?

PR #854, opened 2022-06-15
[newrelic-pixie] Add init container to newrelic-pixie job

aimichelle:
#### Is this a new chart
No

This job is dependent on a secret created by Pixie once the user's cluster has successfully connected and registered with Pixie Cloud. Since this registration can take some time, this job currently surfaces as a `CreateContainerConfigError` which can be confusing. Instead, this PR adds an init-container which waits for the registration to complete. This prints out helpful error messaging and properly communicates why the job hasn't started yet.

#### Checklist
- [x] Chart Version bumped


vihangm (approved):

charts/newrelic-pixie/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart for the New Relic Pixie integration.
 name: newrelic-pixie
-version: 2.0.0
+version: 2.0.1
```
sigilioso, line 4: An update of the chart version is needed because `2.0.1` was previously used.

sigilioso (approved):

htroisi (approved):

PR #887, opened 2022-07-24
[newrelic-logging] - support for appending filters

smcavallo:
Signed-off-by: smcavallo <smcavallo@hotmail.com>

#### Is this a new chart
No

In the current implementation, if you would like to add additional filters you have to copy the entire filter section and append your own.   This is not ideal.
* It is more configuration that the user will need to maintain, since it does not append to the defaults
* Users must override the default filters and lowDataMode filters - if the defaults change user will not pick up those changes
* Because of the conditional ( `config.lowDataModeFilters` OR `config.filters`) users might experience undesired behavior with filters if they toggle lowDataMode on or off. 
* The expected/assumed behavior would be that `lowDataModeFilters` would be appended to any defined filters which is not the case.

This feature adds support to always use the built-in features (lowDataMode OR default) and append additional filters.

  - fixes #

#### Checklist
- [X ] Chart Version bumped
- [X ] Variables are documented in the README.md
- [X ] Title of the PR starts with chart name (e.g. `[mychartname]`)


jsubirat (approved):
Great one, thank you!

jsubirat:
Hi @smcavallo , I'm closing this PR in favour of [this other one](https://github.com/newrelic/helm-charts/pull/609), which also includes extra inputs and outputs.

Issue #875, opened 2022-07-13
[common-library] global service account not honored


gsanchezgavier:
**Bug description**

When `global.serviceAccount.create: false` charts are rendering the serviceAccount template anyway.

**Version of Helm and Kubernetes**

`v3.6.3`

**Which chart?**

All 

**What happened?**

Sub charts still render the Service accounts instead of just selecting the one specified in the `name`

`helm template nr newrelic/nri-bundle -f values.yaml`
where `values.yaml`:
```yaml
newrelic-infrastructure:
  enabled: true
nri-prometheus:
  enabled: true
nri-metadata-injection:
  enabled: true
kube-state-metrics:
  enabled: true
  collectors:
    certificatesigningrequests: false
    ingresses: false
nri-kube-events:
  enabled: true
newrelic-logging:
  enabled: true


global:
  cluster: "fake"
  licenseKey: "fake"

  serviceAccount:
    annotations: {}
    create: false
    name: "my-service-account"
```

kang-makes:
Charts reviewed that could have this issue:
- [ ] newrelic-infrastructure
- [ ] nri-prometheus
- [ ] nri-metadata-injection
- [ ] newrelic-k8s-metrics-adapter
- [ ] kube-state-metrics
- [ ] nri-kube-events
- [ ] newrelic-logging
- [ ] newrelic-infra-operator

PR #888, opened 2022-07-25
[newrelic-logging] Add global service account

kang-makes:
#### Is this a new chart
No.

There is an issue open by GTS saying that not all the charts honor the flag `.global.serviceAccount.create` and `.global.serviceAccount.name`: #875

So at the same time I was reviewing all the charts we maintain, I made also the PR to support the logging chart too.

Enjoy :)

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-logging/templates/serviceaccount.yaml
+++ b/charts/newrelic-logging/templates/serviceaccount.yaml
@@ -1,16 +1,17 @@
-{{- if .Values.serviceAccount.create }}
+{{- if include "newrelic.common.serviceAccount.create" . -}}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  namespace: {{ .Release.Namespace }}
+  {{- if include "newrelic.common.serviceAccount.annotations" . }}
+  annotations:
+    {{- include "newrelic.common.serviceAccount.annotations" . | nindent 4 }}
+  {{- end }}
   labels:
     app: {{ template "newrelic-logging.name" . }}
     chart: {{ template "newrelic-logging.chart" . }}
     heritage: "{{ .Release.Service }}"
     release: "{{ .Release.Name }}"
-  annotations:
-  {{- if .Values.serviceAccount.annotations }}
-{{ toYaml .Values.serviceAccount.annotations | indent 4 }}
-  {{- end }}
-  name: {{ template "newrelic-logging.serviceAccountName" . }}
-{{- end -}}
+    {{- /*include "newrelic.common.labels" . | nindent 4  /!\ Breaking change /!\ */}}
+  name: {{ include "newrelic.common.serviceAccount.name" . }}
+  namespace: {{ .Release.Namespace }}
+{{- end }}
--- a/charts/newrelic-logging/tests/rbac_test.yaml
+++ b/charts/newrelic-logging/tests/rbac_test.yaml
@@ -38,7 +38,7 @@ tests:
           path: subjects[0].name
           value: sa-test
 
-  - it: RBAC points to the service account the user supplies when serviceAccount is disabled
+  - it: RBAC points to the default service account when serviceAccount is disabled
     set:
       rbac.create: true
       serviceAccount.create: false
--- a/charts/newrelic-logging/tests/rbac_test.yaml
+++ b/charts/newrelic-logging/tests/rbac_test.yaml
@@ -1,0 +1,48 @@
+suite: test RBAC creation
+templates:
+  - templates/clusterrolebinding.yaml
+release:
+  name: my-release
+  namespace: my-namespace
+tests:
+  - it: template rbac if it is configured to do it
+    set:
+      rbac.create: true
+    asserts:
+      - hasDocuments:
+          count: 1
+
+  - it: don't template rbac if it is disabled
+    set:
+      rbac.create: false
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  - it: RBAC points to the service account that is created by default
+    set:
+      rbac.create: true
+      serviceAccount.create: true
+    asserts:
+      - equal:
+          path: subjects[0].name
+          value: my-release-newrelic-logging
+
+  - it: RBAC points to the service account the user supplies when serviceAccount is disabled
+    set:
+      rbac.create: true
+      serviceAccount.create: false
+      serviceAccount.name: sa-test
+    asserts:
+      - equal:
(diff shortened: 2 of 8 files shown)


charts/newrelic-logging/templates/serviceaccount.yaml
```diff
@@ -1,16 +1,17 @@
...
     chart: {{ template "newrelic-logging.chart" . }}
     heritage: "{{ .Release.Service }}"
     release: "{{ .Release.Name }}"
-  annotations:
-  {{- if .Values.serviceAccount.annotations }}
-{{ toYaml .Values.serviceAccount.annotations | indent 4 }}
-  {{- end }}
-  name: {{ template "newrelic-logging.serviceAccountName" . }}
-{{- end -}}
+    {{- /*include "newrelic.common.labels" . | nindent 4  /!\ Breaking change /!\ */}}
```
jsubirat, line 14: Hi @kang-makes , is this commented line intended to be here?

charts/newrelic-logging/values.yaml
```diff
@@ -169,7 +169,7 @@ rbac:
 
 serviceAccount:
   # Specifies whether a ServiceAccount should be created
-  create: true
+  create:
```
jsubirat, line 172: Would it make sense to include a comment here that in order for this to be `boolean false` it needs to be empty, due to [this](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_serviceaccount.tpl#L11-L13)?

Also, the mention to `include "newrelic-logging.serviceAccount"` from [here](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_serviceaccount.tpl#L12) is obsolete now, please replace it by `include "newrelic.common.serviceAccount.name"`

charts/newrelic-logging/tests/rbac_test.yaml
```diff
@@ -0,0 +1,48 @@
...
+    set:
+      rbac.create: true
+      serviceAccount.create: false
+      serviceAccount.name: sa-test
+    asserts:
+      - equal:
+          path: subjects[0].name
+          value: sa-test
+
+  - it: RBAC points to the service account the user supplies when serviceAccount is disabled
```
jsubirat, line 41: Please modify the name of the test (it currently has the same name as the previous one).

charts/newrelic-logging/tests/rbac_test.yaml
```diff
@@ -0,0 +1,48 @@
...
+    set:
+      rbac.create: true
+      serviceAccount.create: false
+    asserts:
+      - equal:
+          path: subjects[0].name
+          value: default
```
jsubirat, line 48: Curiosity: where is this "default" value obtained from? I could not find it mentioned [here](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_serviceaccount.tpl#L42-L57).

charts/newrelic-logging/tests/rbac_test.yaml
```diff
@@ -0,0 +1,48 @@
+suite: test RBAC creation
```
jsubirat, line 1: Thank you for taking care of adding all these test cases! 🙇

jsubirat (commented):
Hi @kang-makes , I left you some comments and some doubts for my personal learning. 😄  Thank you very much for your contribution! 🙇

charts/newrelic-logging/templates/serviceaccount.yaml
kang-makes: This line is for the future self: in case you are doing a breaking change at some time, remember to change this one too to be idiomatic and create the labels using the `common-library` too.

charts/newrelic-logging/values.yaml
kang-makes: > Would it make sense to include a comment here that in order for this to be boolean false it needs to be empty, due to [this](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_serviceaccount.tpl#L11-L13)?

We can comment it. That it has to be `nil` so the common library manages the possibility to set it globally. But in the end, it defaults to true [here](https://github.com/newrelic/helm-charts/blob/ebf870a/library/common-library/templates/_serviceaccount.tpl#L34-L37
)

> Also, the mention to include "newrelic-logging.serviceAccount" from [here](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_serviceaccount.tpl#L12) is obsolete now, please replace it by include "newrelic.common.serviceAccount.name"

This is a comment that was copied and pasted ages ago and it managed to enter the common library because the PRs were huge. But in the end, it makes no difference at all.

The tests for charts do not parse the files, simply look for changes in files so making that change means having to re-release the common library and upgrade all the charts. It can wait for the next issue/feature to do it.

charts/newrelic-logging/tests/rbac_test.yaml
kang-makes: rebased to the latest `master` commit and fixed at [a85649b](https://github.com/newrelic/helm-charts/pull/888/commits/a85649b8d106382c3f0613e78a960ccf384a1ba2).

charts/newrelic-logging/tests/rbac_test.yaml
kang-makes: It is [here](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_serviceaccount.tpl#L62)

If the user specifies to not create the service account and there is no global or local set, defaults to `default`.

charts/newrelic-logging/tests/rbac_test.yaml
kang-makes: You are welcome :)

charts/newrelic-logging/values.yaml
jsubirat: Yeah, regarding the second comment, I'm totally OK with this: let's wait for the next release to introduce this change.

charts/newrelic-logging/tests/rbac_test.yaml
jsubirat: I see, thanks!

charts/newrelic-logging/templates/serviceaccount.yaml
jsubirat: I see, thanks

jsubirat (approved):

PR #890, opened 2022-07-25
[newrelic-logging] - support for exposing metrics and monitoring info

smcavallo:
Signed-off-by: smcavallo <smcavallo@hotmail.com>

#### Is this a new chart
No
Fluentbit exposes metrics and monitoring info on 2020 port by default. 
See - https://docs.fluentbit.io/manual/administration/monitoring

Users of this chart may want to expose /metrics interface so they can be ingested by the `nri-prometheus` integration.
This chart does not currently support exposing ports.
This combined with the required annotation (below) will allow the pods in this daemonset to expose metrics to be scraped by `nri-prometheus`
```
daemonSet:
  annotations:
    prometheus.io/path: /api/v1/metrics/prometheus
    prometheus.io/port: '2020'
    prometheus.io/scrape: 'true'

ports:
  - containerPort: 2020
```

  - fixes #

#### Checklist
- [x] Chart Version bumped


charts/newrelic-logging/templates/daemonset.yaml
```diff
@@ -110,6 +110,9 @@ spec:
               mountPath: /fluent-bit/etc
             - name: var
               mountPath: /var
+          {{- if .Values.ports }}
+          ports: {{ toYaml .Values.ports | nindent 12 }}
+          {{- end }}
```
jsubirat, line 115: Please add this port export to the Windows DaemonSet as well 🙏

charts/newrelic-logging/README.md
```diff
@@ -126,6 +126,7 @@ See [values.yaml](values.yaml) for the default values
...
 | `image.tag`                                                | The version of the container to pull.                                                                                                                                                                                                                                        | See value in [values.yaml]`                                                     |
+| `ports`                                                    | Any ports you wish to expose from the pod.  Ex. 2020 for metrics                                                                                                                                                                                                             | `[]`                                                                            |
```
jsubirat, line 129: Or even, I was thinking on making this a boolean called `exposeMetrics`, which internally results in exposing the 2020 port that Fluent Bit uses. I believe it could be more user friendly. What do you think?

jsubirat (commented):
Thank you @smcavallo , this is indeed a great contribution. Monitoring the Fluent Bit pod with Prometheus could be really useful in some scenarios. I left you some comments, let me know what you think. Thanks!

charts/newrelic-logging/README.md
smcavallo: @jsubirat - yes agree that an `exposeMetrics` would be useful.  If you can accept this PR then I can add that functionality in a separate PR.   The ability to expose ports may be useful outside of any default prometheus implementation.

smcavallo:
@jsubirat  - thank you so much for reviewing and the feedback.  Changes have been implemented.  Hoping you can re-review.

danybmx (approved):
Hey thanks for this changes @smcavallo!!, I've just added a change in the windows-daemon as all the .Values should be prefixed with $ because we're inside a `range` for all windows versions.

PR #906, opened 2022-08-19
[newrelic-logging] add checksum of configmap to deamonset to allow for an auto-restart on config change

YentlFrickx:
Add checksum of configmap to deamonset to allow for an auto-restart on config change

#### Is this a new chart
No, addition to existing chart

This adds the functionality to auto-restart the deamonset of newrelic-logging when the configmap was changed.

#### Checklist


danybmx:
Hi @YentlFrickx!

Indeed, we need to include this change in the DaemonSet to be able to re-create the pod if the ConfigMap is updated.

Thanks for your contribution! As this was from some time ago, I just update the branch to master and aligned the change to match how this is done in other charts like here https://github.com/newrelic/nri-kubernetes/blob/main/charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml#L31-L32

I hope it's ok for you, and thanks again for the contribution.

jsubirat (approved):
Thank you for this contribution 🙇‍♂️

PR #907, opened 2022-08-25
chore: adjust codespell settings

sigilioso:
#### Is this a new chart

No

It adjust codespell settings to ignore the whole `.github/actions/chart-version-bumper` directory, addressing [this kind of failures](https://github.com/newrelic/helm-charts/runs/7956429281?check_suite_focus=true).

It also fixes a spelling error in newrelic-logging, so chart version is updated.


kang-makes (approved):

sigilioso:
As we are deleting chart-version bumper in #914, this is not needed anymore.

PR #910, opened 2022-08-26
[common-library] fix truncate to dns corner case

### LINKED ISSUE #902: [common-library] Tests for `newrelic.common.naming.truncateToDNSWithSuffix` are incorrect
### Bug description

The tests are located here: https://github.com/newrelic/helm-charts/blob/a0b1859/library/CHART-TEMPLATE/tests/cl-naming_suffix_test.yaml#L16-L22

Tests are passing but they are incorrect:

`% echo -n this-job-suffix-is-too-long-so-it-is-not-possible-to-include-at | wc -c`
> 63

`% echo -n including-full-name-as-it-is-would-way-be-longer-than-job-suffix | wc -c`
> 64

### What happened?

`helm install newrelic/nri-bundle new-relic-agent -f values.yaml`

With a proper values file that has a cluster and a license key, it should template but fails with this error:

> `Error: failed pre-install: warning: Hook pre-install nri-bundle/charts/nri-metadata-injection/templates/admission-webhooks/job-patch/job-createSecret.yaml failed: Job.batch "new-relic-agent-installer-nri-metadata-injectio-admission-create" is invalid: spec.template.labels: Invalid value: "new-relic-agent-installer-nri-metadata-injectio-admission-create": must be no more than 63 characters`

### What you expected to happen?**

The name should be templated correctly so the Helm releases could install properly.

### Anything else we need to know?

A workaround until we solve this issue is to make the release name shorter than 25 chars.

sigilioso:
#### Is this a new chart
No

It fixes a bug in the common.library's  `truncateToDNSWithSuffix` helper.

  - fixes #902 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/CHART-TEMPLATE/tests/cl-naming_suffix_test.yaml
+++ b/library/CHART-TEMPLATE/tests/cl-naming_suffix_test.yaml
@@ -15,11 +15,11 @@ tests:
 
   - it: name is truncated but suffix included
     set:
-      fullnameOverride: including-full-name-as-it-is-would-be-way-be-longer-than-limit
+      fullnameOverride: including-full-name-as-it-is-would-be-way-longer-than-limit
     asserts:
       - equal:
           path: metadata.name
-          value: including-full-name-as-it-is-would-be-way-be-longer-job-suffix
+          value: including-full-name-as-it-is-would-be-way-longer-tha-job-suffix
 
   - it: suffix is longer than maximum length supported
     set:
--- a/library/CHART-TEMPLATE/tests/cl-naming_suffix_test.yaml
+++ b/library/CHART-TEMPLATE/tests/cl-naming_suffix_test.yaml
@@ -15,17 +15,17 @@ tests:
 
   - it: name is truncated but suffix included
     set:
-      fullnameOverride: including-full-name-as-it-is-would-way-be-longer-than-limit
+      fullnameOverride: including-full-name-as-it-is-would-be-way-be-longer-than-limit
     asserts:
       - equal:
           path: metadata.name
-          value: including-full-name-as-it-is-would-way-be-longer-than-job-suffix
+          value: including-full-name-as-it-is-would-be-way-be-longer-job-suffix
 
   - it: suffix is longer than maximum length supported
     set:
       fullnameOverride: fno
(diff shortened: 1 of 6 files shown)


sigilioso:
We would need to merge #907 in order to the unrelated fix spellcheck issues.

library/CHART-TEMPLATE/tests/cl-naming_suffix_test.yaml
```diff
@@ -15,17 +15,17 @@ tests:
...
   - it: suffix is longer than maximum length supported
     set:
       fullnameOverride: fno
-      jobSuffix: this-job-suffix-is-too-long-so-it-is-not-possible-to-include-at-it-is
+      jobSuffix: this-job-suffix-is-too-long-so-it-is-not-possible-to-include-as-it-is
     asserts:
       - equal:
           path: metadata.name
-          value: this-job-suffix-is-too-long-so-it-is-not-possible-to-include-at
+          value: this-job-suffix-is-too-long-so-it-is-not-possible-to-include-as
```
kang-makes, line 31: NIT: Could we add tests for prefixes and suffixes with exactly 62 and 63 chars?

My rationale is to avoid other issues that we could have in case we fail again doing the maths. I would also want to test the behavior of the dash in these extreme cases.

kang-makes (approved):
`% echo -en "including-full-name-as-it-is-would-be-way-longer-tha-job-suffix" | wc -c`
```
63
```

`% echo -en this-job-suffix-is-too-long-so-it-is-not-possible-to-include-as | wc -c`
```
63
```

Left a nit but LGTM.

library/CHART-TEMPLATE/tests/cl-naming_suffix_test.yaml
sigilioso: Sure, thanks for pointing out!

sigilioso: I added some tests to cover those cases here: [9926128](https://github.com/newrelic/helm-charts/pull/910/commits/9926128253f4ba1396fe2013284acf80ca0c5432)

PR #924, opened 2022-08-30
[CI] Test `nri-bundle` integrity

### LINKED ISSUE #883: [nri-bundle] The dependency to `common-library` has to be aligned through all the dependencies
To understand this issue you should have the context about the overriding system that we removed months ago: https://github.com/newrelic/helm-charts/issues/773

TL;DR: all the names of the helpers in Helm have to be unique and if not they are overridden in a deterministic but not controllable way.

Because of that, all the dependencies of the `nri-bundle` that use `common-library` should have the same version so the helpers are overridden with the same implementation of the function.

We should add a test in which all the dependencies of `nri-bundle` share the same version of `common-library` when a dependency of `nri-bundle` is changed. We should not test that is the latest version (as it could have an issue that should be fixed).

kang-makes:
#### Is this a new chart
No.

We need that all the dependencies of the `nri-bundle` have the same version of the `common-library`. For that, the PRs of dependencies of the `nri-bundle` must be grouped in the same PR by renovatebot and we should parse the dependencies for the versions and compare them.

  - fixes #883

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -13,11 +13,12 @@
   ],
   "packageRules": [
     {
-      // Group all GHA bumps together in a single PR.
-      "matchManagers": [
-        "github-actions"
+      // Group all nri-bundle dependencies together in a single PR.
+      "matchPackageNames": [
+        "newrelic-infrastructure"
       ],
-      "groupName": "Github Actions"
+      "groupName": "Bundle dependencies",
+      "groupSlug": "nri-bundle-dependencies"
     },
     {
       // Disable major updates of Kube State Metrics.
@@ -50,5 +51,5 @@
       ],
       "enabled": true
     }
-  ],
+  ]
 }
--- a/.github/workflows/lint_test_charts.yaml
+++ b/.github/workflows/lint_test_charts.yaml
@@ -19,25 +19,26 @@ jobs:
   lint-test:
     name: Lint and test charts
     runs-on: ubuntu-latest
+    outputs:
+      charts-changed: ${{ steps.changed.outputs.charts }}
+      bundle-changed: ${{ steps.changed.outputs.bundle }}
     steps:
       - name: Checkout
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
 
-      - uses: azure/setup-helm@v1
       - uses: actions/setup-python@v2
-
-      - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.1.0
+      - uses: helm/chart-testing-action@v2.1.0
+      - uses: azure/setup-helm@v3
+        with:
+          version: 'v3.0.0'
 
       - name: Set up helm-unittest
-        run: |
-          helm plugin install https://github.com/quintush/helm-unittest
+        run: helm plugin install https://github.com/quintush/helm-unittest
 
       - name: Lint charts
-        run: |
-          ct lint --config .github/ct-lint.yaml
+        run: ct lint --config .github/ct-lint.yaml
 
       - name: Run unit tests
         run: |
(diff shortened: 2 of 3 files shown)


.github/renovate.json5
```diff
@@ -13,11 +13,12 @@
   ],
   "packageRules": [
     {
-      // Group all GHA bumps together in a single PR.
-      "matchManagers": [
-        "github-actions"
+      // Group all nri-bundle dependencies together in a single PR.
```
gsanchezgavier, line 16: nice!

.github/workflows/lint_test_charts.yaml
```diff
@@ -66,3 +89,56 @@ jobs:
...
+      - name: Decompress nri-bundle and all its dependencies recursively
+        run: |
+          while [ `find charts/nri-bundle -name \*.tgz -or -name \*.tar.gz | wc -l` -gt 0 ]; do
+            find charts/nri-bundle -name \*.tgz -or -name \*.tar.gz | while read line; do
+              echo "Decompressing $line..."
+              tar xzf "$line" -C $(dirname "$line")
+              echo "Removing $line..."
+              rm "$line"
+            done
+          done
```
kang-makes, line 129: When you use `helm package` (that the chart releaser uses) it packages the dependency uncompressed but it could not be the case ins case somebody does something manually and we start supporting external charts or something weird. That is the reason I prefer to do not trust charts we are depending even if they are from us.

.github/workflows/lint_test_charts.yaml
```diff
@@ -48,13 +49,35 @@ jobs:
...
+            echo "::set-output name=charts::true"
+          fi
+          if [[ -n "$(echo -n "$charts" | grep charts/nri-bundle)" ]]; then
+            echo "::set-output name=bundle-changed::true"
           fi
 
+  install-test:
+    name: Test chart installability
+    runs-on: ubuntu-latest
+    if: ${{ needs.lint-test.outputs.charts-changed == 'true' }}
```
gsanchezgavier, line 67: what about just adding a temporal commit modifing the logging chart (which has outdated versions of library) to check the pipeline fail, then another commit updating the library version of on the chart checking that it passes ?

gsanchezgavier (approved):
Looks good! i just commented a possible way you could test this in this same PR

.github/workflows/lint_test_charts.yaml
```diff
@@ -66,3 +89,56 @@ jobs:
...
+          done
+
+      - name: Find versions of the common library
+        run: |
+          (
+            echo CHART NAME: common-library VERSION
+            find charts/nri-bundle -name Chart.yaml | \
+              xargs yq eval-all '{ .name: (.dependencies | map(select(contains({"name": "common-library"}))) | .[].version ) }' | \
+              tee versions.txt;
+          ) | column -t -s:
```
kang-makes, line 138: This `tee` the versions to stdout and `versions.txt` to be able to parse them and also have a debug of which chart is offending this test.

This should print something like this:
```
CHART NAME                     common-library VERSION
newrelic-infra-operator        1.0.5
newrelic-k8s-metrics-adapter   1.0.5
nri-prometheus                 1.0.5
nri-metadata-injection         1.0.5
nri-kube-events                1.0.5
newrelic-infrastructure        1.0.5
newrelic-logging               1.0.5
```

I tested also with an aliased `common-library` with a different version for testing purposes and it printed this:
```
CHART NAME                     common-library VERSION
newrelic-infra-operator        1.0.5
newrelic-infra-operator        1.0.6
newrelic-k8s-metrics-adapter   1.0.5
nri-prometheus                 1.0.5
nri-metadata-injection         1.0.5
nri-kube-events                1.0.5
newrelic-infrastructure        1.0.5
newrelic-logging               1.0.5
```

gsanchezgavier (approved):

.github/workflows/lint_test_charts.yaml
gsanchezgavier: oh but this runs just on the bundle 🤔  not sure if my proposition makes sense

.github/workflows/lint_test_charts.yaml
```diff
@@ -66,3 +89,56 @@ jobs:
...
+
+          echo ""
+          echo Versions of the common library detected:
+          cut -d\  -f2 versions.txt | sort -u
```
kang-makes, line 142: The more to be able to debug, the better:
```
Versions of the common library detected:
1.0.5
1.0.6
```

.github/workflows/lint_test_charts.yaml
kang-makes: No, this runs if any chart is modified.
```yaml
if: ${{ needs.lint-test.outputs.charts-changed == 'true' }}
[...]     
if: ${{ needs.lint-test.outputs.bundle-changed == 'true' }}
```
I cannot test how GitHub actions or renovate bot is going toe behave. Renovate should open a single PR and in that PR we should be able to test it properly.

PR #931, opened 2022-08-31
[common-library] Add default registry to `images.image` helper

### LINKED ISSUE #895: [all] chart with patch jobs does not honor `common-library`'s image registry
## Bug description
The common library expects to have empty fields to choose between a global set registry and a local override from the user: https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_images.tpl#L19-L33

But there are charts that has the registry field set so a user cannot use the global to override all the registries.

## Which chart?

Here is a list of offending charts:

 - [x] `newrelic-k8s-metrics-adapter`: https://github.com/newrelic/newrelic-k8s-metrics-adapter/blob/24fd202/charts/newrelic-k8s-metrics-adapter/values.yaml#L129-L133
 - [x] `newrelic-infra-operator`: https://github.com/newrelic/newrelic-infra-operator/blob/f6d1d59/charts/newrelic-infra-operator/values.yaml#L28-L35
 - [x] `k8s-metadata-injection`: https://github.com/newrelic/k8s-metadata-injection/blob/8b6ee3a/charts/nri-metadata-injection/values.yaml#L22-L29

## What happened?

If you set in your values a `.global.images.registry` it is ignored for the image that is used for the patch job.

## What you expected to happen?

That the chart has a helper for those images and in case of it being empty the helper fallbacks to `k8s.gcr.io` registry on those images. Maybe implement it in the common library as it is a shared behavior.

kang-makes:
#### Is this a new chart
No

This adds `newrelic.common.images.image` the possibility to add a default registry that is not in the `values.yaml` so it can be overridden by a globally set registry.

  - This is the first step to fix #895

#### Checklist
- [x] Chart Version bumped


library/common-library/DEVELOPERS.md
```diff
@@ -189,9 +189,10 @@ we are going to use during the documentation of these functions:
 
 ```yaml
 global:
-  registry: nexus-3-instance.internal.clients-domain.tld
+  images:
+    registry: nexus-3-instance.internal.clients-domain.tld
 jobImage:
-  registry: k8s.gcr.io
+  registry: # defaults to "example.tld" when empty in these examples
```
marcsanmi, line 195: I have a question. Here you mean it defaults to `nexus-...`, isn't it?

marcsanmi (approved):
LGTM, just left a question 👍

sigilioso (commented):
Thanks for addressing this 👍 

I simply left a suggestion to check a comment.

library/common-library/DEVELOPERS.md
kang-makes: No, this is the in the docs, not the tests: https://github.com/newrelic/helm-charts/pull/931/files/4fcdd11cada8d32ff938e3a9768662930414d16b#diff-ead97ed97e34739841bc3318ba1886dbdeb49568c970a273c57acc8f7001c2a4R227

library/common-library/templates/_images.tpl
```diff
@@ -18,20 +18,29 @@ Return the proper image name
 
 {{- /*
 Return the proper image registry
-{{ include "newrelic.common.images.registry" ( dict "imageRoot" .Values.path.to.the.image "context" .) }}
+{{ include "newrelic.common.images.registry" ( dict "imageRoot" "defaultRegistry" "your.private.registry.tld" .Values.path.to.the.image "context" .) }}
```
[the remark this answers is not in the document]
kang-makes: Good catch!

PR #939, opened 2022-09-07
[synthetics-job-manager]: Adding resource requests and limits to runtimes

vkasanneni-nr:
#### Is this a new chart? 
NO. 

Adds resource request and limits to runtimes. This is a bug ticket that was reported by ANZ: [NR-46521](https://issues.newrelic.com/browse/NR-46521). 

Previously, what was happening is that the resources for the runtimes were just using the resource request and limits defined in the job manager values yaml file. 

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/synthetics-job-manager/charts/node-browser-runtime/values.yaml
+++ b/charts/synthetics-job-manager/charts/node-browser-runtime/values.yaml
@@ -28,7 +28,13 @@ image:
 ## The AppArmor profile name that will be applied to the pods. If set, then the AppArmor profile must exist on the Kubernetes node(s) for this to work.
 # appArmorProfileName: ""
 
-resources: {}
+resources:
+  requests:
+    cpu: "500m"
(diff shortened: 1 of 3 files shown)


charts/synthetics-job-manager/charts/node-browser-runtime/values.yaml
```diff
@@ -28,7 +28,13 @@ image:
 ## The AppArmor profile name that will be applied to the pods. If set, then the AppArmor profile must exist on the Kubernetes node(s) for this to work.
 # appArmorProfileName: ""
 
-resources: {}
+resources:
+  requests:
+    cpu: "500m"
```
kondracek-nr, line 33: should we have a lower default cpu size? 500m is half a cpu, which can really limit how many runner pods can get scheduled to a node & result in lots of pending pods (depending on how the user sets parallelism).

PR #948, opened 2022-09-14
[synthetics-job-manager]: Adding resource limits and request for runtimes

vkasanneni-nr:
#### Is this a new chart? 
NO

Adds resource request and limits to runtimes. This is a bug ticket that was reported by ANZ: [NR-46521](https://issues.newrelic.com/browse/NR-46521).

**Promise these numbers didn't come out of thin air!!!! These are actually the old minion request / limit configs based on job type. I used them as a jumping off point and didn't need to increase them because they are working.**

Node Runtime: 
7FB5B327-7991-43E3-99EA-2152986D17EA_1_201_a
API Runtime: 
6F9EB60D-A869-4C91-9DA2-4BBFFB8E1C15_1_201_a

Proof of concept: 

These two are run on the eks cluster with the new resource limits and request I implemented in this PR and it shows that the values defined here will pass checks without any problem.

Monitor 1 (Scripted Browser Monitor) : https://staging.onenr.io/0PoR87rG6wG
Monitor 2 (Scripted API Monitor): https://staging.onenr.io/09MR2NvX2RY

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/synthetics-job-manager/charts/node-api-runtime/values.yaml
+++ b/charts/synthetics-job-manager/charts/node-api-runtime/values.yaml
@@ -31,7 +31,13 @@ image:
 ## The AppArmor profile name that will be applied to the pods. If set, then the AppArmor profile must exist on the Kubernetes node(s) for this to work.
 # appArmorProfileName: ""
 
-resources: {}
+resources:
+  requests:
+    cpu: "500m"
+    memory: "1250Mi"
+  limits:
+    cpu: "750m"
+    memory: "2500Mi"
 
 podAnnotations: {}
 
--- a/charts/synthetics-job-manager/charts/node-browser-runtime/values.yaml
+++ b/charts/synthetics-job-manager/charts/node-browser-runtime/values.yaml
@@ -31,7 +31,13 @@ image:
 ## The AppArmor profile name that will be applied to the pods. If set, then the AppArmor profile must exist on the Kubernetes node(s) for this to work.
 # appArmorProfileName: ""
 
-resources: {}
+resources:
+  requests:
+    cpu: "500m"
+    memory: "2000Mi"
+  limits:
+    cpu: "750m"
+    memory: "3000Mi"
 
 podAnnotations: {}
 
--- a/charts/synthetics-job-manager/Chart.yaml
+++ b/charts/synthetics-job-manager/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: synthetics-job-manager
 description: New Relic Synthetics Containerized Job Manager
 type: application
-version: 1.0.3
+version: 1.0.4
 appVersion: release-155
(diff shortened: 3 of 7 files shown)


charts/synthetics-job-manager/charts/node-api-runtime/values.yaml
```diff
@@ -31,7 +31,13 @@ image:
...
 # appArmorProfileName: ""
 
-resources: {}
+resources:
+  requests:
+    cpu: "500m"
+    memory: "1250Mi"
+  limits:
+    cpu: "750m"
+    memory: "2500Mi"
```
toddwest, line 40: Could you also put this same block up in the `synthetics-job-manager/values.yaml` file under the `node-api-runtime` section?

This will ensure that users who deploy everything via the root level synthetics-job-manager will be able to get the proper defaults. The values.yaml file here would most likely only be used if someone wanted to deploy this node-api-runtime subchart as a standalone deployment. So it still has value but I think we want it in both places.

charts/synthetics-job-manager/charts/node-browser-runtime/values.yaml
```diff
@@ -31,7 +31,13 @@ image:
...
 # appArmorProfileName: ""
 
-resources: {}
+resources:
+  requests:
+    cpu: "500m"
+    memory: "2000Mi"
+  limits:
+    cpu: "750m"
+    memory: "3000Mi"
```
toddwest, line 40: Could you also put this same block up in the `synthetics-job-manager/values.yaml` file under the `node-browser-runtime` section?

This will ensure that users who deploy everything via the root level synthetics-job-manager will be able to get the proper defaults. The values.yaml file here would most likely only be used if someone wanted to deploy this node-browser-runtime subchart as a standalone deployment. So it still has value but I think we want it in both places.

toddwest (changes requested):
Just a few minor tweaks. We'll probably also want to make sure we do this same thing for the ping runtime, unless it's a lot more work and then we can save it for another PR.

charts/synthetics-job-manager/charts/node-api-runtime/values.yaml
vkasanneni-nr: Nice catch! I feel so stupid for not seeing that, but it should be fixed now! Thank you!

charts/synthetics-job-manager/charts/node-browser-runtime/values.yaml
vkasanneni-nr: Nice catch! I feel so stupid for not seeing that, but it should be fixed now! Thank you!

vkasanneni-nr:
@toddwest As for the ping runtime, I will address it in another PR, I was just made aware of the scope change for the ticket, I will add in another PR.

toddwest (approved):
👍

PR #982, opened 2022-11-10
test: include script to check bundle dependencies

sigilioso:
#### Is this a new chart
no

Include a shell-script based check to test nri-bundle dependencies chart conditions

### DIFF
--- a/.github/workflows/lint_test_charts.yaml
+++ b/.github/workflows/lint_test_charts.yaml
@@ -99,19 +99,18 @@ jobs:
         with:
           version: 'v3.0.0'
 
+      - name: Install yq dependency
+        run: |
+          sudo wget -O /usr/local/bin/yq https://github.com/mikefarah/yq/releases/download/v4.9.3/yq_linux_amd64
+          sudo chmod +x /usr/local/bin/yq
+
       - name: Add helm repositories
         run: |
-          helm repo add newrelic-helm-charts https://newrelic.github.io/helm-charts
-          helm repo add newrelic-cdn-helm-charts https://helm-charts.newrelic.com
-          helm repo add newrelic-infrastructure https://newrelic.github.io/nri-kubernetes
-          helm repo add nri-prometheus https://newrelic.github.io/nri-prometheus
-          helm repo add newrelic-prometheus-configurator https://newrelic.github.io/newrelic-prometheus-configurator
-          helm repo add nri-metadata-injection https://newrelic.github.io/k8s-metadata-injection
-          helm repo add newrelic-k8s-metrics-adapter https://newrelic.github.io/newrelic-k8s-metrics-adapter
-          helm repo add kube-state-metrics https://kubernetes.github.io/kube-state-metrics
-          helm repo add nri-kube-events https://newrelic.github.io/nri-kube-events
-          helm repo add pixie-operator-chart https://pixie-operator-charts.storage.googleapis.com
-          helm repo add newrelic-infra-operator https://newrelic.github.io/newrelic-infra-operator
+          yq '.chart-repos[]' .github/ct-lint.yaml | while read line; do
+            repo_name=$(cut -d= -f1 <<<"$line");
+            repo_url=$(cut -d= -f2- <<<"$line");
+            helm repo add "$repo_name" "$repo_url";
+          done
 
       - name: Build nri-bundle dependencies
         run: helm dependency update ./charts/nri-bundle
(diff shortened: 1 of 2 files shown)


.github/workflows/lint_test_charts.yaml
```diff
@@ -141,3 +141,34 @@ jobs:
           cut -d\  -f2 versions.txt | sort -u
 
           [ $(cut -d\  -f2 versions.txt | sort -u | wc -l) -eq 1 ] || exit 1
+
+  bundle-dependencies-check:
+    name: Test the bundle dependency conditions
+    runs-on: ubuntu-latest
+    if: ${{ needs.lint-test.outputs.bundle-changed == 'true' }}
+    needs: [ lint-test ]
```
sigilioso, line 149: It will be executed only when something is changed, but I tried it works by commenting this lines. Example: https://github.com/newrelic/helm-charts/actions/runs/3438721588/jobs/5735147287#step:6:1

scripts/check-bundle-dependencies.sh
```diff
@@ -0,0 +1,107 @@
...
+
+function checkDependencies()
+{
+    local values="$1"
+    local expected_charts="$2"
+    [[ "$values" == "" ]] && local values_set="global.cluster=test,global.licenseKey=fake" || local values_set="global.cluster=test,global.licenseKey=fake,$values"
+    chart_sources=$(mktemp)
+    helm template charts/nri-bundle --set $values_set | grep "# Source" > $chart_sources
+    # Check no unexpected chart would be installed
+    local expected_charts_re=${expected_charts// /|} # replace ' ' with '|' "x y" -> "x|y"
```
kang-makes, line 15: If you don't use quotes, multiple spaces become only one space.
```suggestion
    local expected_charts_re="${expected_charts// /|}" # replace ' ' with '|' "x y" -> "x|y"
```

scripts/check-bundle-dependencies.sh
```diff
@@ -0,0 +1,107 @@
...
+    local unexpected_documents_count=$(cat $chart_sources | grep -v -E $expected_charts_re | wc -l | xargs)
```
kang-makes, line 16: Same here (and almost everywhere). If the is a space, they will become two arguments and `grep` will take the first one as the expression and the second one as a file name. It will fail with a weird "cannot stat file" error.
```suggestion
    local unexpected_documents_count=$(cat $chart_sources | grep -v -E "$expected_charts_re" | wc -l | xargs)
```

kang-makes (commented):
Most of the issues are NITs because they will fail if there are spaces in the paths but having spaces is evil so everything should work nicely.

I left comments, let me know what you think and I would re-review and approve it :)

.github/workflows/lint_test_charts.yaml
```diff
@@ -141,3 +141,34 @@ jobs:
...
+          helm repo add newrelic-cdn-helm-charts https://helm-charts.newrelic.com
+          helm repo add newrelic-infrastructure https://newrelic.github.io/nri-kubernetes
+          helm repo add nri-prometheus https://newrelic.github.io/nri-prometheus
+          helm repo add newrelic-prometheus-configurator https://newrelic.github.io/newrelic-prometheus-configurator
+          helm repo add nri-metadata-injection https://newrelic.github.io/k8s-metadata-injection
+          helm repo add newrelic-k8s-metrics-adapter https://newrelic.github.io/newrelic-k8s-metrics-adapter
+          helm repo add kube-state-metrics https://kubernetes.github.io/kube-state-metrics
+          helm repo add nri-kube-events https://newrelic.github.io/nri-kube-events
+          helm repo add pixie-operator-chart https://pixie-operator-charts.storage.googleapis.com
+          helm repo add newrelic-infra-operator https://newrelic.github.io/newrelic-infra-operator
```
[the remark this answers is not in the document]
sigilioso: It is a little bit more cryptic and we also need to install `yq` but I think it does totally worth it: the less places we are repeating the repos the better!

Really good suggestion! I will update this kind of repo add in the pipeline.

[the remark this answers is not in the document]
sigilioso: The installing `yq` concern was only because I didn't know it is already in the ubuntu runners (I saw we install it in [other pipeline](https://github.com/newrelic/newrelic-prometheus-exporters-packages/blob/main/.github/workflows/prerelease.yml#L29) and I though we needed it (even so I already thought it does worth it).

Regarding being a little bit more cryptic, I think it is... But, as I see it, it is still easy to understand.

scripts/check-bundle-dependencies.sh
sigilioso: I have already updated it, please take another look when you have a while.

kang-makes (approved):
LGTM. Thank you for taking care...

PR #998, opened 2022-12-01
Add KMS upgrade path workaround

marcsanmi:
This PR fixes an issue while upgrading the `nri-bundle` with KSM v2.

For more details, check this [issue](https://github.com/prometheus-community/helm-charts/issues/568).

### DIFF
--- a/charts/nri-bundle/values.yaml
+++ b/charts/nri-bundle/values.yaml
@@ -14,6 +14,9 @@ kube-state-metrics:
   # kube-state-metrics.enabled -- Install the [`kube-state-metrics` chart](https://github.com/kubernetes/kube-state-metrics/tree/master/charts/kube-state-metrics) from the stable helm charts repository.
   # This is mandatory if `infrastructure.enabled` is set to `true` and the user does not provide its own instance of KSM version >=1.8 and <=2.0
   enabled: false
+  # As the chart was migrated to https://prometheus-community.github.io/helm-charts, it comes with labels updated. So, when upgrading the chart if fails as it tries
+  # to update the deployment spec.selector which is an immutable field.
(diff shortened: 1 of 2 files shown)


charts/nri-bundle/values.yaml
```diff
@@ -14,6 +14,9 @@ kube-state-metrics:
   # kube-state-metrics.enabled -- Install the [`kube-state-metrics` chart](https://github.com/kubernetes/kube-state-metrics/tree/master/charts/kube-state-metrics) from the stable helm charts repository.
   # This is mandatory if `infrastructure.enabled` is set to `true` and the user does not provide its own instance of KSM version >=1.8 and <=2.0
   enabled: false
+  # As the chart was migrated to https://prometheus-community.github.io/helm-charts, it comes with labels updated. So, when upgrading the chart if fails as it tries
```
sigilioso, line 17:
```suggestion
  # As the chart was migrated to https://prometheus-community.github.io/helm-charts, it comes with labels updated. So, when upgrading the chart if fails as it tries
```

charts/nri-bundle/values.yaml
```diff
@@ -14,6 +14,9 @@ kube-state-metrics:
...
+  # to update the deployment spec.selector which is an immutable field.
```
sigilioso, line 18:
```suggestion
  # As the chart was migrated to https://prometheus-community.github.io/helm-charts, it comes with labels updated. So, when upgrading the chart it fails as it tries
  # to update the deployment spec.selector which is an immutable field.
  # See: https://github.com/prometheus-community/helm-charts/issues/568 for details.
```

sigilioso (approved):
🚀

paologallinaharbur (approved):

PR #1010, opened 2022-12-27
[nri-bundle] Add configuration for Argo app creation

maxlemieux:
Pixie requires some chart values to be set at installation time. The guided installer sets the minimum required values for the Pixie integration with New Relic to work as expected, but three of those settings are missing from the `values.yaml` in the chart repo: 

```
newrelic-pixie.apiKey
pixie-chart.deployKey
pixie-chart.clusterName
```

When installing the New Relic chart bundle with Argo CD using the Argo CD console, it is not possible to set those values because there are no placeholders for these values in the `values.yaml` file for the chart. The "Helm parameters" tab which appears if you create an Application from our Helm repo/chart, only includes values which appear in values.yaml - by design of Argo CD.

A workaround: at first, do not enable any Pixie related settings for the chart bundle, then after New Relic is installed and synced, run `argocd app edit newrelic-bundle` and add the Pixie config to the app's parameters list manually, then sync again. Another workaround could be using `argocd` to create the entire application with the CLI, including all the required parameters.

Here is the text which must be added to the `newrelic-bundle` Application manually after install time, showing the required values for Pixie:

`argocd edit app newrelic-bundle`

```
    - name: pixie-chart.clusterName
      value: mycluster
    - name: pixie-chart.deployKey
      value: MY_PIXIE_DEPLOY_KEY
    - name: newrelic-pixie.apiKey
      value: MY_PIXIE_API_KEY
    - name: newrelic-pixie.enabled
      value: "true"
    - name: pixie-chart.enabled
      value: "true"
```

Optionally, two settings which are required at install time for some environments:

```
    - name: pixie-chart.registry
      value: "1234.dkr.ecr.us-east-1.amazonaws.com"
    - name: pixie-chart.useEtcdOperator
      value: "false"
```

This change adds all these values to the `newrelic-pixie` and `pixie-chart` subcharts in the bundle config, to make it possible to set them at install time when adding a new Helm Application in Argo CD.

#### Checklist
- [x] Chart Version bumped


charts/nri-bundle/values.yaml
```diff
@@ -26,10 +26,20 @@ newrelic-logging:
...
   # newrelic-pixie.enabled -- Install the [`newrelic-pixie`](https://github.com/newrelic/helm-charts/tree/master/charts/newrelic-pixie)
   enabled: false
+  # newrelic-pixie.apiKey -- The [Pixie API key](https://docs.pixielabs.ai/reference/admin/api-keys/#create-an-api-key), generated by the New Relic guided installer
+  apiKey: ""
 
 pixie-chart:
   # pixie-chart.enabled -- Install the [`pixie-chart` chart](https://docs.pixielabs.ai/installing-pixie/install-schemes/helm/#3.-deploy)
   enabled: false
+  # pixie-chart.clusterName -- The cluster name for the Kubernetes cluster. Set this to match the value in `global.cluster`.
+  clusterName: ""
```
roobre, line 36: Asking the user to set two properties to the same value does not seem like the best possible UX. Wouldn't it be better if the pixie chart read `global.cluster` directly?

maxlemieux: Thanks @roobre I will look into that and update this branch.

maxlemieux: Based on [Helm docs](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/), it seems like `pixie-operator-chart/templates/04_vizier.yaml` template can use this logic:

```
{{- if .Values.global.cluster }}
clusterName: {{ .Values.global.cluster }}
{{- else if .Values.clusterName }}
clusterName: {{ .Values.clusterName }}
{{- end }}
```

I'll follow up with Pixie about that part.

maxlemieux: https://github.com/pixie-io/pixie/pull/674

roobre (commented):
Thanks @maxlemieux for taking care of the changes!
Can you re-run the `helm-docs` command to regenerate the README with the new changes?

charts/nri-bundle/README.md
```diff
@@ -177,18 +177,20 @@ honors global options as described below.
 | global.serviceAccount.name | string | `nil` | Change the name of the service account. This is honored if you disable on this chart the creation of the service account so you can use your own |
 | global.tolerations | list | `[]` | Sets pod's tolerations to node taints |
 | global.verboseLog | bool | false | Sets the debug logs to this integration or all integrations if it is set globally |
-| kube-state-metrics.collectors | object | See [`values.yaml`](values.yaml) of the kube-state-metric chart | Collectors configuration of kube-state-metric |
```
maxlemieux, line 180: This does not seem to appear in the current `values.yaml` file, so the `helm docs` command didn't generate it. Let me know if we should add it to `values.yaml`.

charts/nri-bundle/README.md
```diff
@@ -16,7 +16,7 @@ here is a list of components that this chart installs and where you can find mor
 |------------------------------|-----------------------|-------------|
 | [newrelic-infrastructure](https://github.com/newrelic/nri-kubernetes/tree/main/charts/newrelic-infrastructure) | Yes | Sends metrics about nodes, cluster objects (e.g. Deployments, Pods), and the control plane to New Relic. |
 | [nri-metadata-injection](https://github.com/newrelic/k8s-metadata-injection/tree/main/charts/nri-metadata-injection) | Yes | Enriches New Relic-instrumented applications (APM) with Kubernetes information. |
-| [kube-state-metrics](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-state-metrics) | | Required for `newrelic-infrastructure` to gather cluster-level metrics. |
+| [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics/tree/master/charts/kube-state-metrics) | | Required for `newrelic-infrastructure` to gather cluster-level metrics. |
```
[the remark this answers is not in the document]
maxlemieux: I think the issue here is that this is automatically generated but shows an outdated URL. Let me know if we can fix this somewhere else.

PR #1016, opened 2023-01-13
Allow installing KSM v2.6.0 to continue support k8s < v1.23

vuqtran88:
Add the kube-state-metrics-v26.enabled flag to the nri-bundle chart to allow installing KSM v2.6.0 to continue support k8s < v1.23.

Notes: Even if this PR is reviewed and approved, the merging of this PR into main will be happening around the same time as updating the install commands from New Relic UI. This is to reduce impact of installation error for customers who are using the guided install on k8s < v1.23


charts/nri-bundle/values.yaml
```diff
@@ -15,6 +15,11 @@ kube-state-metrics:
   # This is mandatory if `infrastructure.enabled` is set to `true` and the user does not provide its own instance of KSM version >=1.8 and <=2.0
   enabled: false
 
+kube-state-metrics-v26:
+  # kube-state-metrics-v26.enabled -- Install the [`kube-state-metrics` chart](https://github.com/kubernetes/kube-state-metrics/tree/master/charts/kube-state-metrics) from the stable helm charts repository.
+  # This is mandatory if `infrastructure.enabled` is set to `true` and the user does not provide its own instance of KSM version >=1.8 and <=2.0
```
sigilioso, line 20: As I see it, including some insight regarding when this or `kibe-state-metrics.enabled` should be used would be useful as it would be updated as well in the README chart docs.

sigilioso (commented):
God job! 👍 

I left a comment to suggest an improvement in the values comment. 

On the other hand, I think you need to update the REDME file by using the [helm-docs](https://github.com/norwoodj/helm-docs) command.

If you think it is useful, you may include as well a new test-case in the [check-bundle-dependencies test](https://github.com/newrelic/helm-charts/blob/master/scripts/check-bundle-dependencies.sh)

charts/nri-bundle/Chart.yaml
```diff
@@ -16,7 +16,7 @@ sources:
   - https://github.com/newrelic/helm-charts/tree/master/charts/newrelic-pixie
   - https://github.com/newrelic/newrelic-infra-operator/tree/master/charts/newrelic-infra-operator
 
-version: 5.0.2
+version: 5.1.0
```
roobre, line 19: What is going to be the new default KSM? If newest KSM is the default, I think we should bump the major.

charts/nri-bundle/Chart.yaml
```diff
@@ -45,13 +45,21 @@ dependencies:
...
+    version: 4.25.0
     condition: ksm.enabled,kube-state-metrics.enabled
     repository: https://prometheus-community.github.io/helm-charts
+    alias: kube-state-metrics
+
+  - name: kube-state-metrics
+    version: 4.23.0
+    condition: ksm_v26.enabled,kube-state-metrics-v26.enabled
+    repository: https://prometheus-community.github.io/helm-charts
+    alias: kube-state-metrics-v26
```
roobre, line 57: It seems like we use a condition named ksm_v26 to enable KSM 4.23. Is this correct? Moreover the other KSM is not 4.26 either, but 4.25.
I would recommend to use the full chart name and number, e.g. `kube-state-metrics-v4-23` for both the alias and the dependency.

roobre, line 57: Have we checked if using an alias changes the KSM `app.kubernetes.io/name` label? If these labels change we should update the labelSelector `nri-kubernetes` uses to find KSM.

This is hard to do from `values.yaml`. I would suggest to not use aliases to avoid this, and have a safety check in `NOTES.txt` preventing the user from setting both dependencies at the same time.

charts/nri-bundle/Chart.yaml
```diff
@@ -45,13 +45,21 @@ dependencies:
...
   - name: kube-state-metrics
-    version: 4.23.0
+    version: 4.25.0
     condition: ksm.enabled,kube-state-metrics.enabled
     repository: https://prometheus-community.github.io/helm-charts
+    alias: kube-state-metrics
+
+  - name: kube-state-metrics
+    version: 4.23.0
+    condition: ksm_v26.enabled,kube-state-metrics-v26.enabled
```
roobre, line 55: The original `kube-state-metrics` dependnecy has two conditions for backwards compatibility reasons. `ksm.enabled` is deprecated, so I wouldn't propagate that pattern to new dependencies. I'd suggest to use the chart name (or alias) plus `.enabled` as the only condition.

roobre (changes requested):
Left some comments, I think there are some typos and naming nits that should be addressed.

Perhaps the most important thing, I'd suggest to not use `alias` for dependencies. I think (although I haven't checked, please correct me if I'm wrong) that using `alias` will cause the name of the KSM labels to change, preventing the integration from finding it in the cluster.

charts/nri-bundle/Chart.yaml
kang-makes: Here you have an example: https://github.com/newrelic/nri-kubernetes/blob/main/charts/newrelic-infrastructure/templates/NOTES.txt#L125-L131

kang-makes: I disagree with both of the approaches as the user should then know which version of KSM is needed for their cluster. 

I would propose to alias it to something like `kube-state-metric-for-kubernetes-pre-1-26`. Maybe `kube-state-metric-pre-1-26` but I prefer the long one because it is self-explanatory.

charts/nri-bundle/Chart.yaml
```diff
@@ -45,13 +45,21 @@ dependencies:
     version: 1.0.0
 
   - name: kube-state-metrics
-    version: 4.23.0
+    version: 4.25.0
```
paologallinaharbur, line 48: If we are taking into consideration upgrading a major would not make sense as well to drop the support for HPA metrics on old clusters <1.23?
I fear that we are making our life and configuration more complex for the HPA sample that most of the customers do not use.
In case they really need that they can always downgrade KSM and we can document that option on docs or even changing the image with the guided install. 

I am not sure we want to split the config to support EOLed version of K8s pinning a specific KSM version (with growing vulnerabilities over time).

paologallinaharbur: Of course I am fine with having this additional complexity, but I fear it is going to be a bad precedent. We cannot assume we are always supporting old clusters if the tools we use do not.
```
1.22
Latest Release:1.22.12 (released: 2022-07-13)
End of Life: 2022-10-28
```

If they run an old cluster they can set an old KSM image. We could provide a warning suggesting that.

https://github.com/kubernetes/kube-state-metrics/issues/1946

vuqtran88: Hi @paologallinaharbur, thanks for raising a really good point. No matter how much I want simplicity, we still have to continue to support the HPA metrics for our existing customers after discussing about this with the team.

charts/nri-bundle/Chart.yaml
vuqtran88: I like the `kube-state-metrics-for-k8s-pre-1-23` naming suggestion.

charts/nri-bundle/Chart.yaml
vuqtran88: I think the newest KSM is the default. I will bump the major.

charts/nri-bundle/Chart.yaml
```diff
@@ -45,13 +45,21 @@ dependencies:
...
     condition: ksm.enabled,kube-state-metrics.enabled
     repository: https://prometheus-community.github.io/helm-charts
+    alias: kube-state-metrics
+
+  - name: kube-state-metrics
```
paologallinaharbur, line 53: [discussed offline, adding it here for tracking porposes]

I still do not see clearly why adding a new dep is better (security or configuration wise) with respect to changing the image tag:
```
helm upgrade --install newrelic-bundle newrelic/nri-bundle 
 --set kube-state-metrics.image.tag = "v1.26.0" [...]
```

vuqtran88:
Close this PR for now as the solution seems to have many concerns.

PR #1045, opened 2023-03-10
[newrelic-logging]: fix missing propagation of priorityClassName

woehrl01:
  - priortyClassName can be defined via global, but is not propgated to the client, resulting in the value not beeing set.

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/newrelic-logging/templates/daemonset-windows.yaml
+++ b/charts/newrelic-logging/templates/daemonset-windows.yaml
@@ -44,8 +44,8 @@ spec:
       imagePullSecrets:
 {{ toYaml $.Values.image.pullSecrets | indent 8 }}
       {{- end }}
-      {{- with include "newrelic.common.priorityClassName" . }}
-      priorityClassName: {{ . }}
+      {{- with include "newrelic.common.priorityClassName" $ }}
+      priorityClassName: {{ $. }}
       {{- end }}
       containers:
         - name: {{ template "newrelic-logging.name" $ }}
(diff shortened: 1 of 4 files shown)


charts/newrelic-logging/templates/daemonset-windows.yaml
```diff
@@ -43,6 +43,9 @@ spec:
       {{- if $.Values.image.pullSecrets }}
       imagePullSecrets:
 {{ toYaml $.Values.image.pullSecrets | indent 8 }}
+      {{- end }}
+      {{- with include "newrelic.common.priorityClassName" . }}
```
danybmx, line 47: Hi! This should be `{{- with include "newrelic.common.priorityClassName" $ }}` instead as for windows daemonset we're iterating over versions

danybmx (changes requested):
Hi @woehrl01, thanks a lot for the contribution!
I've left a comment as I think this will fail on windows nodes.

charts/newrelic-logging/templates/daemonset-windows.yaml
woehrl01: Thanks, I'll add the changes in the next days

woehrl01: @danybmx Just pushed a new version, I also saw that there is already a priorityClassName set, but not from "commons" so I removed that one on windows.

PR #1075, opened 2023-05-04
[newrelic-logging] feat: upgrade logging helm chart to v1.16.0

jcsobrino:
- Upgraded logging helm chart to v1.16.0. 

#### Checklist
- [x] Chart Version bumped


charts/newrelic-logging/k8s/new-relic-fluent-plugin.yml
```diff
@@ -41,7 +41,7 @@ spec:
               value: "docker"
             - name: PATH
               value: "/var/log/containers/*.log"
-          image: newrelic/newrelic-fluentbit-output:1.8.1
```
luckslovez, line 44: Hi @jcsobrino! I think the approach here is to just update `Chart.yaml#appVersion` since that's what is used in the [`templates/daemonset.yaml`](https://github.com/newrelic/helm-charts/blob/master/charts/newrelic-logging/templates/daemonset.yaml#L53).
I am not sure of the purpose of this `k8s` folder, it does not seem to be frequently updated. Wdyt @danybmx ? Should we just delete this?

luckslovez: As discussed, despite planning on removing this, it does not harm to keep it updated. Thanks @jcsobrino !

PR #1100, opened 2023-06-01
set hostNetwork:false by default

dpacheconr:
The [newrelic-logging Helm chart](https://github.com/newrelic/helm-charts/tree/master/charts/newrelic-logging) is currently [using the hostNetwork:true setting](https://github.com/newrelic/helm-charts/blob/master/charts/newrelic-logging/templates/daemonset.yaml#L36)  when creating the DaemonSet 
The recommended value for this setting is false, unless strictly necessary.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204883406985385

### DIFF
--- a/charts/nri-bundle/Chart.lock
+++ b/charts/nri-bundle/Chart.lock
@@ -1,16 +1,16 @@
 dependencies:
 - name: newrelic-infrastructure
   repository: https://newrelic.github.io/nri-kubernetes
-  version: 3.21.0
+  version: 3.22.0
 - name: nri-prometheus
   repository: https://newrelic.github.io/nri-prometheus
   version: 2.1.17
 - name: newrelic-prometheus-agent
   repository: https://newrelic.github.io/newrelic-prometheus-configurator
-  version: 1.2.3
+  version: 1.3.0
 - name: nri-metadata-injection
   repository: https://newrelic.github.io/k8s-metadata-injection
-  version: 4.4.0
+  version: 4.7.0
 - name: newrelic-k8s-metrics-adapter
   repository: https://newrelic.github.io/newrelic-k8s-metrics-adapter
   version: 1.2.1
@@ -19,7 +19,7 @@ dependencies:
   version: 5.12.1
 - name: nri-kube-events
   repository: https://newrelic.github.io/nri-kube-events
-  version: 3.1.2
+  version: 3.2.0
 - name: newrelic-logging
   repository: https://newrelic.github.io/helm-charts
   version: 1.18.1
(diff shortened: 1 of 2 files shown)


charts/newrelic-logging/templates/_host_network.tpl
```diff
@@ -0,0 +1,20 @@
+{{/* Returns whether the we should run with hostNetwork: true based on the user configuration. */}}
+{{- define "newrelic-logging.hostNetwork" -}}
```
luckslovez, line 2: Hi! I can not see any usage of these two definitions. Do we really need this file at all?

dpacheconr: @luckslovez if the tests pass without it, it can be removed.
this file was just copied from other NR helm charts and updated accordingly 
I will remove it

immae1:
Hi @dpacheconr, @luckslovez - I also need to set hostNetwork=false as right now the default helmchart deployment of the logger is getting DNS/log transfer issues in my clusters when the hardcoded hostNetwork=true.  
When I deploy the `nri-bundle' without modification in the `newrelic-logging' chart I see the following DNS issues:
```log
[ warn] [net] getaddrinfo(host='kubernetes.default.svc.cluster.local', err=-3): Temporary failure in name resolution
[error] [filter:kubernetes:kubernetes.0] kubelet upstream connection error
```
And very often I see that the logs could not be transferred to my NewRelic instance:
```log
[2023/07/27 10:58:25] [ warn] [input] tail.0 paused (mem buf overlimit)
[2023/07/27 10:58:25] [ info] [input] pausing tail.0
time="2023-07-27T10:58:30Z" level=info msg="Retryable error received. Will retry to send the logs (if there are attempts remaining, check Retry_Limit option)" error="Post \"https://log-api.newrelic.com/log/v1\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"
[2023/07/27 10:58:30] [error] [engine] chunk '1-1690421202.734631846.flb' cannot be retried: task_id=69, input=tail.0 > output=newrelic.0
[2023/07/27 10:58:30] [ info] [input] tail.0 resume (mem buf overlimit)
```

I manually tested, set the flag to false and then all the above errors disappeared and I see almost real-time logs streaming from my cluster to NewRelic. 
So 3 questions from me:

1. is it safe to set hostNetwork=false? 
2. When will this PR be merged/released into a new Helm-Chart version? 
3. Is it enough to comment here or should I create a new issue/bug report? 

Thanks in advance and best regards 
Immae1

PR #1137, opened 2023-08-02
[synthetics-job-manager] add command option for cronjob

ac427:
#### Is this a new chart
yes
This PR adds value to run command for cronjob on condition. This will help to run overwrite default commands. 
Example : istio sidecar requires ` http://127.0.0.1:15020/quitquitquit ` as final step

  - fixes #

#### Checklist
- [x] Chart Version bumped


Philip-R-Beckwith (changes requested):
Can you add an example of how a customer would use this? 
and a proof that this new feature would work?

PR #1151, opened 2023-09-22
pixie-non-operator 

dpacheconr:
Allow users to choose to install pixie as operator or non-operator deployment methods as detailed here https://docs.px.dev/installing-pixie/install-schemes/helm/

Install helm command will look like this
function ver { printf "%03d%03d" $(echo "$1" | tr '.' ' '); } && \
K8S_VERSION=$(kubectl version --short 2>&1 | grep 'Server Version' | awk -F' v' '{ print $2; }' | awk -F. '{ print $1"."$2; }') && \
if [[ $(ver $K8S_VERSION) -lt $(ver "1.25") ]]; then KSM_IMAGE_VERSION="v2.6.0"; else KSM_IMAGE_VERSION="v2.7.0"; fi && \
helm repo add newrelic https://helm-charts.newrelic.com && helm repo update && \
kubectl create namespace newrelic ; helm upgrade --install newrelic-bundle ./newrelic/nri-bundle \
 --set global.licenseKey=XXX \
 --set global.cluster=CLUSTER_NAME \
 --namespace=newrelic \
 --set newrelic-infrastructure.privileged=true \
 --set global.lowDataMode=true \
 --set kube-state-metrics.image.tag=${KSM_IMAGE_VERSION} \
 --set kube-state-metrics.enabled=true \
 --set kubeEvents.enabled=true \
 --set newrelic-prometheus-agent.enabled=true \
 --set newrelic-prometheus-agent.lowDataMode=true \
 --set newrelic-prometheus-agent.config.kubernetes.integrations_filter.enabled=false \
 --set logging.enabled=true \
 --set newrelic-logging.lowDataMode=true \
 --set newrelic-pixie.enabled=true \
 --set newrelic-pixie.apiKey=px-api-XXXX \
 --set pixie-non-operator-chart.enabled=true \
 --set pixie-non-operator-chart.deployKey=px-dep-XXX \
 --set pixie-non-operator-chart.clusterName=CLUSTER_NAME

### DIFF
--- a/charts/nri-bundle/Chart.lock
+++ b/charts/nri-bundle/Chart.lock
@@ -16,7 +16,7 @@ dependencies:
   version: 1.13.6
 - name: kube-state-metrics
   repository: https://prometheus-community.github.io/helm-charts
-  version: 5.26.0
+  version: 5.30.1
 - name: nri-kube-events
   repository: https://newrelic.github.io/nri-kube-events
   version: 3.11.7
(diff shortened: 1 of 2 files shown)


philomory:
I would love to see this (actually, I think my favorite option would be if you could install the _Vizier_ operator, for managing Pixie, surfacing deployment status, etc, but not have to have _that_ managed by the Operator Lifecycle Manager; but since the Pixie project doesn't currently offer a way to do that, I'd choose the no-operator-at-all option every time)

PR #1156, opened 2023-10-13
ci: Automate weekly release of bundle

juanjjaramillo:
  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -1,0 +1,79 @@
+name: Post Release Tasks
+
+on:
+  release:
+    types: [released]
+
+env:
+  ORIGINAL_REPO_NAME: ${{ github.event.repository.full_name }}
+  RELEASE_PATH: charts/nri-bundle
+  RELEASE_FILE: RELEASE.md
+  SLACK_FILE: RELEASE.slack
+
+jobs:
+  release-notes:
+    if: startsWith(github.event.release.tag_name, 'nri-bundle-')
+    name: Publish release notes
+    runs-on: ubuntu-latest
+    outputs:
+      announcement: ${{ steps.slack-announcement.outputs.announcement }}
+    steps:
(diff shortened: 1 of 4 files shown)


svetlanabrennan:
Looks good to me overall. Just one question:

Shouldn't we create a [release workflow](https://github.com/newrelic/helm-charts/blob/master/.github/workflows/release.yml) just for our k8s-agent team so it doesn't interfere with other teams' release process. And that way we only get release failures for [our specific release failures](https://github.com/newrelic/helm-charts/pull/1156/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R107-R117)?

svetlanabrennan (approved):

juanjjaramillo:
> Looks good to me overall. Just one question:
> 
> Shouldn't we create a [release workflow](https://github.com/newrelic/helm-charts/blob/master/.github/workflows/release.yml) just for our k8s-agent team so it doesn't interfere with other teams' release process. And that way we only get release failures for [our specific release failures](https://github.com/newrelic/helm-charts/pull/1156/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R107-R117)?

That is a very good point. I looked at creating a second file, but `helm/chart-releaser-action` does not have a clean way to specify that `nri-bundle` should be excluded. What is feasible to do is to run our steps only when we have the `schedule` event:
```
if: github.event_name == 'schedule'
```
To avoid affecting other teams I also restored the `push` event, so other teams can create a chart as soon as they merge changes to `master`.

These two changes accomplish the following:
- Other teams keep releasing their charts on a push to `master`, so they do not get affected by our changes
- The only chart that will be released on schedule will be `nri-bundle` since it is the only one on a schedule
- We only get notified of `nri-bundle` failures

PR #1175, opened 2023-10-18
[super-agent] First iteration of the flux chart [NR-168649]

kang-makes:
#### Is this a new chart
Yes

This is the umbrella chart that installs Flux and bootstraps the Super Agent on top of Kubernetes.

#### Checklist
- [X] Chart Version bumped


charts/super-agent/values.yaml
```diff
@@ -0,0 +1,181 @@
...
+      remediation:
+        retries: 3  # Retries is the number of retries that should be attempted on failures before givin up. An uninstall is performed between each attempt.
+        strategy: rollback  # rollback or uninstall.
+
+# -- Values for the Flux chat. Ref.: https://github.com/fluxcd-community/helm-charts/blob/flux2-2.10.2/charts/flux2/values.yaml
+# @default -- See `values.yaml`
+flux2:
+  # -- Enable or disable FluxCD completely. New Relic' Super Agent need Flux to work, but the user can use an already
+  # existing Flux deployment. With that use case, the use can disable Flux and use this chart to only install the CRs
+  # to deploy the Super Agent.
```
sigilioso, line 90:
```suggestion
  # -- Enable or disable FluxCD instasllation. New Relic' Super Agent need Flux to work, but the user can use an already
  # existing Flux deployment. When disabled, this chart can be used to install only the CRs needed to deploy the Super Agent.
```

sigilioso: It is probably out of the scope of this PR and maybe it should be somewhere else but, should we describe the 'already existing Flux' requirements to deploy the Super Agent?

Maybe this values file is can be considered self-documentation, if that is the case we could say it explicitly.

charts/super-agent/values.yaml
```diff
@@ -0,0 +1,181 @@
...
+    chart: super-agent-deployment
+    # -- The Helm chart of the super-agent. This values is meant to be changed only on air-gapped environments or for
+    # development/testing purposes.
+    #
+    # TODO: Point renovatebot here.
+    version: "0.0.0-beta"
+
+    # -- Set values to the super agent helm release directly from this `values.yaml` file. Refer to https://fluxcd.io/flux/components/helm/helmreleases/#values-overrides
+    # @default -- no values
+    values:
```
gsanchezgavier, line 38: would make sense to add the license , perhaps create the secret and configurate it here by default ?

charts/super-agent/README.md
```diff
@@ -0,0 +1,81 @@
...
+as TODO:
+ * `licenseKey`: comes from the common library but it would not be needed as it is an ingestion API Key, not a REST one.
+ * `rbac`: the is a placeholder for RBAC that simply list pods. this has to be narrowed to the use case of this agent.
+ * `customAttributes`: decorate everything with this custom attributes, maybe as they come from opamp.
+ * `proxy`, `nrStaging` and `fedramp` support on the meta agent. This could be made from the chart itself changing the opamp endpoint.
+ * `verboseLog`: legacy toggle to enable verbosity.
+
+Other TODOs:
+ * Add probes for liveness and readiness.
+ * See if a service is needed or not. Enable it or delete the file.
```
paologallinaharbur, line 49: I am not sure these still apply

charts/super-agent/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+Bootstraps New Relic' Super Agent
+
+# Helm installation
+
+You can install this chart using directly this Helm repository:
+
+```shell
+helm repo add newrelic https://newrelic.github.io/helm-charts
+helm upgrade --install newrelic/CHART-TEMPLATE -f your-custom-values.yaml
```
paologallinaharbur, line 15: newrelic/CHART-TEMPLATE?

charts/super-agent/README.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+## Values managed globally
+
+This chart implements the [New Relic's common Helm library](https://github.com/newrelic/helm-charts/tree/master/library/common-library) which
+means that it honors a wide range of defaults and globals common to most New Relic Helm charts.
+
+Options that can be defined globally include `affinity`, `nodeSelector`, `tolerations`, `proxy` and others. The full list can be found at
+[user's guide of the common library](https://github.com/newrelic/helm-charts/blob/master/library/common-library/README.md).
```
paologallinaharbur, line 24: isn't this a bit misleading? Since we are including Flux how are we making sure the features are actually implemented?

charts/super-agent/values.yaml
```diff
@@ -0,0 +1,181 @@
+helm:
+  # -- Enable the installation of the CRs so FluxCD deploy the Super Agent is deployed. This an advanced/debug flag. It
+  # should be always be true unless you know what you are going. 
+  create: true
```
paologallinaharbur, line 4: if it is for debugging only, should we expose it?

charts/super-agent/values.yaml
paologallinaharbur: I agree, it would be nice to have a working config by default

charts/super-agent/values.yaml
```diff
@@ -0,0 +1,181 @@
...
+      #   name: prod-tls-values
+      #   valuesKey: crt
+      #   targetPath: tls.crt
+
+    # -- Change the behavior of the operator while installing the chart for the first time. This should only be changed
+    # by advanced users that know what they are doing. Exposes the remediations that the operator is going to try
+    # before give up installing the chart in case it hits an error.
+    # @default -- See `values.yaml`
+    install:
+      crds: CreateReplace  # New CRDs are created, existing CRDs are updated (replaced) but not deleted.
```
paologallinaharbur, line 68: I think we are not creating crds in the SuperAgent chart, I would simplify this removing this option

charts/super-agent/values.yaml
```diff
@@ -0,0 +1,181 @@
...
+  # Kustomize instead of using templating systems like Helm.
+  # 
+  # New Relic' Super Agent will use this controller by creating `Kustomization` CRs based in the configuration stored
+  # on OpAmp.
+  # @default -- Enabled
+  kustomizeController:
+    create: true
+    annotations:
+      prometheus.io/port: "8080"
+      prometheus.io/scrape: "false"  # TODO: Should we create a FluxCD entity to monitor flux itself!?
```
paologallinaharbur, line 131: it is not used yet I would not deploy it for now

charts/super-agent/values.yaml
```diff
@@ -0,0 +1,181 @@
+helm:
```
paologallinaharbur, line 1: why helm and not super-agent?

charts/super-agent/README.md
kang-makes: This part is removed following the review of the chat downstream. Removed on [#d2e49c1](https://github.com/newrelic/helm-charts/pull/1175/commits/d2e49c17d289c548346804a27429633fbe279b38)

charts/super-agent/README.md
kang-makes: This part is removed following the review of the chat downstream. Removed on [#d2e49c1](https://github.com/newrelic/helm-charts/pull/1175/commits/d2e49c17d289c548346804a27429633fbe279b38)

charts/super-agent/README.md
```diff
@@ -0,0 +1,81 @@
...
+low data modes or any other quirk that it could have.
+
+### TODOs
+There are values that should be planned at some point of removed from the `values.yaml`. I leave them here documented
+as TODO:
+ * `licenseKey`: comes from the common library but it would not be needed as it is an ingestion API Key, not a REST one.
+ * `rbac`: the is a placeholder for RBAC that simply list pods. this has to be narrowed to the use case of this agent.
+ * `customAttributes`: decorate everything with this custom attributes, maybe as they come from opamp.
+ * `proxy`, `nrStaging` and `fedramp` support on the meta agent. This could be made from the chart itself changing the opamp endpoint.
+ * `verboseLog`: legacy toggle to enable verbosity.
```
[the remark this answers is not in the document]
kang-makes: This part is removed following the review of the chat downstream. Removed on [#d2e49c1](https://github.com/newrelic/helm-charts/pull/1175/commits/d2e49c17d289c548346804a27429633fbe279b38)

charts/super-agent/values.yaml
kang-makes: I agree, that we have to also pin the version and capabilities. But I think it is too soon as everything can change. We can do it at the release of the MVP.

If you have an idea of something that we can start defining right now that can be written in stone?

charts/super-agent/values.yaml
kang-makes: I added it as a TODO on [#d2e49c1](https://github.com/newrelic/helm-charts/pull/1175/commits/d2e49c17d289c548346804a27429633fbe279b38)

We need to make it work downstream before adding it here.

charts/super-agent/README.md
kang-makes: I rephrased it on [#d2e49c1](https://github.com/newrelic/helm-charts/pull/1175/commits/d2e49c17d289c548346804a27429633fbe279b38)

Let me know what you think.

charts/super-agent/values.yaml
```diff
@@ -0,0 +1,181 @@
...
+  # 
+  # On the other hand, user might want to leverage having FluxCD installed for their own purposes.
+  # @default -- Disabled
+  imageAutomationController:
+    create: false
+
+  # -- The image reflector controller scans image repositories and reflects the image metadata in Kubernetes resources
+  # ready to be used by other controllers.
+  #
+  # From New Relic, all releases are managed via OpAmp and there is no need to touch user's repositories.
```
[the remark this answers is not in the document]
kang-makes: I will keep only one conversation here so we do not cross messages :)

https://github.com/newrelic/helm-charts/pull/1175#discussion_r1364991816

charts/super-agent/values.yaml
```diff
@@ -0,0 +1,181 @@
...
+  kustomizeController:
+    create: true
+    annotations:
+      prometheus.io/port: "8080"
+      prometheus.io/scrape: "false"  # TODO: Should we create a FluxCD entity to monitor flux itself!?
+
+  # -- The image automation controller updates YAML files based on the latest images scanned by image reflector
+  # controller, and commits the changes to a given Git repository.
+  #
+  # From New Relic, all releases are managed via OpAmp and there is no need to touch user's repositories.
```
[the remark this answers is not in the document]
kang-makes: I have to set `create` below to false. From a user perspective, it is better to have the rationale for why it is not being created.

I would prefer to rephrase it rather than remove it.

The other alternative would be to create a custom chart for Flux which is a possibility that I did not consider as it would add way more maintainability effort.

charts/super-agent/values.yaml
kang-makes: I cannot think of many reasons to disable it but I am thinking of things like the guided install that could leverage it somehow on templating stages.

I prefer to document everything and let the user footgun itself rather than having undocumented things that we would need. At least, at the stage that we are right now, we need all the documentation for ourselves.

charts/super-agent/values.yaml
kang-makes: This is not for the superagent but for the things that Flux is going to deploy. If the Super Agent deploys something that needs a CRD (let's say an absurd example like KEDA or the New Relic's operator for alerts) that will need CRDs and will need to upgrade them.

charts/super-agent/values.yaml
kang-makes: Fair. Removed on [#244a4b4](https://github.com/newrelic/helm-charts/pull/1175/commits/244a4b4c444f4b09068dd3d939dcc099571153f6)

charts/super-agent/values.yaml
kang-makes: The way I see if a values object makes sense or is redundant is to ask myself what would the values look like if we set an umbrella chart. Let's create an example. Imagine this chart goes inside the `nri-bundle`. The values on the bundle with your suggestion would look like this:
```yaml
super-agent:
  super-agent:
    create: true
```

On the other hand, with `helm` it would look like:
```yaml
super-agent:
  helm:
    create: true
```
With this example, I can say that the object `helm` is not the agent but the helm release, and it makes sense to call it `helm` :)

paologallinaharbur: Really nice explanation 😃 

I guess having it "at level 0" (without `helm:`) it does not make sense. Then I agree that helm makes sense!

charts/super-agent/values.yaml
paologallinaharbur: fair enough

charts/super-agent/values.yaml
paologallinaharbur: I am not following, isn't this being used in the helm-release (charts/super-agent/templates/helm-release.yaml) of the superAgent? Therefore it is going to be useful only if at some point we have CRDs as part of the SA, right?

charts/super-agent/README.md
paologallinaharbur: I like the disclaimer regarding flux! From my point of view it is enough 😄

charts/super-agent/values.yaml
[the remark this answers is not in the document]
gsanchezgavier: We could state that SA only use the helm/source controller to deploy sub agents and the other Flux components are disable by default. , I would put all disable configs together under the same comment

charts/super-agent/values.yaml
```diff
@@ -0,0 +1,181 @@
...
+  # existing Flux deployment. With that use case, the use can disable Flux and use this chart to only install the CRs
+  # to deploy the Super Agent.
+  enabled: true
+
+  # -- The installation of the CRDs is managed by the chart itself.
+  installCRDs: true
+  # -- This is the domain name of the cluster.
+  clusterDomain: cluster.local
+
+  # -- Source controller provides a way to fetch artifacts to the rest of controllers. The source API (which reference
```
marcsanmi, line 99: Is it a good practice in Helm to add comments about components used in other charts? I mean, describing Flux controllers should be the responsibility of our `values.yaml` or the Flux2 one? I was thinking of a way to simplify the values.

marcsanmi (commented):
Great job! 👍  
I just left some nits and questions.

charts/super-agent/values.yaml
kang-makes: You are completely right. I don't know what I was thinking.

Removed on [#5573b5b](https://github.com/newrelic/helm-charts/pull/1175/commits/5573b5b638dce8efdb24dae685bdaab622774bed)

charts/super-agent/values.yaml
kang-makes: I defined a simple set of features that might be useful for us and for clients... but the whole values file of the upstream chart is not documented. See:
 * https://github.com/newrelic/helm-charts/pull/1175/files#diff-3bd49393f47bdfcb35b90c33237f384109da730bc892fa317ba8db58d29d7975R84-R85
 * https://github.com/newrelic/helm-charts/pull/1175/files#diff-20bc91206f6699b12924a1284a7942be082f0b582651eedb8f73449236b34c81R23-R26

charts/super-agent/templates/uninstall-job.yaml
```diff
@@ -0,0 +1,49 @@
...
+  template:
+    spec:
+      restartPolicy: Never
+      serviceAccountName: helm-controller  # Hardcoded on flux2 chart.
+      containers:
+        - name: apply-crs
+          image: bitnami/kubectl
+          command:
+            - kubectl
+            - apply
```
gsanchezgavier, line 39: shouldn't be `delete`?

charts/super-agent/templates/_helm-release.yaml
```diff
@@ -0,0 +1,37 @@
+{{- define "newrelic-super-agent.helm.release" -}}
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
+kind: HelmRelease
+metadata:
+  labels:
+    {{- include "newrelic.common.labels" . | nindent 4 }}
+  name: {{ include "newrelic.common.naming.fullname" . }}
```
gsanchezgavier, line 7: if `super-agent` to install the chart this name ends up also having `super-agent` which ends up in a really wired bug. 
I think that the common library internally removes the name duplication and this is why both names are the same.

kang-makes: Solved on [#96cd608](https://github.com/newrelic/helm-charts/pull/1175/commits/96cd608894994e249a94c104964dd6834031b8d5)

charts/super-agent/templates/uninstall-job.yaml
kang-makes: 🤦 Solved on [#64f34cb](https://github.com/newrelic/helm-charts/pull/1175/commits/64f34cb80ccec822b72448d6e0002a38e800bf7d)

charts/super-agent/templates/uninstall-job.yaml
```diff
@@ -0,0 +1,49 @@
...
+  template:
+    spec:
+      restartPolicy: Never
+      serviceAccountName: helm-controller  # Hardcoded on flux2 chart.
+      containers:
+        - name: delete-crs
+          image: bitnami/kubectl
+          command:
+            - kubectl
+            - delete
```
paologallinaharbur, line 39: is there a race condition after deleting this? Should we wait  some seconds before cleaning everything else? (maybe a sidecar with a 60s sleep
)

kang-makes: Deleting an object in Kuberentes is blocking. It will not be deleted until the helm release is removed.

Well... Actually and being completely specific, is blocking until the object is removed from the API Server but Flux controller adds a finalizer to the object and blocks it until the release is uninstalled.

.github/workflows/lint_test_charts.yaml
```diff
@@ -83,6 +83,11 @@ jobs:
       - name: Create test ns
         run: kubectl create ns ct
 
+      - name: Test Flux CD on the kind instance that is created by this CI/CD
```
paologallinaharbur, line 86: If this is not for debugging purposes I'd either remove it or add a comment explaining why it is needed and for which chart if we cannot test it just for the SA one

charts/super-agent/templates/uninstall-job.yaml
```diff
@@ -0,0 +1,49 @@
...
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+    helm.sh/hook-weight: "10000"
+  name: {{ $configMapName }}
+data:
+  helm-release.yaml: |
+    {{- include "newrelic-super-agent.helm.release" . | nindent 4 }}
+  helm-repository.yaml: |
+    {{- include "newrelic-super-agent.helm.repository" . | nindent 4 }}
+---
+apiVersion: batch/v1
```
paologallinaharbur, line 19: I would place either installer +deleter in the same file or all of them separated

charts/super-agent/templates/install-job.yaml
```diff
@@ -0,0 +1,49 @@
+{{- if .Values.helm.create -}}
```
paologallinaharbur, line 1: I would add some docs either here or in the readme explaining why we have a job

charts/super-agent/templates/uninstall-job.yaml
paologallinaharbur: forget this, my bad

charts/super-agent/templates/uninstall-job.yaml
```diff
@@ -0,0 +1,49 @@
...
+{{- $installJobName := include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "uninstall-job" ) -}}
+{{- $configMapName := include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "manifests" ) -}}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  annotations:
+    helm.sh/hook: pre-delete
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+    helm.sh/hook-weight: "10000"
```
paologallinaharbur, line 11: I do not think the weight is needed, if it is the same Helm should still respect the classical order

charts/super-agent/templates/uninstall-job.yaml
```diff
@@ -0,0 +1,49 @@
+{{- if .Values.helm.create -}}
+{{- $installJobName := include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "uninstall-job" ) -}}
+{{- $configMapName := include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "manifests" ) -}}
```
paologallinaharbur, line 3: job-manifests instead of manifests?

paologallinaharbur (approved):
Thanks for taking care of this! 😄

.github/workflows/lint_test_charts.yaml
kang-makes: This is the commit that introduced this. It is a test.

I was unable to know why the CI was failing tests because the chart testing action does not give you logs of failing pre-install hooks.

Screenshot 2023-10-25 at 10 30 31

charts/super-agent/values.yaml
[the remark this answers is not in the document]
kang-makes: Fair. Done on [#7d0f2f0](https://github.com/newrelic/helm-charts/pull/1175/commits/7d0f2f0556057c63203ac9b721dad8ca91b9b3b6)

charts/super-agent/templates/uninstall-job.yaml
kang-makes: Done on [#b5da77e](https://github.com/newrelic/helm-charts/pull/1175/commits/b5da77e706e5cce2c1b77e36f72063bd23c1e715)

charts/super-agent/templates/uninstall-job.yaml
kang-makes: Done on [#b5da77e](https://github.com/newrelic/helm-charts/pull/1175/commits/b5da77e706e5cce2c1b77e36f72063bd23c1e715)

charts/super-agent/templates/install-job.yaml
kang-makes: Done on [#5157357](https://github.com/newrelic/helm-charts/pull/1175/commits/51573576d221c3e4ea5e8ba0278965e4860cd5ed)

charts/super-agent/templates/uninstall-job.yaml
kang-makes: I reopen it because you might have a different view of this.

I removed it on the commit above and it started to fail all the tests.
Screenshot 2023-10-25 at 16 50 12

I readded the weight on the t commit ([#e7cfe43](https://github.com/newrelic/helm-charts/pull/1175/commits/e7cfe435e02f6e76258e114a9bff65fd20d2406b)) and it started to pass the tests again.

I think there is another race condition because of hook defaults.

[Deletion policies](https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies) default to `before-hook-creation` so the job and the configMap would keep dangling on the cluster and possibly leaking data as the values are contained on the `HelmRelease` object, hence my policy of delete also on `hook-succeeded`.

But it seems that `hook-succeeded` creates the configMap and deletes it as fast as it is created if you do not set a `weight`.

I am still investigating this tho. Let me know what you think.

charts/super-agent/templates/job-manifests.yaml
```diff
@@ -0,0 +1,34 @@
...
+    helm.sh/hook: pre-delete,post-install,post-upgrade
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+    helm.sh/hook-weight: "10000"
+  name: {{ $configMapName }}
+  namespace: {{ .Release.Namespace }}
+data:
+{{- /*
+Let's start with some background that allow to understand why configmap has manifests on it.
+
+Helm behaves differently depending on where you.
```
paologallinaharbur, line 17: "Helm behaves differently depending on where you."?

kang-makes: You have seen nothing. And you have not seen the `push --force` I made. (Solved on [#2cd5d8c](https://github.com/newrelic/helm-charts/pull/1175/commits/2cd5d8cde4029707fdf1cccc58c411dae53f07a5))

paologallinaharbur (approved):

charts/super-agent/values.yaml
```diff
@@ -0,0 +1,164 @@
...
+    # development/testing purposes.
+    #
+    # TODO: Point renovatebot here.
+    version: "0.0.0-beta"
+
+    # -- Set values to the super agent helm release directly from this `values.yaml` file. Refer to https://fluxcd.io/flux/components/helm/helmreleases/#values-overrides
+    # @default -- no values
+    values:
+    # replicaCount: 5
+    # TODO: Wait for PR #1155 to be merged so we know how to send the `licenseKey` upstream to the SuperAgent.
```
paologallinaharbur, line 40: NIT This was merged already, could we change the snippet so that it is meaningful?

"  replicaCount: 5 " does nothing right now

kang-makes: Done in [#58a67dc](https://github.com/newrelic/helm-charts/pull/1175/commits/58a67dc386373475171f8110e7e4e67339bb1e51).

charts/super-agent/templates/_helm-release.yaml
```diff
@@ -0,0 +1,42 @@
...
+        name: newrelic
+        namespace: {{ .Release.Namespace }}
+      interval: {{ .Values.helm.repository.updateInterval }}
+  values:
+    fullnameOverride: super-agent
+    {{- with .Values.helm.release.values }}
+    {{- . | toYaml | nindent 4 }}
+    {{- end }}
+  {{- with .Values.helm.release.valuesFrom }}
+  valuesFrom:
```
gsanchezgavier, line 30: are changes on this values retriggering a redeploy of the SA? do you know how much time is needed to see the effects?

charts/super-agent/templates/uninstall-job.yaml
```diff
@@ -0,0 +1,37 @@
...
+{{- $configMapName := include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "job-manifests" ) -}}
+{{- /*
+To understand why this job installs manifests instead of using Helm, read the comment at `job-manifests.yaml`.
+*/ -}}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  annotations:
+    helm.sh/hook: pre-delete
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
```
gsanchezgavier, line 12: is this capable of understand that the super-agent-deployment helm release has been unninstalled by Flux before continue?

gsanchezgavier (approved):
LGTM I just left some questions

charts/super-agent/templates/job-manifests.yaml
```diff
@@ -0,0 +1,34 @@
...
+Flux have it CRDs on the `template` folder, So we cannot have the CRs we need on our template folder. They will fail to apply.
+
+To be able to apply them, we have a post-installation hook that run `kubectl apply` with the manifests after the CRDs exist on the cluster.
+To delete them, there is a pre-delete hook that removes them.
+*/}}
+  helm-release.yaml: |
+    {{- include "newrelic-super-agent.helm.release" . | nindent 4 }}
+  helm-repository.yaml: |
+    {{- include "newrelic-super-agent.helm.repository" . | nindent 4 }}
+{{- end }}
```
sigilioso, line 34: 📖 Thank you!

sigilioso (approved):
Great job! Thanks for all the effort 👍

marcsanmi (approved):
Great job! 👏
 
Completely unrelated, but just in case you want to leverage this PR, the release name is missing as well in the `README.md` of the super-agent-deployment chart [here](https://github.com/newrelic/helm-charts/tree/master/charts/super-agent-deployment#helm-installation).

charts/super-agent/templates/_helm-release.yaml
kang-makes: Really good question, it took me some time to figure out.

There is an issue that says that they planned to have watch capabilities that never implemented: https://github.com/fluxcd/helm-controller/issues/42

So, the way to trigger a release is by using `interval`.

There are two intervals:
 * [Helm releases' `.spec.interval`](https://github.com/newrelic/helm-charts/pull/1175/files#diff-74e915ed23d7fc934887bc2a55f84ef93d8ca1e7d0c6a3e235be39830a6532d9R14) spscifies the reconcile interval so it would take `.spec.interval` to set the release with the new values.
 * [Helm releases' `.spec.chart.spec.interval`](https://github.com/newrelic/helm-charts/pull/1175/files#diff-74e915ed23d7fc934887bc2a55f84ef93d8ca1e7d0c6a3e235be39830a6532d9R23) checks for chart updates.
 * [Helm repository' `.spec.interval`](https://github.com/newrelic/helm-charts/pull/1175/files#diff-470e71a1fb8171edfe1dc2f4493e4b25fe730119f40ca39ee905d680d459521eR13) set the amount of time the source controller publishes a new chart for the helm controller to set it up.
 
Do you think it makes sense to have different intervals? I reused the same value ([this one](https://github.com/newrelic/helm-charts/pull/1175/files#diff-3bd49393f47bdfcb35b90c33237f384109da730bc892fa317ba8db58d29d7975R10)) for all of them.

gsanchezgavier: I would say that .spec.interval should be as faster as it could to make changes to values be applied as fast as possible. 
i think the other two we can tackled in the future when autoupdates are analysed

charts/super-agent/templates/uninstall-job.yaml
gsanchezgavier: oh i understood that by default flux does not remove any deployed workload if its removed so that was my concern but thanks for checking it

PR #1178, opened 2023-10-23
[common-library][newrelic-logging] feature: Adding support for GKE Autopilot

jiggyjigsj:
This pull request enhances the New Relic Helm chart by introducing a feature flag for Google Kubernetes Engine Autopilot, making it more compatible and functional in Autopilot environments. 

#### Is this a new chart
No

Currently, there is only support for AWS Fargate.  This PR adds support for GKE Autopilot.

* Disabling Hostnetwork: The PR disables the hostnetwork, ensuring that network configurations align with GKE Autopilot's requirements.

* Fluentbit DB Location Override: It addresses an issue where Fluentbit attempts to write to a read-only location. This change allows for proper Fluentbit operation within the chart.

* Common Library Addition: The PR introduces a common library to detect whether GKE Autopilot is enabled, either through the gkeAutopilot or global.gkeAutopilot setting. This addition streamlines the codebase and makes it more maintainable.

* Updated Mounted Path of logging: Additionally, this PR updates the mounted path to /var/log instead of just /var. This adjustment is necessary because GKE Autopilot does not support mounting the /var directory directly from the host. By specifying the /var/log path, the PR ensures that the New Relic Helm chart functions seamlessly in GKE Autopilot environments, enhancing its compatibility and usability.

This is my first chart change, so I'm eager to get this pull request out and see if we can get a build created. Your feedback and guidance are greatly appreciated.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/common-library/templates/_gke-autopilot.tpl
+++ b/library/common-library/templates/_gke-autopilot.tpl
@@ -1,6 +1,26 @@
-{{/*
-Returns gkeAutopilot
-*/}}
+{{- /*
+This helper allows to override the global `.global.gkeAutopilot` with the value of `.gkeAutopilot`.
+Returns "true" if `gkeAutopilot` is enabled, otherwise "" (empty string)
+# Source: ./_hostNetwork.tpl
+*/ -}}
 {{- define "newrelic.common.gkeAutopilot" -}}
-{{- coalesce .Values.global.gkeAutopilot .Values.gkeAutopilot }}
+{{- /* `get` will return "" (empty string) if value is not found, and the value otherwise, so we can type-assert with kindIs */ -}}
+{{- if (get .Values "gkeAutopilot" | kindIs "bool") -}}
+    {{- if .Values.gkeAutopilot -}}
+        {{- /*
+            We want only to return when this is true, returning `false` here will template "false" (string) when doing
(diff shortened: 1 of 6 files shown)


library/common-library/templates/_gke-autopilot.tpl
```diff
@@ -0,0 +1,8 @@
+{{/*
+Returns gkeAutopilot
+*/}}
+{{- define "newrelic.common.gkeAutopilot" -}}
+{{- if (or .Values.global.gkeAutopilot .Values.gkeAutopilot) -}}
+{{- default .Values.global.gkeAutopilot .Values.gkeAutopilot -}}
```
jiggyjigsj, line 6: I've been playing around a bit, and i think this is best updated to below.  I haven't commit it yet cause my unit tests fail due to nil values.  

```
{{- define "newrelic.common.gkeAutopilot" -}}
{{- coalesce .Values.global.gkeAutopilot .Values.gkeAutopilot }}
{{- end -}}
```

danybmx:
Hi! Support for GKE autopilot has been added through generic persistence configurations #1259

https://github.com/newrelic/helm-charts/blob/master/charts/newrelic-logging/README.md#gke-autopilot-example

Thanks!

PR #1191, opened 2023-11-09
[newrelic-logging]: Remove unneeded hostNetwork setting

### LINKED ISSUE #335: [newrelic-logging] Mount /etc/hostname from host rather than use "hostNetwork: true"

`hostNetwork: true` for Pods is often considered a privileged configuration, when possible, alternative solutions should be, which are better tailored to the actual requirements.

I'd like to replace `hostNetwork: true`, so logging agent use pod network and mount `/etc/hostname` from the host instead, to retain existing functionality of properly identifying the nodes.

**Additional context**

Spotted while reviewing the manifests.

bmcfeely:
#### Is this a new chart

Nope

We think this value got copied-and-pasted from another chart when it was originally created. We want the `newrelic-logging` pod to have as little access as possible, to increase security -- it should just need as much access as it needs to do its job and no more. As far as we can tell, it doesn't need to connect to the host network.

Also, PodSecurityPolicy is now deprecated and
[removed from the latest versions](https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/) of Kubernetes, so just removing it (since it hasn't worked in a few k8s releases, and I haven't heard any issues). There is no direct replacement for a PSP, see [here](https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/#what-does-this-mean-for-you) for Kubernetes's advice. 

There is a risk that customers who use this flag will be surprised that the PSP has gone away, but I'm not sure how to identify which customers are using `rbac.pspEnabled`.

- Fixes #335 
  - Note: I'll still need to figure out whether we need to mount `/etc/hostname`

I've verified that the logs forwarded by the chart have the same attributes (specifically, the same `host` value) before and after this change. So this `hostNetwork` change might be a no-op -- no functionality will change (besides us not asking for the pod to have access to the host network)

#### Checklist
- [x] Chart Version bumped


bmcfeely:
We're going to add the PSP back, since we currently support back to k8s 1.23, which supports PSPs. I'll create a new PR

PR #1240, opened 2024-01-17
WIP [newrelic-logging] Add persistent volumes for fb db

danybmx:
This is a WIP

#### Is this a new chart
No

This will enable to use the logging chart in environments where hostPath is not possible as it doesn't need to have read/write permission over /var/log.

- 

- 

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/newrelic-logging/values.yaml
+++ b/charts/newrelic-logging/values.yaml
@@ -284,3 +284,12 @@ windows:
 
 # -- Sets pod dnsConfig. Can also be configured with `global.dnsConfig`
 dnsConfig: {}
+
+# -- Use persistent volumes to store the fluent-bit db, this will override `fluentBit.db` path and use `/db/${NODE_NAME}-fb.db` file instead
+persistentVolume:
+  enabled: false
+  size: 10Gi
+  # -- The storage class should allow ReadWriteMany mode
+  storageClassName: "standard-rwx"
+  # -- If dynamicProvisioning is enabled the chart will create only the PersistentVolumeClaim
+  dynamicProvisioning: true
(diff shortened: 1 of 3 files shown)


charts/newrelic-logging/values.yaml
```diff
@@ -284,3 +284,12 @@ windows:
 
 # -- Sets pod dnsConfig. Can also be configured with `global.dnsConfig`
 dnsConfig: {}
+
+# -- Use persistent volumes to store the fluent-bit db, this will override `fluentBit.db` path and use `/db/${NODE_NAME}-fb.db` file instead
+persistentVolume:
+  enabled: false
+  size: 10Gi
+  # -- The storage class should allow ReadWriteMany mode
+  storageClassName: "standard-rwx"
```
luckslovez, line 293: Is this cloud specific? AFAIK each cloud provider have its own volume provisioning solution... We maybe could even skip the PV completely and use a storage class (cloud specific) that can dynamically provide (create) a PV for a given PVC 🤔

danybmx: I'm looking now at Local Persistent Volumes, AFAIK the use of ReadWriteMany is mandatory in this case as this will only create a volume for all nodes (daemonsets and deployments share the same volumes, statefulset create a different one per instance)

danybmx: Sure, this one is for GKE, each cloud provider will provide their own storage classes but those should be configured manually, or at least IDK how to get the one we're interested (RWX) without setting it directly. The option to change it should be available because maybe someone want to use this in a custom kubernetes cluster or so and they need to set the PV (that's why I'm adding the dynamicProvisioning flag)

danybmx:
Closed in favor of #1259

PR #1255, opened 2024-01-23
Revert "chore(deps): update helm release flux2 to v2.12.2 (#1222)"

juanjjaramillo:
This reverts commit 594230530073f1ac622a727040230560a3d3f9a8.

### DIFF
--- a/charts/super-agent/Chart.yaml
+++ b/charts/super-agent/Chart.yaml
@@ -3,12 +3,12 @@ name: super-agent
 description: Bootstraps New Relic' Super Agent
 
 type: application
-version: 0.0.1
+version: 0.0.1-beta
 
 dependencies:
   - name: flux2
     repository: https://fluxcd-community.github.io/helm-charts
-    version: 2.12.2
+    version: 2.12.1
     condition: flux2.enabled
   - name: common-library
     version: 1.1.1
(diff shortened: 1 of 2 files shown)


charts/super-agent/Chart.yaml
```diff
@@ -3,12 +3,12 @@ name: super-agent
 description: Bootstraps New Relic' Super Agent
 
 type: application
-version: 0.0.1
+version: 0.0.1-beta
```
paologallinaharbur, line 6: I am not sure about the effects of merging this.. would a new release would be pushed again with 0.0.1-beta?

charts/super-agent/Chart.yaml
```diff
@@ -3,12 +3,12 @@ name: super-agent
...
 
 dependencies:
   - name: flux2
     repository: https://fluxcd-community.github.io/helm-charts
-    version: 2.12.2
+    version: 2.12.1
```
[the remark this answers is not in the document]
juanjjaramillo: Done

charts/super-agent/Chart.yaml
juanjjaramillo: No. The chart releaser will check [index.yaml looking for version 0.0.1-beta](https://github.com/newrelic/helm-charts/blob/gh-pages/index.yaml#L22528-L22558), and since it is available no release will be triggered.

PR #1256, opened 2024-01-23
Revert "Update index.yaml"

juanjjaramillo:
This reverts commit ce1925fe6b5ae1fbfef93c61623941bd75c1ad83.

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -22526,37 +22526,6 @@ entries:
     version: 0.1.0
   super-agent:
   - apiVersion: v2
-    created: "2024-01-23T11:27:52.519945385Z"
-    dependencies:
-    - condition: flux2.enabled
-      name: flux2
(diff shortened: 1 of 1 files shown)


index.yaml
```diff
@@ -26151,4 +26120,4 @@ entries:
     urls:
     - https://github.com/newrelic/helm-charts/releases/download/synthetics-minion-1.0.17/synthetics-minion-1.0.17.tgz
     version: 1.0.17
-generated: "2024-01-23T11:27:52.524988816Z"
+generated: "2024-01-23T05:55:18.144829537Z"
```
paologallinaharbur, line 26123: I would not revert the time here since it can create issues (helm repo update does not work properly).
Instead, I would update it with a new time (2024-01-24T11:27:52.524988816Z)

paologallinaharbur (approved):

PR #1259, opened 2024-01-26
[newrelic-logging] Add fluent-bit persistence modes

danybmx:
#### Is this a new chart
No

This configurations will give the customer more flexibility on how they want to persist the fluent-bit database.

The default is still the same, but it will help to run fluent-bit in clusters with more restrictive security constraints like GKE autopilot which is the default mode in Google Cloud when creating a K8S cluster.

To ensure the pervious behavior isn't modified I've run helm in dry-run with master and this branch and comparted the results

```
# On this branch
helm upgrade --install newrelic-logging . --set global.licenseKey=SAMPLE-KEY --namespace=newrelic --set global.cluster=testing-pvc-drodriguez --dry-run > now

# On master
helm upgrade --install newrelic-logging . --set global.licenseKey=SAMPLE-KEY --namespace=newrelic --set global.cluster=testing-pvc-drodriguez --dry-run > before

# Compare
diff now before
3c3
< LAST DEPLOYED: Fri Jan 26 14:09:32 2024
---
> LAST DEPLOYED: Fri Jan 26 14:10:08 2024
17c17
<     chart: newrelic-logging-1.20.0
---
>     chart: newrelic-logging-1.19.0
30c30
<     chart: newrelic-logging-1.20.0
---
>     chart: newrelic-logging-1.19.0
46c46
<     chart: newrelic-logging-1.20.0
---
>     chart: newrelic-logging-1.19.0
102c102
<     chart: newrelic-logging-1.20.0
---
>     chart: newrelic-logging-1.19.0
120c120
<     chart: newrelic-logging-1.20.0
---
>     chart: newrelic-logging-1.19.0
141c141
<     chart: newrelic-logging-1.20.0
---
>     chart: newrelic-logging-1.19.0
157c157
<         checksum/fluent-bit-config: ee20edded8fd6049cd9509a3b5829165049a8f6308790dd7f09a0512f0806da7
---
>         checksum/fluent-bit-config: 7752c3ef9f7d76446a914cf3b1faab2a214956dc59ec568ca9f5f8fb93c3d96f
167d166
<       initContainers:
214c213
<             - name: logs
---
>             - name: var
227c226
<         - name: logs
---
>         - name: var
```

As you can see we only change the volume name that makes more sense to have it set as "logs" instead "var" because it could be other paths. The other diffs are because checksums and chart version.

#### Checklist
- [x] Chart Version bumped


jcsobrino (approved):
Great job, @danybmx! Congratulations 💯

charts/newrelic-logging/templates/daemonset.yaml
```diff
@@ -47,8 +47,17 @@ spec:
       securityContext:
         {{- . | nindent 8 }}
       {{- end }}
-      {{- if .Values.initContainers }}
       initContainers:
+      {{- if and (.Values.fluentBit.persistence) (eq .Values.fluentBit.persistence.mode "persistentVolume") }}
+        - name: init
+          image: busybox:1.36
+          command: ["/bin/sh", "-c"]
+          args: ["/bin/find /db -type f -mtime +1 -delete"] # Delete all db files not updated in the last 24h
```
luckslovez, line 55: So clean ✨

luckslovez (approved):

charts/newrelic-logging/templates/daemonset.yaml
```diff
@@ -118,8 +139,15 @@ spec:
...
-              mountPath: /var
+            - name: logs
+              mountPath: {{ .Values.fluentBit.linuxMountPath | default "/var" }}
+              {{- if and (.Values.fluentBit.persistence) (ne .Values.fluentBit.persistence.mode "hostPath") }}
+              readOnly: true
+              {{- end }}
+            {{- if and (.Values.fluentBit.persistence) (eq .Values.fluentBit.persistence.mode "persistentVolume") }}
+            - name: fb-db-pvc
+              mountPath: /db
+            {{- end }}
```
jsubirat, line 150: This is how the DB path will be observed from _inside_ the pod (and this will be a new pod). Even if we change it here, we won't incur any breaking change and we will have a clear separation of the "logs" mount and the "db" mount:

```
{{- if ne .Values.fluentBit.persistence.mode "none"}}
  - name: fb-db
    mountPath: /db
{{- end }}
```

charts/newrelic-logging/templates/daemonset.yaml
```diff
@@ -118,8 +139,15 @@ spec:
           volumeMounts:
             - name: fluent-bit-config
               mountPath: /fluent-bit/etc
-            - name: var
-              mountPath: /var
+            - name: logs
+              mountPath: {{ .Values.fluentBit.linuxMountPath | default "/var" }}
+              {{- if and (.Values.fluentBit.persistence) (ne .Values.fluentBit.persistence.mode "hostPath") }}
+              readOnly: true
+              {{- end }}
```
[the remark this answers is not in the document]
jsubirat: Ah, now I understand why you kept "/var", is it because we also need to keep /var/lib for symlinks? I would suggest to just mount the needed directories, [like the Fluent Bit Helm chart does](https://github.com/fluent/helm-charts/blob/main/charts/fluent-bit/values.yaml#L444-L449), then.

charts/newrelic-logging/templates/daemonset.yaml
```diff
@@ -134,9 +162,18 @@ spec:
         - name: fluent-bit-config
           configMap:
             name: {{ template "newrelic-logging.fluentBitConfig" . }}
-        - name: var
+        - name: logs
           hostPath:
-            path: /var
+            path: {{ .Values.fluentBit.linuxMountPath | default "/var" }}
```
jsubirat, line 167:
```suggestion
            path: /var/log
```

And then a separate mount for `/var/lib/docker/containers`?

charts/newrelic-logging/templates/daemonset.yaml
```diff
@@ -47,8 +47,17 @@ spec:
       securityContext:
         {{- . | nindent 8 }}
       {{- end }}
-      {{- if .Values.initContainers }}
       initContainers:
+      {{- if and (.Values.fluentBit.persistence) (eq .Values.fluentBit.persistence.mode "persistentVolume") }}
```
jsubirat, line 51: General comment (for here and the rest of the PR). If we set a default value for `.Values.fluentBit.persistence.mode` in the `values.yml` file, I suspect we don't need to do `and (.Values.fluentBit.persistence) (eq .Values.fluentBit.persistence.mode "<something>")` everywhere, right? We can just go for `{{- if eq .Values.fluentBit.persistence.mode "<something>" }}`

charts/newrelic-logging/templates/daemonset.yaml
```diff
@@ -47,8 +47,17 @@ spec:
       securityContext:
```
jsubirat, line 47: **⚠️ ⚠️ ⚠️ ⚠️   Changes in `daemonset-windows.yaml are missing!!! ⚠️ ⚠️ ⚠️ ⚠️**

charts/newrelic-logging/values.yaml
```diff
@@ -27,6 +27,7 @@
 fluentBit:
   logLevel: "info"
   path: "/var/log/containers/*.log"
+  linuxMountPath: /var
```
jsubirat, line 30: I believe we don't need this 🤔

charts/newrelic-logging/values.yaml
```diff
@@ -43,6 +44,39 @@ fluentBit:
   #     fieldRef:
   #       fieldPath: metadata.name
 
+  # Indicates how fluent-bit database is persisted
+  persistence:
+    # Define the persistent mode for fluent-bit db, allowed options are `hostPath` (default), `none`, `persistentVolume`.
+    # - `hostPath` will use hostPath to store the db file on the node disk.
+    # - `none` will disable the fluent-bit db file, this could cause log duplication or data loss in case fluent-bit gets restarted.
+    # - `persistentVolume` will use a ReadWriteMany persistent volume to store the db file. This will override `fluentBit.db` path and use `/db/${NODE_NAME}-fb.db` file instead.
```
jsubirat, line 52: Ah, this is what I was looking for. Can you somehow display this in the readme as well?

charts/newrelic-logging/templates/daemonset.yaml
```diff
@@ -92,8 +101,20 @@ spec:
...
+            - name: NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+            - name: FB_DB
+              value: "/db/$(NODE_NAME)-fb.db"
+            {{- else }}
+            - name: FB_DB
+              value: ""
+            {{- end }}
```
jsubirat, line 117: This is going to be a 🌵  comment, in fact it's just a "thinking-out-loud-comment" rather something that we have to change.

With `.Values.fluentBit.db`, we allowed the user to choose the path of the DB in the _host_ (because we were using hostPath and mounting the whole `/var` of the host). And indirectly, this also became the path of the DB as seen from inside the Fluent Bit pod. But nothing forces us to use the same path _inside_ the pod as well.

What I mean is that `FB_DB` could _always_ be `/db/$(NODE_NAME)-fb.db` from now on, even if we're using `hostPath`, as long as we mount the `fb-db` mount at `/db`. The only caveat is that during this transition,  customers may see log duplication/loss as a new DB would be created in their hosts. Also, customers who may have configured another value for `fluentBit.db` would also have some effects I don't even want to consider. 😓

charts/newrelic-logging/templates/persistentvolume.yaml
```diff
@@ -0,0 +1,52 @@
+{{- if not (empty .Values.fluentBit.persistence) }}
+{{- if and (eq .Values.fluentBit.persistence.mode "persistentVolume") (not .Values.fluentBit.persistence.persistentVolume.existingVolumeClaim) }}
+{{- if and (not .Values.fluentBit.persistence.persistentVolume.dynamicProvisioning) (not .Values.fluentBit.persistence.persistentVolume.existingVolume) }}
```
jsubirat, line 3: I see 3 ifs here and just 2 closing "ends" at the end of the file, is this correct? 🤔 

Also, can these conditions be probably simplified as in just a single "if":
```
{{- if and (eq .Values.fluentBit.persistence.mode "persistentVolume") 
  (not .Values.fluentBit.persistence.persistentVolume.existingVolumeClaim) 
  (not .Values.fluentBit.persistence.persistentVolume.dynamicProvisioning) 
  (not .Values.fluentBit.persistence.persistentVolume.existingVolume) }}
```

(don't know if multiline ifs are allowed though)

charts/newrelic-logging/templates/persistentvolume.yaml
```diff
@@ -0,0 +1,52 @@
...
+  {{- end }}
+spec:
+  storageClassName: {{ .Values.fluentBit.persistence.persistentVolume.storageClass }}
+  accessModes:
+    - ReadWriteMany
+{{- if .Values.fluentBit.persistence.persistentVolume.existingVolume }}
+  volumeName: {{ .Values.fluentBit.persistence.persistentVolume.existingVolume }}
+{{- else if not .Values.fluentBit.persistence.persistentVolume.dynamicProvisioning }}
+  volumeName: {{ template "newrelic-logging.fullname" . }}-pv
+{{- end }}
```
jsubirat, line 44: Should we prevent somehow the user to be able to specify both `fluentBit.persistence.persistentVolume.existingVolume` and `fluentBit.persistence.persistentVolume.dynamicProvisioning`? I feel like these are contradictory options.

charts/newrelic-logging/templates/daemonset.yaml
danybmx: It'll break for people that already have their values with the db path set to `/var/log/flb_kube.db`

charts/newrelic-logging/templates/daemonset.yaml
danybmx: In that case we need to skip mounting it on GKE autopilot anyways because /var/lib is not allowed to be mounted. So instead linuxMountPath we should have skipMountingVarLib or something that have not much sense to me

charts/newrelic-logging/templates/daemonset.yaml
danybmx: We need it for upgrades or that's demonstrated by the tests included. If it tries to upgrade using previous values it'll fail. https://github.com/newrelic/helm-charts/actions/runs/7669042487/job/20902151151

charts/newrelic-logging/templates/daemonset.yaml
danybmx: Nice catch! Thanks :bow:

charts/newrelic-logging/values.yaml
danybmx: Will do

charts/newrelic-logging/README.md
```diff
@@ -106,56 +106,66 @@ helm upgrade --install newrelic-bundle newrelic/nri-bundle \
...
+| `fluentBit.additionalEnvVariables`                           | Additional environmental variables for fluentbit pods                                                                                                                                                                                                                                                                                                                                | `[]]`                                                                           |
+| `fluentBit.persistence.mode`                                 | The persistence mode you want to use, options are "hostPath", "persistentVolume" or "none"                                                                                                                                                                                                                                                                                           | hostPath                                                                        |
```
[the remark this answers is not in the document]
danybmx: Will do

charts/newrelic-logging/templates/daemonset.yaml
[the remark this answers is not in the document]
danybmx: Mentioned already in another comment but that way we'll need to add a "skipVarLib" flag or something for GKE autopilot, I think it's more flexible to do it that way. Could be people using other path than var for logs (that's why we allow `fluentBit.path` in the configuration I believe)

charts/newrelic-logging/values.yaml
danybmx: @jsubirat - You mean the value `/var` or the variable itself?

charts/newrelic-logging/templates/daemonset.yaml
danybmx: Yep, I'm skipping this to prevent undesired behaviors, although could be something good to do in a new PR / release

charts/newrelic-logging/templates/persistentvolume.yaml
danybmx: I tried and it failed so I assume helm only allow 2 conditions but will recheck.

charts/newrelic-logging/templates/persistentvolume.yaml
danybmx: I mean`dynamicProvisioning` doesn't make sense if an `existingVolume` is provided, but if someone is providing an `existingVolume` I don't think they should need to add the `dynamicProvisioning: false` as well (as it defaults to true).

charts/newrelic-logging/values.yaml
jsubirat: I meant the variable itself, but after your explanation about /var/lib you can ignore this comment

charts/newrelic-logging/templates/persistentvolume.yaml
jsubirat: Exactly. Somehow Helm should throw an error (or a warning at least) informing of this inconsistent configuration. This validation could happen in the `tpl` file. We could use `fail`, like in [here](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_license_secret.yaml.tpl#L7-L9).

charts/newrelic-logging/templates/daemonset.yaml
jsubirat: I see, thanks for verifying

charts/newrelic-logging/templates/persistentvolume.yaml
danybmx: I was saying the opposite but i'll take it.

charts/newrelic-logging/values.yaml
danybmx: Done! [2948836](https://github.com/newrelic/helm-charts/pull/1259/commits/2948836cfb7cf16788942de4139ff7e7ec02ef2e)

danybmx:
@jsubirat windows will have the none retention mode but not the persistentVolume, this could come up later a newer PR if requested.

charts/newrelic-logging/README.md
```diff
@@ -167,12 +167,33 @@ See [values.yaml](values.yaml) for the default values
...
 | `fluentBit.criEnabled`                                       | We assume that `kubelet`directly communicates with the container engine using the [CRI](https://kubernetes.io/docs/concepts/overview/components/#container-runtime) specification. Set this to `false` if your K8s installation uses [dockershim](https://kubernetes.io/docs/tasks/administer-cluster/migrating-from-dockershim/) instead, in order to get the logs properly parsed. | `true`                                                                          |
 
+### Fluent Bit persistence modes
+
+Fluent Bit uses a database file to keep track of tailed and sent log lines from files, this database file is by default stored on the node using a `hostPath` mount. It's specifically stored (by default) on `/var/log/flb_kube.db` to keep things simple, as we're already mounting `/var` for accessing container logs.
```
jsubirat, line 172:
```suggestion
Fluent Bit uses a database file to keep track of log lines read from files (offsets). This database file is stored in the host node by default, using a `hostPath` mount. It's specifically stored (by default) in `/var/log/flb_kube.db` to keep things simple, as we're already mounting `/var` for accessing container logs.
```

charts/newrelic-logging/README.md
```diff
@@ -167,12 +167,33 @@ See [values.yaml](values.yaml) for the default values
...
+
+Sometimes the security constraints of some clusters doesn't allow to mount hostPath in read-write mode, that's why you can chose among the following
```
jsubirat, line 174:
```suggestion
Sometimes the security constraints of some clusters don't allow mounting `hostPath`s in read-write mode. That's why you can chose among the following
```

charts/newrelic-logging/README.md
```diff
@@ -167,12 +167,33 @@ See [values.yaml](values.yaml) for the default values
...
+persistence modes. Each one have their pros and cons.
```
jsubirat, line 175:
```suggestion
persistence modes. Each one has their pros and cons.
```

charts/newrelic-logging/README.md
```diff
@@ -167,12 +167,33 @@ See [values.yaml](values.yaml) for the default values
...
+
+- `hostPath` (default) will use hostPath mount to store the db file on the node disk, the easier an most reliable, but prohibited by some security policies.
```
jsubirat, line 177:
```suggestion
- `hostPath` (default) will use a `hostPath` mount to store the DB file on the node disk. This is the easiest, cheapest an most reliable option, but prohibited by some cloud vendor security policies.
```

charts/newrelic-logging/README.md
```diff
@@ -167,12 +167,33 @@ See [values.yaml](values.yaml) for the default values
...
+- `none` will disable the fluent-bit db file, this could cause log duplication or data loss in case fluent-bit gets restarted.
```
jsubirat, line 178:
```suggestion
- `none` will disable the Fluent Bit DB file. This can cause log duplication or data loss in case Fluent Bit gets restarted.
```

charts/newrelic-logging/README.md
```diff
@@ -167,12 +167,33 @@ See [values.yaml](values.yaml) for the default values
...
+- `persistentVolume` (linux only) will use a ReadWriteMany persistent volume to store the db file. This will override `fluentBit.db` path and use `/db/${NODE_NAME}-fb.db` file instead. If you use this option in a windows cluster it will default to none on windows nodes.
```
jsubirat, line 179:
```suggestion
- `persistentVolume` (Linux only) will use a `ReadWriteMany` persistent volume to store the DB file. This will override the `fluentBit.db` path and use `/db/${NODE_NAME}-fb.db` instead. If you use this option in a Windows cluster it will default to `none` on Windows nodes.
```

charts/newrelic-logging/README.md
```diff
@@ -167,12 +167,33 @@ See [values.yaml](values.yaml) for the default values
...
+
+#### GKE Autopilot example
+
+If you're using the persistentVolume persistence mode you need to provide at least the storageClass, and it should be ReadWriteMany. This is an example of the configuration for persistence in GKE Autopilot.
```
jsubirat, line 183:
```suggestion
If you're using the `persistentVolume` persistence mode you need to provide at least the `storageClass`, and it should be `ReadWriteMany`. This is an example of the configuration for persistence in [GKE Autopilot](https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-overview).
```

jsubirat (commented):
Added a few text suggestions 👍

charts/newrelic-logging/templates/daemonset.yaml
danybmx: Finally windows have partial support, just adding "none" persistence mode there.

PR #1269, opened 2024-02-07
[super-agent] feat: disableWait for SA installation

gsanchezgavier:
use `disableWait` to unblock modification/deletion of the HelmRelease
  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/super-agent/values.yaml
+++ b/charts/super-agent/values.yaml
@@ -82,6 +82,7 @@ helm:
     # before give up installing the chart in case it hits an error.
     # @default -- See `values.yaml`
     install:
+      disableWait: true  # Prevents blocking the modification/deletions of the HelmRelease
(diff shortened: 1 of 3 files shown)


charts/super-agent/values.yaml
```diff
@@ -91,6 +92,7 @@ helm:
     # installing the chart in case it hits an error.
     # @default -- See `values.yaml`
     upgrade:
```
paologallinaharbur, line 94: do we need here as well:
 - disableWaitForJobs
and 
```
          rollback:
            disableWait: true
            disableWaitForJobs: true
```
?

gsanchezgavier: i miss that!

paologallinaharbur (approved):

PR #1277, opened 2024-02-09
Improve super-agent helm chart error handling

marcsanmi:
This PR improves the Kubernetes Job script by better distinguishing between actual errors and warnings in kubectl apply output. 

It checks the command's exit status and looks for success indicators, ensuring only failures trigger error messages. This change should clarify a bit the feedback on CRs application.

### DIFF
--- a/charts/super-agent/templates/install-job.yaml
+++ b/charts/super-agent/templates/install-job.yaml
@@ -15,6 +15,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   ttlSecondsAfterFinished: 120
+  backoffLimit: 1
   template:
     spec:
       restartPolicy: Never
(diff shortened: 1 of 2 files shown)


charts/super-agent/templates/install-job.yaml
```diff
@@ -22,11 +22,22 @@ spec:
       containers:
         - name: apply-crs
           image: bitnami/kubectl
-          command:
-            - kubectl
-            - apply
-            - -f
-            - /manifests/
+          command: [ "/bin/bash","-c" ]
+          args:
```
gsanchezgavier, line 26: I just did a quick check on this with using a bersion like `version: ">=0.0.8-beta"` and the installation enters into an infinite loop of launching the job, woudl be possible to fix that state?

marcsanmi: `Jobs` in Kubernetes have a default `backoffLimit`  of 6, so it wouldn't happen to face an infinite loop here. However, let's set it explicitly in case the default change in the future and reduce it to 1.

gsanchezgavier (approved):

PR #1278, opened 2024-02-09
[super-agent-deployment] include uninstall job

sigilioso:
#### Is this a new chart?
No

This PR includes a helm pre-delete hook to delete that are managed by the super-agent, since some of the resources may not be garbage collected.

There are still _race conditions_ when this chart is deployed by flux and flux is uninstalled in the same step (the CRs can be deleted after flux is uninstalled and the underlying kubernetes resources stop being managed). This will be addressed in a follow-up PR by updating the uninstall-job in the `super-agent` chart (#1279).

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/super-agent-deployment/templates/uninstall-job.yaml
+++ b/charts/super-agent-deployment/templates/uninstall-job.yaml
@@ -33,12 +33,11 @@ spec:
             - |
               set -o pipefail
               {{- /* Delete the standard resources (configmaps) managed by the super-agent */ -}}
-              {{ $managedResourcesNamespace := .Release.Namespace }}
-              kubectl delete configmaps -n {{ $managedResourcesNamespace }} -l {{ $saResourcesLabelSelector }}
+              kubectl delete configmaps -n {{ .Release.Namespace }} -l {{ $saResourcesLabelSelector }}
               {{- /* Delete the CRs managed by the super-agent if the corresponding CRDs exist */ -}}
               {{ range $i, $cr := $saCRList }}
               if kubectl api-resources -o name |grep {{ $cr }}; then
-                kubectl delete {{ $cr }} -n {{ $managedResourcesNamespace }} -l {{ $saResourcesLabelSelector }}
+                kubectl delete {{ $cr }} -n {{ $.Release.Namespace }} -l {{ $saResourcesLabelSelector }}
               fi
               {{ end -}}
 
--- a/charts/super-agent-deployment/templates/uninstall-job.yaml
+++ b/charts/super-agent-deployment/templates/uninstall-job.yaml
@@ -3,7 +3,8 @@
 {{- /*
 The resources managed by the super-agent and the label selector are hardcoded on the super-agent.
 */ -}}
-{{- $saResourcesList := (list "configmaps" "helmreleases" "helmrepositories") -}}
+{{- $saResourceList := (list "configmaps") -}}
+{{- $saCRList := (list "helmreleases.helm.toolkit.fluxcd.io" "helmrepositories.source.toolkit.fluxcd.io") -}}
 {{- $saResourcesLabelSelector := "app.kubernetes.io/managed-by=newrelic-super-agent" -}}
 {{- /*
 To understand why this job installs manifests instead of using Helm, read the comment at `job-manifests.yaml`.
@@ -30,7 +31,17 @@ spec:
             - bash
           args:
             - -c
-            - {{ range $i, $resourceName := $saResourcesList }}
-              {{- if $i }} &&{{ end }} kubectl delete {{ $resourceName }} -l {{ $saResourcesLabelSelector }}
-              {{- end -}}
+            - |
+              set -o pipefail
+              {{- /* Delete the standard resources managed by the super-agent */ -}}
+              {{ range $i, $resourceName := $saResourceList }}
+              kubectl delete {{ $resourceName }} -l {{ $saResourcesLabelSelector }}
+              {{ end -}}
+              {{- /* Delete the CRs managed by the super-agent if the corresponding CRDs exist */ -}}
+              {{ range $i, $cr := $saCRList }}
+              if kubectl api-resources -o name |grep {{ $cr }}; then
+                kubectl delete {{ (splitList "." $cr) |first }} -l {{ $saResourcesLabelSelector }}
+              fi
+              {{ end -}}
+
 {{- end -}}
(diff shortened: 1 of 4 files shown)


charts/super-agent-deployment/values.yaml
```diff
@@ -96,6 +96,10 @@ fedramp:
 # @default -- `false`
 verboseLog:
 
+# -- Enable the cleanup of super-agent managed resources when the chart is uninstalled.
+# If disabled, agents and / or agent configurations could remain in the cluster after the chart is uninstalled.
```
paologallinaharbur, line 100: I would specify that also the subchart has a similar mechanism

charts/super-agent-deployment/templates/uninstall-job.yaml
```diff
@@ -0,0 +1,47 @@
...
+        - name: delete-crs
+          image: bitnami/kubectl
+          command:
+            - bash
+          args:
+            - -c
+            - |
+              set -o pipefail
+              {{- /* Delete the standard resources managed by the super-agent */ -}}
+              {{ range $i, $resourceName := $saResourceList }}
```
paologallinaharbur, line 37: why creating a list with one element? I find it confusing

charts/super-agent-deployment/templates/uninstall-job.yaml
```diff
@@ -0,0 +1,47 @@
...
+              kubectl delete {{ $resourceName }} -l {{ $saResourcesLabelSelector }}
```
paologallinaharbur, line 38: I would restrict the namespace, currently I am not sure what it would do

charts/super-agent-deployment/templates/uninstall-job.yaml
```diff
@@ -0,0 +1,47 @@
...
+              {{ end -}}
+              {{- /* Delete the CRs managed by the super-agent if the corresponding CRDs exist */ -}}
+              {{ range $i, $cr := $saCRList }}
+              if kubectl api-resources -o name |grep {{ $cr }}; then
+                kubectl delete {{ (splitList "." $cr) |first }} -l {{ $saResourcesLabelSelector }}
```
paologallinaharbur, line 43: I don't think it is an issue to use the fqdn:
`k delete helmreleases.helm.toolkit.fluxcd.io`

charts/super-agent-deployment/templates/uninstall-job.yaml
```diff
@@ -0,0 +1,47 @@
...
+  name: {{ $uninstallJobName }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  ttlSecondsAfterFinished: 120
+  template:
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ include "newrelic.common.serviceAccount.name" . }}
+      containers:
+        - name: delete-crs
```
paologallinaharbur, line 28: delete-managed-resources?

charts/super-agent-deployment/templates/uninstall-job.yaml
```diff
@@ -0,0 +1,47 @@
+{{- if .Values.cleanupManagedResources -}}
+{{- $uninstallJobName := include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "uninstall-job" ) -}}
```
paologallinaharbur, line 2: in other places I think we have variables directly where they are used, any reason to move them here?

charts/super-agent-deployment/templates/uninstall-job.yaml
sigilioso: At first config-maps and CRs were together in the same list but they are not anymore so it doesn't make more sense. I'll change it.

charts/super-agent-deployment/templates/uninstall-job.yaml
sigilioso: We algo have this approach in [some other places](https://github.com/newrelic/helm-charts/blob/master/charts/super-agent/templates/install-job.yaml#L2). Since the variable value is quite long I think is easier to read this way.

charts/super-agent-deployment/values.yaml
sigilioso: I've just rephrased, let me know what you think

charts/super-agent-deployment/templates/uninstall-job.yaml
sigilioso: Changed here: [426f796](https://github.com/newrelic/helm-charts/pull/1278/commits/426f79658f207031c8c128f9d3bfded3b70d1b49)

charts/super-agent-deployment/templates/uninstall-job.yaml
sigilioso: Thanks for pointing it out, it simplifies the chart. Updated it here: [426f796](https://github.com/newrelic/helm-charts/pull/1278/commits/426f79658f207031c8c128f9d3bfded3b70d1b49)

charts/super-agent-deployment/templates/uninstall-job.yaml
sigilioso: Updated here: [426f796](https://github.com/newrelic/helm-charts/pull/1278/commits/426f79658f207031c8c128f9d3bfded3b70d1b49)

paologallinaharbur (approved):

charts/super-agent-deployment/templates/uninstall-job.yaml
```diff
@@ -0,0 +1,45 @@
...
+        - name: delete-managed-resources
+          image: bitnami/kubectl
+          command:
+            - bash
+          args:
+            - -c
+            - |
+              set -o pipefail
+              {{- /* Delete the standard resources (configmaps) managed by the super-agent */ -}}
+              {{ $managedResourcesNamespace := .Release.Namespace }}
```
paologallinaharbur, line 36: I think we could use directly .Release.Namespace, I do not see a real benefit, but if you see it easier to read I am fine keeping it 😄

sigilioso: I added the variable to deal with the scope of the `range` below, but I've just realized we can simply use `$.Release.Namespace` there. Thanks! [0960eac](https://github.com/newrelic/helm-charts/pull/1278/commits/0960eac689244d46c693a19881e3beed64f135cc)

PR #1279, opened 2024-02-09
[super-agent] remove super-agent crs on uninstall

sigilioso:
#### Is this a new chart
No

~Extends the `super-agent` chart helm delete hook to remove the CRs managed by the super-agent. This is needed because there is a _race condition_ when uninstalling the chart. The CRs might be deleted after Flux is uninstalled, consequently, the underlying resources are keep in the cluster. Since the pre-delete hook is executed before uninstalling the chart, this we we assure the CR deletion is handled by Flux.~

Since the resources managed by the super-agent are deleted during the `super-agent-deployment` uninstallation, in order to guarantee those are properly deleted here, we need to assure that the corresponding job is executed before Flux is uninstalled. The [default uninstall configuration](https://fluxcd.io/flux/components/helm/helmreleases/#uninstall-configuration) for Flux helmreleaes is already like that, but we set it up explicitly to be aware of the change and avoid issues if those defaults change.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/super-agent/templates/_helm-release.yaml
+++ b/charts/super-agent/templates/_helm-release.yaml
@@ -43,4 +43,7 @@ spec:
   rollback:
     {{- . | toYaml | nindent 4 }}
   {{- end }}
+  uninstall: # Explicitly set to false since executing hooks and waiting for uninstall termination is needed to remove super-agent managed reseources.
+    disableWait: false
+    disableHooks: false
 {{- end -}}
--- a/charts/super-agent/templates/_helm-release.yaml
+++ b/charts/super-agent/templates/_helm-release.yaml
@@ -43,7 +43,7 @@ spec:
   rollback:
     {{- . | toYaml | nindent 4 }}
   {{- end }}
-  uninstall: # Explicitly set to false since executing hooks and waiting for uninstall termination is needed to remove super-agent managed reseources.
+  uninstall: # This is setup explicitly since executing hooks and waiting for uninstall termination is needed to remove super-agent managed resources
     disableWait: false
     disableHooks: false
 {{- end -}}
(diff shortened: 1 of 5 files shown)


charts/super-agent/templates/uninstall-job.yaml
```diff
@@ -23,10 +28,13 @@ spec:
         - name: delete-crs
           image: bitnami/kubectl
           command:
-            - kubectl
-            - delete
-            - -f
-            - /manifests/
+            - bash
+          args:
```
gsanchezgavier, line 32: shouldn't be part of the `super-agent-deployment` ?

paologallinaharbur (approved):
Thanks!

charts/super-agent/templates/uninstall-job.yaml
sigilioso: After properly testing it, it is not needed here. As long as we assure that deleting these manifest will wait for the underlying chart to be uninstalled there should not be race conditions here.

charts/super-agent/templates/_helm-release.yaml
```diff
@@ -43,4 +43,7 @@ spec:
   rollback:
     {{- . | toYaml | nindent 4 }}
   {{- end }}
+  uninstall: # This is setup explicitly since executing hooks and waiting for uninstall termination is needed to remove super-agent managed resources
```
paologallinaharbur, line 46: have we changed the approach, right?
The idea is to simply leverage the chart and wait I guess

paologallinaharbur (approved):

charts/super-agent/templates/_helm-release.yaml
sigilioso: Yes we have! 
Since we can wait for the `super-agent-deployment` uninstall, there is no need to repeat the super-agent managed resources deletion in this chart.

PR #1287, opened 2024-02-19
[nri-bundle] [newrelic-logging] feat: added support for templated cluster name value

pcanilho:
#### Is this a new chart
No.

We're deploying the chart using ArgoCD and ApplicationSets with a cluster generator. To avoid having to bake NewRelic-specific logic into our wider ArgoCD Application generator, we need support for rendering `cluster` as a template variable in order to generate an appropriate value from generic values.

This is a new feature.

#### Checklist
- [x] Chart Version bumped


ndrissi (approved):

isometry (approved):

charts/super-agent-deployment/Chart.yaml
```diff
@@ -9,7 +9,7 @@ appVersion: 0.8.0
 
 dependencies:
   - name: common-library
-    version: 1.1.1
+    version: 1.1.2
```
paologallinaharbur, line 12: [Just providing feedback on chart we own]
modifying all charts at the same time is not allowed by the CI/CD
Moreover, the version was not bumped.

---

On the other hand, this change could be quite important and I'll leave the decision to the K8s team whether we actually want to support this use case (and then "how")

PR #1292, opened 2024-02-23
[NOT FOR MERGING]: prototype: Send logs to team account based on namespace

bmcfeely:
We're trying to sort out a way that CF can allow teams to put their logs in their own team's account instead of account 1.

This PR prototypes providing a way for teams to tell the chart which key (and therefore to which account) they want their logs sent with based on a pod's namespace.  Any pod that doesn't have an explicit mapping from its namespace to an account key will default to be sent with the account 1 key.

This idea is:
* Creates a ConfigMap that teams can put the Vault location for their api keys (this prototype just hardcodes keys locally, I was lazy :P) 
* Creates a chart "hook" that runs on install and upgrade that generates the Fluent Bit `[OUTPUT]` blocks by running a NodeJS app to do that
* Including those `[OUTPUT]` blocks in the "main" fluent-bit.conf

For CF, we'd do something a little different (but basically the same), since CF uses their own Helm chart. I think it would still be mostly the same. I didn't do the prototype there because, uh, I don't know. 😄 

I tested this by: 
- Editing the chart to replace `REDACTED_KEY_FOR_PRODUCTION_756053` with the actual key for production 756053
- Installing the chart with `helm uninstall newrelic-logging && helm install newrelic-logging ./charts/newrelic-logging --set licenseKey=<REDACTED_ACCOUNT_1_KEY> --set cluster=brian-kind && kubectl logs $(kubectl get pod | grep newrelic-logging | grep -v creator | awk '{print $1}')`
- Creating a pod in the `logging` namespace
- Verifying that the `logging` namespace pod's logs go to account 756053 (and _not_ account 1)
- Verifying that the other pods in non-`logging` namespaces go to account 1

### Account 1 logs
(note that it doesn't include logs from the `logging` namespaces)

Screenshot 2024-02-22 at 9 03 41 PM

### Account 756053 logs
(note that it doesn't include logs from other namespaces besides `logging`)

Screenshot 2024-02-22 at 9 03 37 PM

### DIFF
--- a/charts/newrelic-logging/templates/fluent-bit-namespace-to-api-key-mapping.yml
+++ b/charts/newrelic-logging/templates/fluent-bit-namespace-to-api-key-mapping.yml
@@ -1,0 +1,13 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: fluent-bit-namespace-to-api-key-mapping
+  # TODO: do we need namespace and labels? I stole them from configmap.yml
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "newrelic-logging.labels" . | indent 4 }}
+data:
+  # TODO: this needs to be a Vault path instead of hard-coding keys
+  fluent-bit-namespace-to-api-key-mapping.json: |
+    {
+      "logging": "<REDACTED_KEY_FOR_PRODUCTION_756053>"
+    }
--- a/charts/newrelic-logging/templates/configcreator.yaml
+++ b/charts/newrelic-logging/templates/configcreator.yaml
@@ -1,0 +1,67 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: newrelic-logging-config-creator
+  # TODO: do we need namespace and labels like in configmap.yml?annotations:
+  annotations:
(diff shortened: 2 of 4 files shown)


charts/newrelic-logging/templates/fluent-bit-namespace-to-api-key-mapping.yml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: fluent-bit-namespace-to-api-key-mapping
+  # TODO: do we need namespace and labels? I stole them from configmap.yml
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "newrelic-logging.labels" . | indent 4 }}
+data:
+  # TODO: this needs to be a Vault path instead of hard-coding keys
```
bmcfeely, line 9: I'm pretty sure that we have a way to get secrets from Vault into k8s secrets so that they can be used. I'm assuming that's possible, I'm skipping it here just so I can get done with the prototype sooner.

bmcfeely:
This didn't get any traction, so closing it

PR #1304, opened 2024-03-05
[newrelic-logging] Adds newrelic-logging.sendMetrics option to enable the collection of Fluent Bit and newrelic-fluent-bit-output metrics

jsubirat:
#### Is this a new chart
No

This PR:
- Enables the collection of low level Fluent Bit metrics if `fluentBit.sendMetrics` is set to `true` by using the `prometheus_scrape` INPUT plugin and subsequent forwarding of these metrics to New Relic using the `prometheus_remote_write` OUTPUT plugin. These metrics are exposed by Fluent Bit at `/api/v2/metrics/prometheus`.
- Little refactor to compute the `logsEndpoint` and `metricsHost` (this one needs to be just the Host, because of how Fluent Bit `prometeus_scrape` plugin is configured) in the `_helpers.tpl` file instead, to avoid code duplication and keep the DaemonSet files shorter.
- Always include `NODE_NAME` (node where each newrelic-logging pod is running) and `HOSTNAME` (newrelic-logging pod hostname) environment variables, so they can be included as dimensions in the forwarded metrics.
- Adds unit test for the endpoint/host resolution depending on the region, as well as checking that all the above environment variables keep getting set.
- Adds `Alias` for all the INPUT/FILTER/OUTPUT plugin instances, so they can be easily identified later on using the `name` dimension of the metrics reported by Fluent Bit.
- Tested on Linux and Windows nodes

None

#### Checklist
- [x] Chart Version bumped


danybmx (approved):
💯👍🏻

danybmx (approved):

PR #1319, opened 2024-03-25
[super-agent-deployment] Removing volume mounts

paologallinaharbur:
#### Is this a new chart
no

This PR depends on https://github.com/newrelic/newrelic-super-agent/pull/525 and is used there already for e2e test, we should not merge it for now


marcsanmi (approved):
LGTM 🚀

Issue #1324, opened 2024-04-04
Support `customSecretName` for `nri-statsd`


RoseSecurity:
### Description

Right now, the `nri-statsd` chart requires the API key to be hardcoded in the `values.yaml` or injected during the `helm install`, but it would be better if a `customSecretName` could be utilized to reference the secret location like the other `nri-bundle` elements utilize.

### Acceptance Criteria

Before:

```yaml
global:
   insightsKey: ""
```

After:

```yaml
global:
  customSecretName: nri-statsd-key
```

### Implementation

- Maybe something like this for the `_helpers.tpl`:

```yaml
{{/*
Return the customSecretName where the New Relic Insights API key is being stored
*/}}
{{- define "nri-statsd.customSecretName" -}}
{{- if .Values.global }}
  {{- if .Values.global.customSecretName }}
      {{- .Values.global.customSecretName -}}
  {{- else -}}
      {{- .Values.customSecretName | default "" -}}
  {{- end -}}
{{- else -}}
    {{- .Values.customSecretName | default "" -}}
{{- end -}}
{{- end -}}
```

At the bottom:

```yaml
{{/*
Returns if the template should render, it checks if the required values
insightsKey or customSecretName and cluster are set.
*/}}
{{- define "nri-statsd.areValuesValid" -}}
{{- $cluster := include "nri-statsd.cluster" . -}}
{{- $insightsKey := include "nri-statsd.insightsKey" . -}}
{{- $customSecretName := include "nri-statsd.customSecretName" . -}}
{{- and $cluster (or $insightsKey $customSecretName) }}
{{- end -}}
```

Then update the configmap to use either value

PR #1327, opened 2024-04-05
`nri-statsd`: Support `customSecretName` Statsd Chart

RoseSecurity:
# What and Why

> [!IMPORTANT]
> I am not deeply familiar with the internals of this chart, so I utilized the values depicted in `newrelic-pixie` to concoct this

### Acceptance Criteria

Before:

```yaml
global:
   insightsKey: ""
```

After:

```yaml
global:
```

## Checks

- [X] Read contributing guidelines
- [X] Chart Version bumped
- [X] Support [this issue](https://github.com/newrelic/helm-charts/issues/1324)

### DIFF
--- a/charts/nri-bundle/Chart.lock
+++ b/charts/nri-bundle/Chart.lock
@@ -1,7 +1,7 @@
 dependencies:
 - name: newrelic-infrastructure
   repository: https://newrelic.github.io/nri-kubernetes
-  version: 3.33.0
+  version: 3.33.1
 - name: nri-prometheus
   repository: https://newrelic.github.io/nri-prometheus
   version: 2.1.17
@@ -25,12 +25,12 @@ dependencies:
   version: 1.21.2
 - name: newrelic-pixie
   repository: https://newrelic.github.io/helm-charts
-  version: 2.1.3
+  version: 2.1.4
 - name: pixie-operator-chart
   repository: https://pixie-operator-charts.storage.googleapis.com
   version: 0.1.4
 - name: newrelic-infra-operator
   repository: https://newrelic.github.io/newrelic-infra-operator
   version: 2.10.0
-digest: sha256:384442c773d58803b9e3552e5152b38ee41263e7eadc1546739d144c30216ee4
-generated: "2024-03-25T12:36:06.208840149Z"
+digest: sha256:dc652de5a270b79377130dd333bfd50284e34541cdfd32d4fbce0268f3c01c99
+generated: "2024-04-01T14:46:40.056684873Z"
--- a/charts/nri-bundle/Chart.yaml
+++ b/charts/nri-bundle/Chart.yaml
@@ -16,13 +16,13 @@ sources:
   - https://github.com/newrelic/helm-charts/tree/master/charts/newrelic-pixie
   - https://github.com/newrelic/newrelic-infra-operator/tree/master/charts/newrelic-infra-operator
 
-version: 5.0.71
+version: 5.0.72
 
 dependencies:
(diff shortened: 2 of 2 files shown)


RoseSecurity:
Just signed the CLA and would love to see how we could make this feature possible @bambash

DavSanchez:
Hello @RoseSecurity !

Just took a look and I have some questions. The way it's defined, I see the `customSecretName` would be like an additional way to provide an API key and place it into `nri-statsd.toml`. This `customSecretName` or `global.customSecretName` would contain the API key directly as a string in your `values.yml`. However, from your comment it seems that what you really want is to reference a Secret (the K8s resource) by name that is already present in a K8s cluster. See [this function](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_license_secret.yaml.tpl) for an example and this [daemonset template](https://github.com/newrelic/helm-charts/blob/2f5c8d36c2774413469d2aca714777113f01ed5b/charts/newrelic-logging/templates/daemonset.yaml#L83) to see how to use it.

Can you confirm if the latter is what you wanted to achieve?

DavSanchez (commented):
Hello @RoseSecurity !

Just took a look and I have some questions. The way it's defined, I see the customSecretName would be like an additional way to provide an API key and place it into nri-statsd.toml. This customSecretName or global.customSecretName would contain the API key directly as a string in your values.yml. However, from your comment it seems that what you really want is to reference a Secret (the K8s resource) by name that is already present in a K8s cluster. See [this function](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_license_secret.yaml.tpl) for an example and this [daemonset template](https://github.com/newrelic/helm-charts/blob/2f5c8d36c2774413469d2aca714777113f01ed5b/charts/newrelic-logging/templates/daemonset.yaml#L83) to see how to use it.

Can you confirm if the latter is what you wanted to achieve?

RoseSecurity:
> Hello @RoseSecurity !
> 
> Just took a look and I have some questions. The way it's defined, I see the customSecretName would be like an additional way to provide an API key and place it into nri-statsd.toml. This customSecretName or global.customSecretName would contain the API key directly as a string in your values.yml. However, from your comment it seems that what you really want is to reference a Secret (the K8s resource) by name that is already present in a K8s cluster. See [this function](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_license_secret.yaml.tpl) for an example and this [daemonset template](https://github.com/newrelic/helm-charts/blob/2f5c8d36c2774413469d2aca714777113f01ed5b/charts/newrelic-logging/templates/daemonset.yaml#L83) to see how to use it.
> 
> Can you confirm if the latter is what you wanted to achieve?

Yes, those functions you provided are exactly what I was trying to achieve! If it would be an easy implementation, feel free to go for it

RoseSecurity:
@DavSanchez Is there any way you would be able to add the funtionality you noted here:

> Hello @RoseSecurity !
> 
> However, from your comment it seems that what you really want is to reference a Secret (the K8s resource) by name that is already present in a K8s cluster. See [this function](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_license_secret.yaml.tpl) for an example and this [daemonset template](https://github.com/newrelic/helm-charts/blob/2f5c8d36c2774413469d2aca714777113f01ed5b/charts/newrelic-logging/templates/daemonset.yaml#L83) to see how to use it.

RoseSecurity:
@DavSanchez Any thoughts on how feasible it would be to incorporate this functionality?

PR #1342, opened 2024-04-23
[nr-k8s-otel-collector] Add K8s Otel Collector

dbudziwojskiNR:
#### Is this a new chart
Yes
Add K8s Otel Collector.
  - fixes #

#### Checklist
- [ ] Chart Version bumped


charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -0,0 +1,100 @@
...
+    # kubernetes.io/tls-acme: "true"
+  hosts:
+    - host: chart-example.local
+      paths:
+        - path: /
+          pathType: ImplementationSpecific
+  tls: []
+  #  - secretName: chart-example-tls
+  #    hosts:
+  #      - chart-example.local
```
paologallinaharbur, line 75: why do we have an ingress?

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -0,0 +1,100 @@
...
+  # We usually recommend not to specify default resources and to leave this as a conscious
+  # choice for the user. This also increases chances charts run on environments with little
+  # resources, such as Minikube. If you do want to specify resources, uncomment the following
+  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+  # limits:
+  #   cpu: 100m
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
```
paologallinaharbur, line 87: my advice is avoiding having the very same resources for the daemonSet and for the deployment. Fixing it in the feature could be complex

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -0,0 +1,100 @@
...
+podSecurityContext: {}
+  # fsGroup: 2000
+
+securityContext: {}
+  # capabilities:
+  #   drop:
+  #   - ALL
+  # readOnlyRootFilesystem: true
+  # runAsNonRoot: true
+  # runAsUser: 1000
```
paologallinaharbur, line 55: same here, we assume to share these, but we have two components. I would suggest to check nri-kubernetes to see how you could configure two components in different ways

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -0,0 +1,100 @@
...
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+
+autoscaling:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 100
+  targetCPUUtilizationPercentage: 80
+  # targetMemoryUtilizationPercentage: 80
```
paologallinaharbur, line 94: Autoscaling?

charts/nr-k8s-otel-collector/templates/clusterrole.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - list
+    - watch
+  - apiGroups:
+      - autoscaling
+    resources:
+      - horizontalpodautoscalers
+    verbs:
+      - get
+      - list
+      - watch
```
paologallinaharbur, line 66: This should not be needed

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,289 @@
+apiVersion: v1 
+kind: ConfigMap
+metadata:
+  name: {{ include "newrelic.common.naming.fullname" . }}-daemonset-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "newrelic.common.labels" . | nindent 4 }}
+data:
+  otel-collector-config.yaml: |
```
paologallinaharbur, line 9: How the user can change this? Can we expose the config in the values.yaml having all of this as a default?

You can get inspired by the community chart to see how they implemented "set of features"

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,289 @@
...
+            #       source_labels:
+            #         - __meta_kubernetes_pod_label_app_kubernetes_io_name
+            #     - action: replace
+            #       source_labels:
+            #         - __meta_kubernetes_pod_node_name
+            #       target_label: instance
+            #     - action: replace
+            #       source_labels:
+            #         - __meta_kubernetes_namespace
+            #       target_label: namespace
```
paologallinaharbur, line 119: I think this is a leftover.
I do not see how a user could change this

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,289 @@
...
+    exporters:
+      otlphttp/newrelic:
+        endpoint: {{ .Values.newRelicEndpoint }}
+        headers:
+          api-key: ${env:NR_LICENSE_KEY}
+      # uncomment when writing daemonset metrics through deployment
+      # otlp:
+      #   endpoint: "nr-k8s-otel-collector-gateway:4317"
+      #   tls:
+      #     insecure: true
```
paologallinaharbur, line 259: how the user could change this? This config is not exposed

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,289 @@
...
+      logging:
```
paologallinaharbur, line 260: I think that the logging exporter has been deprecated https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/loggingexporter

paologallinaharbur: > This exporter is being deprecated in favour of the [debug exporter](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/debugexporter/README.md). It will be removed in September 2024.

charts/nr-k8s-otel-collector/templates/NOTES.txt
paologallinaharbur, line 1: I suggest as well to add unit tests. Maintaining chart without them is really painful

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,289 @@
...
+            - filter/exclude_network
+            - attributes/exclude_system_paging
+            - resource
+            - k8sattributes
+            - batch
+          exporters: [otlphttp/newrelic, logging]
+        logs:
+          receivers: [otlp, filelog]
+          processors: [transform/truncate, resource, k8sattributes, batch]
+          exporters: [logging, otlphttp/newrelic]
```
paologallinaharbur, line 289: isn't there the risk of creating a loop having both a filelog receiver and the logging exporter? 🤔

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,289 @@
...
+      logging:
+        verbosity: detailed
+
+    service:
+      telemetry:
+        logs:
+          level: "debug"
+      pipelines:
+        metrics:
+          receivers: [otlp, hostmetrics, kubeletstats, prometheus]
```
paologallinaharbur, line 269: why otlp here?

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,289 @@
...
+            - filter/exclude_filesystem_utilization
+            - filter/exclude_filesystem_usage
+            - filter/exclude_filesystem_inodes_usage
+            - filter/exclude_system_disk
+            - filter/exclude_system_paging
+            - filter/exclude_network
+            - attributes/exclude_system_paging
+            - resource
+            - k8sattributes
+            - batch
```
paologallinaharbur, line 284: if we have a unique gateway why are we have filters replicated on all instances? Should we have them in the gateway only?

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -0,0 +1,334 @@
...
+        tls:
+          insecure: true
+      otlphttp/newrelic:
+        endpoint: {{ .Values.newRelicEndpoint }}
+        headers:
+          api-key: ${env:NR_LICENSE_KEY}
+      debug:
+        verbosity: detailed
+        sampling_initial: 5
+        sampling_thereafter: 200
```
paologallinaharbur, line 306: why the debug is enabled by default?

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -0,0 +1,334 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "newrelic.common.naming.fullname" . }}-deployment-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "newrelic.common.labels" . | nindent 4 }}
+data:
+  deployment-config.yaml: |
```
paologallinaharbur, line 9: same for the daemonset, the user should be able to customise this (not sure if we want it in the first iteration)

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -0,0 +1,100 @@
+# Default values for nr-k8s-otel-collector.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+kube-state-metrics:
+  # kube-state-metrics.enabled -- Install the [`kube-state-metrics` chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-state-metrics) from the stable helm charts repository.
+  # This is mandatory if `infrastructure.enabled` is set to `true` and the user does not provide its own instance of KSM version >=1.8 and <=2.0. Note, kube-state-metrics v2+ disables labels/annotations
+  # metrics by default. You can enable the target labels/annotations metrics to be monitored by using the metricLabelsAllowlist/metricAnnotationsAllowList options described [here](https://github.com/prometheus-community/helm-charts/blob/159cd8e4fb89b8b107dcc100287504bb91bf30e0/charts/kube-state-metrics/values.yaml#L274) in
+  # your Kubernetes clusters.
+  enabled: true
```
paologallinaharbur, line 10: if we are aiming to include this under the nri-bundle we could have some issues since KSM is already a dependency and it could create some confusion 🤔

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+replicaCount: 1
```
paologallinaharbur, line 12: This seems to be unused

charts/nr-k8s-otel-collector/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "nr-k8s-otel-collector.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create }}
+{{- default (include "nr-k8s-otel-collector.fullname" .) .Values.serviceAccount.name }}
+{{- else }}
+{{- default "default" .Values.serviceAccount.name }}
+{{- end }}
+{{- end }}
```
paologallinaharbur, line 62: This function is not used

charts/nr-k8s-otel-collector/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "nr-k8s-otel-collector.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "nr-k8s-otel-collector.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
```
paologallinaharbur, line 51: same here

charts/nr-k8s-otel-collector/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+{{- end }}
+{{- end }}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "nr-k8s-otel-collector.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
```
paologallinaharbur, line 31: also this one

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,289 @@
...
+  labels:
+    {{- include "newrelic.common.labels" . | nindent 4 }}
+data:
+  otel-collector-config.yaml: |
+    receivers:
+      otlp:
+        protocols:
+          grpc:
+          http:
+      hostmetrics:
```
paologallinaharbur, line 15: We might want to sync with @newrelic/caos to understand the metrics they are planning to scrape

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -0,0 +1,334 @@
...
+    service:
+      telemetry:
+        logs:
+          level: "debug"
+      pipelines:
+        # Uncomment below if exporting traces to NR
+        # traces:
+        #   receivers: [otlp]
+        #   processors: [batch]
+        #   exporters: [otlphttp/newrelic, debug]
```
paologallinaharbur, line 316: leftover?

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -0,0 +1,334 @@
...
+        pod_association:
+          - sources: 
+            - from: resource_attribute
+              name: k8s.pod.uid
+
+    exporters:
+      otlp:
+        endpoint: "nr-k8s-otel-collector-gateway:4317"
+        tls:
+          insecure: true
```
paologallinaharbur, line 298: I am confused, isn't this already the gateway?

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -0,0 +1,334 @@
...
+            - k8sattributes
+          exporters: [otlphttp/newrelic, debug]
+        logs:
+          receivers: [otlp]
+          processors: [batch]
+          exporters: [otlphttp/newrelic, debug]
+        logs/events:
+          receivers: [k8s_events]
+          processors: [resource/events, batch]
+          exporters: [otlp, debug]
```
paologallinaharbur, line 334: I have the feeling we are sending data to the same component sending the data 😅

charts/nr-k8s-otel-collector/templates/NOTES.txt
kang-makes: Also, the NOTES showed here is the default one made by `helm create` and shows information to connect to an example of a backend. It doesn't makes sense anymore and could break the installation while being templated.

Also, the `NOTES.txt` is not a needed file. If you do not want to show information to the user just delete it.

charts/nr-k8s-otel-collector/templates/clusterrole.yaml
kang-makes, line 1: Can we add some comments on which component needs each RBAC? This seems a broad one and customers tend to be scared of these permissions. More if they are not commented.

RBAC should also have a toggle to disable it. There are comanies that have strict security that ask RBAC to be managed by security teams.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,289 @@
...
+                - regex: (.+)
+                  replacement: /api/v1/nodes/$${1}/proxy/metrics
+                  source_labels:
+                    - __meta_kubernetes_node_name
+                  target_label: __metrics_path__
+              scheme: https
+              tls_config:
+                ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+                insecure_skip_verify: false
+                server_name: kubernetes
```
kang-makes, line 102: Doesn't this duplicate the kubeletstat receiver?

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,289 @@
...
+          process:
+            metrics:
+              process.cpu.utilization:
+                enabled: true
+              process.cpu.time:
+                enabled: false
+      kubeletstats:
+        collection_interval: 20s
+        auth_type: "serviceAccount"
+        endpoint: "${KUBE_NODE_NAME}:10250"
```
kang-makes, line 61: If you take a look at [line 94](https://github.com/newrelic/helm-charts/pull/1342/files#diff-b9dcfedad3cb705d7e2b97544a9c7802a11e8c32eb5113ed4ae65cbc7b176901R94) you'll see that the exporter goes to the kubelet using the API server as proxy in case there are firewall rules that forbid to connect directly to the node.

You should parametrize this even if the default is "`NODE_IP`".

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,289 @@
...
+          - /var/log/*
+          - /var/log/alternatives.log
+          - /var/log/cloud-init.log
+          - /var/log/auth.log
+          - /var/log/dpkg.log
+          - /var/log/syslog
+          - /var/log/messages
+          - /var/log/secure
+          - /var/log/yum.log
+          - /var/log/alternatives.log
```
kang-makes, line 133: These are not mounted in the pod, right?

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,289 @@
...
+        exclude:
+          # Exclude logs from all containers named nr-k8s-otel-collector and otelcontribcol
+          - /var/log/pods/*/nr-k8s-otel-collector/*.log
+          - /var/log/pods/*/otelcontribcol/*.log
```
kang-makes, line 137: These should be variable depending on chart release name, right?

Can this change depending on which CRI is in use?

charts/nr-k8s-otel-collector/templates/daemonset.yaml
```diff
@@ -0,0 +1,94 @@
...
+    metadata:
+      labels:
+        {{- include "newrelic.common.labels.selectorLabels" . | nindent 8 }}
+        component: daemonset
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/daemonset-configmap.yaml") . | sha256sum }}
+    spec:
+      serviceAccountName: {{ include "newrelic.common.serviceAccount.name" . }}
+      containers: 
+        - name: {{ .Chart.Name }}
```
kang-makes, line 23: You might not have this name variable so, in the docs, you will have a constant container name where to run a troubleshoot guide.

charts/nr-k8s-otel-collector/templates/deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+                  apiVersion: v1
+                  fieldPath: status.podIP
+            - name: NR_LICENSE_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "newrelic.common.license.secretName" . }}
+                  key: {{ include "newrelic.common.license.secretKeyName" . }}
+          ports:
+            - containerPort: 4317   # otlp grpc
+            - containerPort: 4318   # otlp http
```
kang-makes, line 49:
```suggestion
            - name:  grpc
              containerPort: 4317
              protocol: TCP
            - name:  otlp
              containerPort: 4318
              protocol: TCP
```

kang-makes (commented):
Here you have a first round of comments :)

charts/nr-k8s-otel-collector/values.yaml
csongnr: Because I used the helm create command 😂 this should be removed, it was my first time creating a helm chart and I may have left a lot of things in it that weren't strictly necessary but I didn't have time to go through and untangle all the helper methods. I'll talk w/ daniel about removing the extra config in this or a followup pr. I'll refer to this comment as `remove extra helm create config` for the other comments that are related to this same issue.

charts/nr-k8s-otel-collector/values.yaml
csongnr: Yeah, we will have to do a load test to confirm these values after we cut down on the amount of metrics being sent over, but can you clarify why fixing the feature could be complex?

charts/nr-k8s-otel-collector/values.yaml
csongnr: `remove extra helm create config`

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
csongnr: We are currently working on a followup PR that will `expose the config in values.yaml`, which won't be included in this PR

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
csongnr: `expose the config in values.yaml in followup pr`

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
csongnr: `expose the config in values.yaml in followup pr`

charts/nr-k8s-otel-collector/values.yaml
paologallinaharbur: Right now we have a deployment A and adeamonSet B

They share the same `resources `, however in a cluster with 200 nodes the gateway could end up consuming 4-5x than a single instance on a node since it receive and process the traffic coming from all over the cluster (and scrapes KSM)
For this reason, I think that is safer to have most of the option under the specific component:
```
depoyment :   <-- not sure about the name 😄 
  enabled: {}
  podannotations: {}
  extraEvnVars: {}
  resources: {}
  [...]
daemonSet:
  enabled: {}
  podannotations: {}
  extraEvnVars: {}
  resources: {}
  [...]
```
Similar to what we have already [here](https://github.com/newrelic/nri-kubernetes/blob/main/charts/newrelic-infrastructure/values.yaml)

Otherwise you could endup giving a lot of resources to all instances, just because the single one from the deployment needs more

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
csongnr: right now, because of memory issues, while the metrics from daemonset _should_ be routed through the gateway, it currently is not, the metrics from the daemonset directly write to newrelic, as well as the metrics from the deployment. 

the filters are only on the daemonset because they refer to the metrics that flow through the daemonset. after i finish up w/ ingest summary and cutting down on the number of metrics, we'll see if we're still having some memory issues. 

this will be adjusted in a followup PR `todo: chris use deployment gateway after data ingest`

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
csongnr: something to remove

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
csongnr: `expose the config in values.yaml in followup pr`

charts/nr-k8s-otel-collector/values.yaml
csongnr: We are not currently aiming to include this in nri-bundle

charts/nr-k8s-otel-collector/templates/_helpers.tpl
csongnr: `remove extra helm create config`

charts/nr-k8s-otel-collector/templates/_helpers.tpl
csongnr: `remove extra helm create config`

charts/nr-k8s-otel-collector/README.md
```diff
@@ -0,0 +1,63 @@
...
+#### [Optional] Enable node-exporter (not required for New Relic Kubernetes monitoring experience) 
+1. Run: 
+```
+helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+helm repo update
+helm install nodeexporter prometheus-community/prometheus-node-exporter 
+```
+2. Comment out [these lines](https://github.com/newrelic/helm-charts/tree/master/charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml#L277-L292) in the configuration. 
+
+
```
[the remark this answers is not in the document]
csongnr: 1. I think we can remove the node exporter totally from this actually, its unnecessary for the config
2. `expose the config in values.yaml in followup pr`

charts/nr-k8s-otel-collector/templates/clusterrole.yaml
csongnr: yep, sounds good, will add comments on which components need RBAC in this PR  
we'll add the RBAC toggle in a followup PR

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
csongnr: Kubeletstats receiver is going to be removed in favor of this receiver in a followup PR

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
paologallinaharbur: usually we try to send the very same metrics so that the experience is similar. Moreover, they can save you some time, since they have done already some research

charts/nr-k8s-otel-collector/values.yaml
dbudziwojskiNR: Fixed.

charts/nr-k8s-otel-collector/values.yaml
dbudziwojskiNR: Fixed

charts/nr-k8s-otel-collector/Chart.yaml
```diff
@@ -0,0 +1,48 @@
...
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
```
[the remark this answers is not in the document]
dbudziwojskiNR: Fixed

charts/nr-k8s-otel-collector/templates/clusterrole.yaml
dbudziwojskiNR: We received errors when we tried to remove this so we've left it in.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
dbudziwojskiNR: Fixed

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
dbudziwojskiNR: Fixed by observing the verbose setting in the common library.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,289 @@
...
+      #   endpoint: "nr-k8s-otel-collector-gateway:4317"
+      #   tls:
+      #     insecure: true
+      logging:
+        verbosity: detailed
+
+    service:
+      telemetry:
+        logs:
+          level: "debug"
```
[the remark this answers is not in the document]
dbudziwojskiNR: Fixed by observing the verbose setting in the common library.

charts/nr-k8s-otel-collector/templates/NOTES.txt
dbudziwojskiNR: Removed :)

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
dbudziwojskiNR: Fixed

charts/nr-k8s-otel-collector/values.yaml
dbudziwojskiNR: Fixed

charts/nr-k8s-otel-collector/templates/_helpers.tpl
dbudziwojskiNR: Removed this file.

charts/nr-k8s-otel-collector/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+Common labels
+*/}}
+{{- define "nr-k8s-otel-collector.labels" -}}
+helm.sh/chart: {{ include "nr-k8s-otel-collector.chart" . }}
+{{ include "nr-k8s-otel-collector.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
```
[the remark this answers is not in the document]
dbudziwojskiNR: Removed this file.

charts/nr-k8s-otel-collector/templates/_helpers.tpl
dbudziwojskiNR: Removed this file.

charts/nr-k8s-otel-collector/templates/_helpers.tpl
dbudziwojskiNR: Removed this file.

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
dbudziwojskiNR: Removed.

charts/nr-k8s-otel-collector/templates/daemonset.yaml
dbudziwojskiNR: Done! TY

charts/nr-k8s-otel-collector/templates/deployment.yaml
dbudziwojskiNR: Updated :)

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
csongnr: no, none of these are mounted in the pod. these paths point to file systems located on the node

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
csongnr: we actually ended up hardcoding the container name, so I updated these to reflect the hardcoded container name

thanks for the headsup on the filelog paths changing based on CRI. I added the Filelog paths for the 3 major CRI, containerd, cri-o, and docker; i figured that should be good for now, couldn't figure out how to generate the filelog paths dynamically based on the CRI, but tested and the collector won't break if the filepaths listed aren't discovered, it'll just log a message and move on w/ the filepaths it can discover.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
csongnr: this was updated, to remove logging exporter since it is deprecated, but could you expand a bit more on this? the only reason I could think of this creating a loop was if the logging exporter exported logs to a log on the filesystem and then the filelog receiver would write those logs back out, but it was my understanding that the logging exporter only wrote logs out to the console, and it wasn't ever persisted/stored or written anywhere

charts/nr-k8s-otel-collector/templates/clusterrole.yaml
csongnr: this is needed to scrape horizontalpodautoscaler metrics from KSM

charts/nr-k8s-otel-collector/values.yaml
csongnr: hmm, that makes a lot of sense, thanks. I'll do this in a followup PR as we work through load and ingest testing.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
csongnr: I actually referenced a lot of the hostmetrics settings from NRDOT when I was pointed there by david gay, https://github.com/newrelic/opentelemetry-collector-releases/blob/main/configs/nr-otel-collector-agent-linux.yaml 
is this what you mean by the CAOS team?

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
csongnr: are you confused by the naming of that endpoint being gateway when its on the deployment (gateway) config? or why this endpoint exists?

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
csongnr: Hmmm, let me circle back to you on this

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
csongnr: can you clarify what you mean by 'parametrize this even if the default is "NODE_IP"' ? 

note: we are trying to remove kubeletstats in a followup PR, but I would appreciate the clarification regardless for my understanding :)

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
paologallinaharbur: yes!

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
paologallinaharbur: I mean the following:
 - this is the config of the deployment.
 - this OTLP exporter is sending data to `nr-k8s-otel-collector-gateway:4317`
 - `nr-k8s-otel-collector-gateway` is a service backed up by this deployment 🤔

charts/nr-k8s-otel-collector/templates/deployment.yaml
```diff
@@ -0,0 +1,72 @@
...
+                secretKeyRef:
+                  name: {{ include "newrelic.common.license.secretName" . }}
+                  key: {{ include "newrelic.common.license.secretKeyName" . }}
+          ports:
+            - name: grpc
+              containerPort: 4317
+              protocol: TCP
+            - name: http
+              containerPort: 4318
+              protocol: TCP
```
paologallinaharbur, line 53: I think that only the 4317 is used (even though it is used in a part of code currently commented out)

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+kube-state-metrics:
+  # kube-state-metrics.enabled -- Install the [`kube-state-metrics` chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-state-metrics) from the stable helm charts repository.
+  # This is mandatory if `infrastructure.enabled` is set to `true` and the user does not provide its own instance of KSM version >=1.8 and <=2.0. Note, kube-state-metrics v2+ disables labels/annotations
+  # metrics by default. You can enable the target labels/annotations metrics to be monitored by using the metricLabelsAllowlist/metricAnnotationsAllowList options described [here](https://github.com/prometheus-community/helm-charts/blob/159cd8e4fb89b8b107dcc100287504bb91bf30e0/charts/kube-state-metrics/values.yaml#L274) in
+  # your Kubernetes clusters.
+  enabled: true
+
+image:
+  repository: otel/opentelemetry-collector-contrib
```
paologallinaharbur, line 13: should we use nrdot?

charts/nr-k8s-otel-collector/README.md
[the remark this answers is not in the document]
dbudziwojskiNR: We went ahead and removed this exporter.

charts/nr-k8s-otel-collector/values.yaml
csongnr: Can you clarify what you mean by using nrdot? I'm assuming this means nrdot has a distribution of otel that theyre using, and if thats the case i'd have to confirm that theyre including all of the processors/receivers/exporters that we're using, but its probably not strictly necessary for what we're trying to achieve at this time w/ this pr.

charts/nr-k8s-otel-collector/templates/deployment.yaml
csongnr: We wanted to leave the option available

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
csongnr: updated

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
csongnr: yes, this would have then sent the data through to the logs pipeline to write to NR, so that any changes that had to be made to _all_ the telemetry being sent through the gateway would go through that single pipeline, but we since had to make some changes to the structure of the collector because of memory issues (as in, right now all the data goes directly from the deployment and daemonset to NR without running through the other collectors) so now that paradigm of all the collected data running through a single pipeline that adds transforms needed on all the data is confusing; we have updated it. thanks :D

charts/nr-k8s-otel-collector/values.yaml
paologallinaharbur: I mean [this](https://github.com/newrelic/opentelemetry-collector-releases/tree/main/distributions/nr-otel-collector) I think @mangulonr can take a final decision on that.
But for sure receivers are missing

mangulonr: We need to provide NRDOT as the default option in the Helm Chart.
NRDOT components are in this [manifest](https://github.com/newrelic/opentelemetry-collector-releases/blob/main/distributions/nr-otel-collector/manifest.yaml), I assume we need what is needed to properly monitor a K8s cluster.

As part of our documentation, we will document how to use a 3rd party OTel collector if required by our customers (including config and components needed).

@csongnr let's sync in our today's meeting

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
paologallinaharbur: aren't we getting the logs of the pods as well with the log receiver?

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,309 @@
...
+          - /var/log/alternatives.log
+          - /var/log/cloud-init.log
+          - /var/log/auth.log
+          - /var/log/dpkg.log
+          - /var/log/syslog
+          - /var/log/messages
+          - /var/log/secure
+          - /var/log/yum.log
+        exclude:
+          # Exclude logs from opentelemetry containers
```
paologallinaharbur, line 120: why?
how a customer could check the logs from the monitoring?

If it is to avoid "creating a loop" I think the way to go is not exporting metrics to logs.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,309 @@
...
+          api-key: ${env:NR_LICENSE_KEY}
+      # TODO(chris): uncomment when writing daemonset metrics through deployment
+      # otlphttp:
+      #   endpoint: "nr-k8s-otel-collector-gateway"
+      #   tls:
+      #     insecure: true
+      debug:
+        verbosity: detailed
+        sampling_initial: 5
+        sampling_thereafter: 200
```
paologallinaharbur, line 270: This could burn a lot of user money in case they export logs automatically (for example in GCP)

It has happened to our team as well 😅

paologallinaharbur: as ok, it is behind `{{ if include "newrelic.common.verboseLog" . }}`

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
paologallinaharbur: actually I have some doubts about this 🤔 Not an easy choice!

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,309 @@
...
+            {{ if include "newrelic.common.verboseLog" . }}
+            - debug
+            {{ end }}
+        logs:
+          receivers: [filelog]
+          processors: [transform/truncate, resource, k8sattributes, batch]
+          exporters:
+            - otlphttp/newrelic
+            {{ if include "newrelic.common.verboseLog" . }}
+            - debug
```
paologallinaharbur, line 308: I think that this is the thing creating the loop. Having debug elsewhere could be really expensive, but somehow useful, having it here creates possibly a loop if we scrape otel logs

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,309 @@
...
+            - filter/exclude_system_paging
+            - filter/exclude_network
+            - attributes/exclude_system_paging
+            - resource
+            - k8sattributes
+            - attributes/self
+            - batch
+          exporters:
+            - otlphttp/newrelic
+            {{ if include "newrelic.common.verboseLog" . }}
```
paologallinaharbur, line 299: I am not sure if this would be "verbose" or even "trace". For example the agent does not print metrics in debug mode

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
kang-makes: Take a look at the connector of `nri-kubernetes`: https://github.com/newrelic/nri-kubernetes/blob/a4bdffd9d96b446c4fa7ba8351f981656b3b091d/src/kubelet/client/connector.go#L56-L91

There are cases were security policies do not allow a pod to connect to the node directly to scrape Kubelet. So the connector fallbacks from using `$NODE_IP:10250` to `$APISERVER/api/v1/nodes/%s/proxy/`.

This is why the endpoint should be parametric somehow.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
kang-makes: If there is not `hostPath` mounting these paths from the host to the pod, the collector will not be able to get logs from the host. Am I right?

charts/nr-k8s-otel-collector/templates/daemonset.yaml
```diff
@@ -0,0 +1,98 @@
...
+                  fieldPath: spec.nodeName
+            - name: OTEL_RESOURCE_ATTRIBUTES
+              value: service.instance.id=$(POD_NAME),k8s.pod.uid=$(POD_UID)
+            - name: NR_LICENSE_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "newrelic.common.license.secretName" . }}
+                  key: {{ include "newrelic.common.license.secretKeyName" . }}
+          volumeMounts:
+            # TODO (chris): Comment out below when not linux
```
paologallinaharbur, line 61: do you mean on Windows nodes?

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -0,0 +1,383 @@
...
+    {{- include "newrelic.common.labels" . | nindent 4 }}
+data:
+  deployment-config.yaml: |
+    receivers:
+      otlp:
+        protocols:
+          {{ if .Values.enableGrpcGatewayEndpoint }}
+          grpc:
+            endpoint: ${env:MY_POD_IP}:4317
+          {{ end }}
```
paologallinaharbur, line 16: I would remove all extra functionalities that are not required for now. Otherwise, you will need to support them forever.

(this unless it is a requirement from product 😄 )

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -0,0 +1,383 @@
...
+            action: upsert 
+            from_attribute: namespace
+
+      batch:
+
+    exporters:
+      #  otlphttp:
+      #    endpoint: "nr-k8s-otel-collector-gateway"
+      #    tls:
+      #      insecure: true
```
paologallinaharbur, line 333: I think this could be deleted

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -0,0 +1,61 @@
+# Default values for nr-k8s-otel-collector.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+kube-state-metrics:
```
paologallinaharbur, line 5: I would suggest to remove the label enabling the autodiscovery from prometheus. We have talked about it countless of times for nri-bundle and we were able to do so to avoid breaking changes.

paologallinaharbur: https://github.com/newrelic/helm-charts/pull/376

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -0,0 +1,61 @@
...
+  #   memory: 128Mi
+
+nodeSelector: {}
+
+tolerations: []
+
+affinity: {}
+
+enableGrpcGatewayEndpoint: false
+verboseLog: false
```
paologallinaharbur, line 61: having a specific value here could overwrite the "global." if you ever put this chart under the nri-bundle.

For example the nri-kubernetes specify it like:
```
# -- (bool) Sets the debug logs to this integration or all integrations if it is set globally. Can be configured also with `global.verboseLog`
# @default -- `false`
verboseLog:
```

charts/nr-k8s-otel-collector/templates/service.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: nr-k8s-otel-collector-gateway
```
paologallinaharbur, line 4: Do we really need a gateway? I fear that it could introduce a bottleneck and a component consuming 10x the memory and cpu.

Which is the down side against having each replica sending data autonomously?

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
kang-makes: I am marking this al solved. I might open a new comment on a second review.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
kang-makes: The choice is easy. You should not duplicate or use the same thing for input and output. As it is a daemonset and it has to mount a folder with the logs files to be sent to the NR. You want to send the data of the collector to NR too.

 - The pod should not write to a file, all the logging should to to stdout.
 - The collector should send every pod in the cluster. Itself included.
 - Do not log the logging pipeline so there is no loop. (It will also duplicate ALL the logs)
 - In case debug logs are needed, create a new log pipeline that includes all the debug information without loop possibility.

charts/nr-k8s-otel-collector/templates/daemonset.yaml
```diff
@@ -0,0 +1,98 @@
...
+        - name: otel-collector-daemonset
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command:
+            - /otelcol-contrib
+            - --config=/conf/otel-collector-config.yaml
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
```
kang-makes, line 32: The reason behind the `CHART-TEMPLATE` chart was not only to have a testing for `common-library` but also to have a template to create new charts instead of `helm create`.

There are still a lot of features from the common library that are not implemented here: https://github.com/newrelic/helm-charts/blob/162de8eac4822b633ad4e08ff6660035a677c328/library/CHART-TEMPLATE/templates/sidecar-deployment.yaml#L21-L44

charts/nr-k8s-otel-collector/templates/daemonset.yaml
```diff
@@ -0,0 +1,98 @@
...
+              mountPath: /var/log/pods
+              readOnly: true
+            - name: varlibdockercontainers
+              mountPath: /var/lib/docker/containers
+              readOnly: true
+      volumes:
+        # TODO (chris): Comment out below when not linux
+        - name: host-fs
+          hostPath:
+            path: /
```
kang-makes, line 77: There is no way to "comment out" something depending on the OS.

You'll need two daemonsets having different selectors:
https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-os

charts/nr-k8s-otel-collector/templates/daemonset.yaml
```diff
@@ -0,0 +1,98 @@
...
+    {{- include "newrelic.common.labels" . | nindent 4 }}
+spec: 
+  selector:
+    matchLabels:
+      {{- include "newrelic.common.labels.selectorLabels" . | nindent 6 }}
+      component: daemonset
+  template:
+    metadata:
+      labels:
+        {{- include "newrelic.common.labels.selectorLabels" . | nindent 8 }}
```
kang-makes, line 16:
```suggestion
        {{- include "newrelic.common.labels.podLabels" . | nindent 8 }}
```

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
kang-makes: If there is a follow up PR to improve the configuration part, I would left this for that PR.

There are still lots of things that are not implemented (like the `LowDataMode`) and too many configurations that need to change depending if they are running on `hostNetwork` mode (for example).

These kind of toogles that are not in a tree (`.Values.enableGrpcGatewayEndpoint` instead of `.Values.gateways.grpc.enable`) are an anti-pattern and prone to be HARD to maintain.

Example:
https://github.com/newrelic/nri-kubernetes/blob/a4bdffd9d96b446c4fa7ba8351f981656b3b091d/charts/newrelic-infrastructure/templates/_helpers_compatibility.tpl#L10-L30
That is used here:
https://github.com/newrelic/nri-kubernetes/blob/a4bdffd9d96b446c4fa7ba8351f981656b3b091d/charts/newrelic-infrastructure/templates/ksm/scraper-configmap.yaml#L14

charts/nr-k8s-otel-collector/templates/deployment.yaml
```diff
@@ -0,0 +1,72 @@
...
+  replicas: 1
+  minReadySeconds: 5
+  selector:
+    matchLabels:
+      {{- include "newrelic.common.labels.selectorLabels" . | nindent 6 }}
+      component: deployment
+  template:
+    metadata:
+      labels:
+        {{- include "newrelic.common.labels.selectorLabels" . | nindent 8 }}
```
kang-makes, line 18:
```suggestion
        {{- include "newrelic.common.labels.podsLabels" . | nindent 8 }}
```

charts/nr-k8s-otel-collector/templates/deployment.yaml
kang-makes: You should not open a port if there is no reason to keep it open. If you want to open a port to listen to GRPC OTLP, create a toogle in the `values.yaml` file that enables the port and the pipelines in the collector.

The user should not configure this kind of things on its own. Anything that is out of the `values.yaml` should be unsupported (if you want to be sane in a couple of years).

charts/nr-k8s-otel-collector/templates/deployment.yaml
```diff
@@ -0,0 +1,72 @@
...
+            - name: http
+              containerPort: 4318
+              protocol: TCP
+          volumeMounts:
+            - name: deployment-config
+              mountPath: /config
+      volumes:
+        - name: deployment-config
+          configMap:
+            name: {{ include "newrelic.common.naming.fullname" . }}-deployment-config
```
kang-makes, line 60: There is a function in the `common-library` to add suffixes: https://github.com/newrelic/helm-charts/blob/162de8eac4822b633ad4e08ff6660035a677c328/library/common-library/DEVELOPERS.md#newreliccommonnamingtruncatetodnswithsuffix

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -0,0 +1,61 @@
...
+  enabled: true
+
+image:
+  repository: otel/opentelemetry-collector-contrib
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: "0.91.0"
+
+# -- This is the New Relic endpoint to send data to.
+newRelicEndpoint: "https://otlp.nr-data.net"
```
kang-makes, line 19: As I said above, this is a bad idea...

kang-makes (commented):
Here you have a second round of comments :D

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
dbudziwojskiNR: Removed the debug exporter :)

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
dbudziwojskiNR: Removed the debug exporter :)

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
dbudziwojskiNR: Renamed to `.Values.gateway.grpc.enable`

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
dbudziwojskiNR: Done :)

charts/nr-k8s-otel-collector/values.yaml
dbudziwojskiNR: Fixed. Thanks!

charts/nr-k8s-otel-collector/templates/daemonset.yaml
dbudziwojskiNR: Gotcha. We can add the missing common library features in a follow-up PR

charts/nr-k8s-otel-collector/templates/deployment.yaml
dbudziwojskiNR: Fixed.

charts/nr-k8s-otel-collector/values.yaml
dbudziwojskiNR: Removed this config value. Will add toggles for staging etc in followup pr.

charts/nr-k8s-otel-collector/values.yaml
csongnr: Not sure what this means, where do we have a label enabling autodiscovery from prometheus? this setting only enabled installing kube-state-metrics, nothing to do w/ autodiscovery

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
csongnr: we are getting pod logs w/ the filelog receiver, yes. 

since we removed logging exporter, shouldn't have concerns with the loop :)

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
csongnr: honestly, i assumed the customer would check the logs from the monitoring by accessing the logs in the pods themselves but perhaps this is not the way to think about it? I'm new to thinking about how customers would interact with the agents, but given that our customers are engineers... I would think that accessing logs in a pod would be basic knowledge, and given that customers would have to pay for that data to be written into new relic, i assumed they would prefer to not have the otel pod logs sent to NR. 

we can add a separate debug logging pipeline in a later PR, if it is deemed necessary to do, i don't think that should be a blocker on this PR, especially since the loop possibility is removed.

charts/nr-k8s-otel-collector/templates/daemonset.yaml
csongnr: yes, this is just a reminder to myself that this only works on linux nodes, we're not adding support for non-linux anytime soon

charts/nr-k8s-otel-collector/templates/service.yaml
csongnr: We're still considering using the gateway, unfortunately, we haven't had time yet to do load testing and cut down on ingest, the bottleneck is the big concern, but the main reason we want to have data written through the same collector is that there are some transforms we have to perform across multiple metrics that are collected separately via the agent pods (daemonset) and the gateway pod (deployment). Because otel does not yet fully support the kind of transforms we need to do, we are currently just not doing those transforms yet, but in the case that what we want to do becomes possible, we'll at least have to have _some_(not all, since bottleneck concern) metrics written through to the gateway from the daemonset. that being said, this is all hypothetical, because while have found a way to make those cross metric transforms in the NR platform, but it requires some engineering effort and isn't as simple as running a nrql query.  given this, we're going to keep the gateway service in here for now, as we may have to work on this sooner rather than later.

charts/nr-k8s-otel-collector/templates/daemonset.yaml
csongnr: Yeah, sorry this is just a reminder to myself that this won't work on non linux nodes, but thanks for the headsup on creating a new daemonset to support that functionality. that makes sense. we won't be including non linux nodes support in this PR, and i'll remove my comments since I've noted them down elsewhere, and we will be including the limitations of this agent in documentation otherwise.

charts/nr-k8s-otel-collector/values.yaml
paologallinaharbur: There is a toggle in the KSM chart `prometheusScrape: true`
That allows the KSM service to be "autodiscovered" by the label "prometheus.io/scrape". This is not bad per se, but if the customer have installed a Prometheus server or POMI or a prometheus agent they might end up scraping a log of additional metrics without realising

paologallinaharbur: https://github.com/prometheus-community/helm-charts/blob/f29a2580f698ab6cac895785fe72886c8a42fec7/charts/kube-state-metrics/values.yaml#L2

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -10,14 +10,14 @@ data:
     receivers:
       otlp:
         protocols:
-          {{ if .Values.enableGrpcGatewayEndpoint }}
+          {{ if .Values.gateway.grpc.enable }}
           grpc:
             endpoint: ${env:MY_POD_IP}:4317
           {{ end }}
```
kang-makes, line 16: If there will be a future PR to templatise the configuration I would remove as many toggles as possible and you decide the templates and the tree of values.

When I [said this](https://github.com/newrelic/helm-charts/pull/1342#discussion_r1594013650) was to give you an example, not encourging you to change it but to follow [Paolo's suggestion](https://github.com/newrelic/helm-charts/pull/1342#discussion_r1593583200) of removing it.

If it is not used yet it makes no sense to support it.

charts/nr-k8s-otel-collector/templates/_naming.tpl
kang-makes, line 1: Beautiful... Awesome... Wonderful... It's perfect :D

charts/nr-k8s-otel-collector/templates/_endpoint.tpl
kang-makes, line 1: I expected this to be in the follow up PR but it looks awesome if you want to have it now :)

charts/nr-k8s-otel-collector/templates/daemonset.yaml
kang-makes: Awesome :)

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
kang-makes: I would leave this to a follow up PR but leave a huge comment because this could be a huge issue if this goes GA.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
kang-makes: This is still unsolved and I think this is quite important.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,299 @@
...
+          - key: host.id
+            from_attribute: host.name
+            action: upsert
+          # TODO (chris): Upsert only when cluster name not found (resource detection override: true)
+          - key: k8s.cluster.name
+            action: upsert
+            value: {{ include "newrelic.common.cluster" . }}
+          - key: newrelicOnly
+            action: upsert
+            value: 'true'
```
kang-makes, line 216: NIT: I'll think about these kind of tags twice as we charge the client for them.

charts/nr-k8s-otel-collector/templates/daemonset.yaml
```diff
@@ -0,0 +1,98 @@
...
+        {{- include "newrelic.common.labels.podLabels" . | nindent 8 }}
+        component: daemonset
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/daemonset-configmap.yaml") . | sha256sum }}
+    spec:
+      serviceAccountName: {{ include "newrelic.common.serviceAccount.name" . }}
+      containers: 
+        - name: otel-collector-daemonset
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
```
kang-makes, line 25: To get the processes from the host to monitor it, the pod should be privileged. Right?

kang-makes (commented):
And the last one!

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
mangulonr: What is our actual proposal to scrape monitoring logs from nri_bundle?
Shouldn't we do the same?

kang-makes: That should be asked to the logging team as they are the ones that understand how fluentbit works.

We never had to go this far. This is why this is becoming tricky.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
dbudziwojskiNR: For sure. Unfortunately, this is a requirement for synthesis from our understanding. This said, I took some notes on this as something to look into in the future.

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
dbudziwojskiNR: We talked about it and we agree that it doesn't make sense to keep it in unused. We've gone ahead and removed it. Thanks!

charts/nr-k8s-otel-collector/values.yaml
dbudziwojskiNR: Fixed :)

charts/nr-k8s-otel-collector/templates/daemonset.yaml
csongnr: Yes, you are right, thanks :D

csongnr (approved):

PR #1390, opened 2024-06-06
nri-statsd: Support customSecretName Statsd Chart

RoseSecurity:
# What and Why

### Acceptance Criteria

Before:

```yaml
global:
   insightsKey: ""
```

After:

```yaml
global:
```

## Checks

- [X] Read contributing guidelines
- [X] Chart Version bumped
- [X] Support [this issue](https://github.com/newrelic/helm-charts/issues/1324)**

### DIFF
--- a/charts/nri-statsd/templates/_helpers.tpl
+++ b/charts/nri-statsd/templates/_helpers.tpl
@@ -78,14 +78,18 @@ Define the statsd metric tags
 Return the insightsKey
 */}}
 {{- define "nri-statsd.insightsKey" -}}
-{{- if .Values.global}}
-  {{- if .Values.global.insightsKey }}
-      {{- .Values.global.insightsKey -}}
-  {{- else -}}
-      {{- .Values.insightsKey | default "" -}}
-  {{- end -}}
+{{- if or (include "newrelic.common.license._customSecretName" .) (include "newrelic.common.license._customSecretKey" .) -}}
+    {{- if .Values.global}}
+        {{- if .Values.global.insightsKey }}
+            {{- .Values.global.insightsKey -}}
+        {{- else -}}
+            {{- .Values.insightsKey | default "" -}}
+        {{- end -}}
+    {{- else -}}
+        {{- .Values.insightsKey | default "" -}}
+    {{- end -}}
 {{- else -}}
-    {{- .Values.insightsKey | default "" -}}
+    {{- "" -}}
 {{- end -}}
 {{- end -}}
 
--- a/charts/nri-statsd/values.yaml
+++ b/charts/nri-statsd/values.yaml
@@ -10,6 +10,7 @@
 #
 # global:
 #   insightsKey: ""
+#   customSecretName: ""
 #   cluster: ""
 
 accountId: ""


RoseSecurity:
@DavSanchez Hello, would this PR accomplish what we discussed [here](https://github.com/newrelic/helm-charts/pull/1327#issuecomment-2104752753)?

charts/nri-statsd/templates/_helpers.tpl
```diff
@@ -78,14 +78,18 @@ Define the statsd metric tags
 Return the insightsKey
 */}}
 {{- define "nri-statsd.insightsKey" -}}
-{{- if .Values.global}}
-  {{- if .Values.global.insightsKey }}
-      {{- .Values.global.insightsKey -}}
-  {{- else -}}
-      {{- .Values.insightsKey | default "" -}}
-  {{- end -}}
+{{- if or (include "newrelic.common.license._customSecretName" .) (include "newrelic.common.license._customSecretKey" .) -}}
```
kang-makes, line 81: You are using internal functions from the `common-library` that should not be used: https://github.com/newrelic/helm-charts/blob/bda20f2057e742cede14dd947ccb579f27993fb0/library/common-library/templates/_license.tpl#L29-L32

The point of using these functions is to set which key inside the secret will be used whenever the case the key is created by the user of by the values file.

You would need to create a new function in the `common-library` that supports `insightsKey` and you should not use the internal ones in the final chart but the final ones directly in the deployment. Check the usage here: https://github.com/newrelic/helm-charts/blob/bda20f2057e742cede14dd947ccb579f27993fb0/library/common-library/DEVELOPERS.md#_licensetpl

kang-makes (changes requested):
Added to the point that the `common-library` is not being used correctly, the chart does not template:
 
Screenshot 2024-06-17 at 16 16 29

You are trying to use `common-library` without being a dependency of this chart.

If you need to add an example on how `licenseKey` was added to the `common-library`, you can use as inspiration the PR that added that feature: https://github.com/newrelic/helm-charts/pull/749

Once the `common-library` has everything to support `.global.insightsKey` and `.insightsKey`, the change in `nri-statsd` should be more a clean up than a refactor :)

Let me know if you need more help/assistance for this feature. We really appreciate PRs that are missing functionality we were not required to develop/support 😃

charts/nri-statsd/templates/_helpers.tpl
RoseSecurity: @kang-makes Although the `common-library` does not support the `insightsKey`, this essentially just a licenseKey I believe? Is there any way we could just use that functionality to implement it?

kang-makes: The way we support installation on Kubernetes is by using `nri-bundle`, not individual charts: https://github.com/newrelic/helm-charts/blob/bda20f2057e742cede14dd947ccb579f27993fb0/charts/nri-bundle/Chart.yaml

I know that `nri-statsd` is an exception but we work in a way that charts are composable and having the same UX so we should support them in a consistent way. Hence the `common-library` creating a seamless UX though all the charts.

If we are going to support a new global key we should support it in the `common-library` so other charts can use it.

RoseSecurity: Understood, well we would love to make use of this feature to ensure that we are not hardcoding API keys into our source control using the existing `nri-statsd` chart. Our goal is to inject the secret into the cluster before referencing the secret name in the chart

kang-makes:
Hi!

The PR to add Insights secrets to the `common-library` was merged: #1402

You can see an example on how to use them in the chart that is used for testing: https://github.com/newrelic/helm-charts/pull/1402/files#diff-394b2b36b0f27c9bf3f7555749adc69f9f92db9fc33020f4393d566efc10933aR72-R76

You can now set the `common-library` as a dependency and use these helpers to template the insights key or use an external secret whenever you want.

PR #1379, opened 2024-05-27
[super-agent] remove jobs

paologallinaharbur:
#### Is this a new chart
no

#### What this PR does / why we need it:

Right now we do not have auto=upgrade in place, therefore it is better to install in the dependency in an easy way.

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version bumped
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[mychartname]`)

PR #1393, opened 2024-06-11
[super-agent-deployment]: add authSecret configuration and validation

marcsanmi:
#### Is this a new chart
no

This PR introduces a new configuration option for managing authentication secrets within the Helm chart:

1. **authSecret Configuration**: Added to the values.yaml, allowing users to specify and create an authentication secret.
2. **Helper Function**: Ensures auth_key is provided when authSecret.create is true, preventing misconfigurations.
3. **Deployment Template Updates**: Includes the new authSecret in the deployment, ensuring it is correctly mounted.
4. **Helm Unittest**: Added test cases to validate the correct creation, mounting, and error handling of authSecret.

Please review the changes and let me know if any adjustments are needed

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/super-agent-deployment/README.md
+++ b/charts/super-agent-deployment/README.md
@@ -2,7 +2,7 @@
 
 # super-agent-deployment
 
-![Version: 0.0.16-beta](https://img.shields.io/badge/Version-0.0.16--beta-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.13.1](https://img.shields.io/badge/AppVersion-0.13.1-informational?style=flat-square)
+![Version: 0.0.17-beta](https://img.shields.io/badge/Version-0.0.17--beta-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.14.0](https://img.shields.io/badge/AppVersion-0.14.0-informational?style=flat-square)
 
 A Helm chart to install New Relic Super agent on Kubernetes
 
@@ -35,6 +35,7 @@ At the point of the creation of the chart, it has no particularities and this se
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | affinity | object | `{}` | Sets pod/node affinities. Can be configured also with `global.affinity` |
+| authSecret | object | `{"create":false}` | Settings controlling authentication secret creation. If `create` is true, a Kubernetes secret will be created containing a key named `auth_key`. This secret will be mounted in the deployment pod at the path `/etc/newrelic-super-agent/auth_key` for authentication purposes. |
 | cleanupManagedResources | bool | `true` | Enable the cleanup of super-agent managed resources when the chart is uninstalled. If disabled, agents and / or agent configurations managed by the super-agent will not be deleted when the chart is uninstalled. |
 | cluster | string | `""` | TODO: Name of the Kubernetes cluster monitored. Can be configured also with `global.cluster`. |
 | config.subAgents | object | See `values.yaml` for examples | Values that the fleet is going to have in the deployment. |
--- a/charts/super-agent-deployment/templates/_helpers.tpl
+++ b/charts/super-agent-deployment/templates/_helpers.tpl
@@ -52,3 +52,32 @@ readOnlyRootFilesystem: true
     {{- toYaml $defaults -}}
 {{- end -}}
 {{- end -}}
+
+{{- /*
+Check if authSecret.create is explicitly set to true. If authSecret is not empty and create is not defined, default it to false.
+*/ -}}
+{{- define "newrelic-super-agent.shouldCreateAuthSecret" -}}
+{{- $authSecret := .Values.authSecret }}
+{{- if and (hasKey $authSecret "create") }}
+  {{- toYaml $authSecret.create -}}
+{{- else if not (empty $authSecret) }}
+  {{- toYaml false -}}
+{{- else }}
+  {{- toYaml false -}}
+{{- end }}
+{{- end -}}
+
+{{- /*
+Check if authSecret.data and auth_key are provided. Fail if not.
+*/ -}}
+{{- define "newrelic-super-agent.authSecret.validateData" -}}
+{{- $authSecret := .Values.authSecret }}
+{{- if and $authSecret (not (empty $authSecret)) }}
+  {{- if not $authSecret.data }}
+    {{- fail "authSecret.data must be provided when authSecret.create is true" }}
+  {{- end }}
+  {{- if not $authSecret.data.auth_key }}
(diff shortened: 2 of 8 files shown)


charts/super-agent-deployment/templates/secret.yaml
paologallinaharbur, line 1: auh_secret?

charts/super-agent-deployment/README.md
```diff
@@ -2,7 +2,7 @@
 
 # super-agent-deployment
 
-![Version: 0.0.16-beta](https://img.shields.io/badge/Version-0.0.16--beta-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.13.1](https://img.shields.io/badge/AppVersion-0.13.1-informational?style=flat-square)
+![Version: 0.0.17-beta](https://img.shields.io/badge/Version-0.0.17--beta-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.14.0](https://img.shields.io/badge/AppVersion-0.14.0-informational?style=flat-square)
```
paologallinaharbur, line 5: can we remove from the template the snippet generating this?

charts/super-agent-deployment/values.yaml
```diff
@@ -116,6 +128,12 @@ config:
       #    endpoint: https://opamp.service.newrelic.com/v1/opamp
       #    headers:
       #      api-key: LICENSE_KEY
+      #    auth_config:
+      #      token_url: "http://fake.com/oauth2/v1/token"
+      #      client_id: "fake"
+      #      provider: "local"
```
paologallinaharbur, line 134: provider? which values are supported?

charts/super-agent-deployment/values.yaml
```diff
@@ -116,6 +128,12 @@ config:
       #    endpoint: https://opamp.service.newrelic.com/v1/opamp
       #    headers:
       #      api-key: LICENSE_KEY
+      #    auth_config:
+      #      token_url: "http://fake.com/oauth2/v1/token"
```
paologallinaharbur, line 132: should nrStaging config change this? I wonder if we should add it "magically" (dame for endpoint few lines above)

charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -52,3 +52,32 @@ readOnlyRootFilesystem: true
     {{- toYaml $defaults -}}
 {{- end -}}
 {{- end -}}
+
+{{- /*
+Check if authSecret.create is explicitly set to true. If authSecret is not empty and create is not defined, default it to false.
+*/ -}}
+{{- define "newrelic-super-agent.authSecretDefaults" -}}
```
paologallinaharbur, line 59: the name authSecretDefaults is not that clear 😕

charts/super-agent-deployment/templates/deployment-superagent.yaml
```diff
@@ -75,6 +75,12 @@ spec:
             {{- with .Values.extraVolumeMounts }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
+            {{- $createAuthSecret := include "newrelic-super-agent.authSecretDefaults" . }}
+            {{- if eq $createAuthSecret "true" }}
```
paologallinaharbur, line 79: I would merge it in one line and compute it twice

charts/super-agent-deployment/values.yaml
marcsanmi: I added all the exposed fields from -> https://github.com/newrelic/newrelic-super-agent/blob/main/super-agent/src/opamp/auth/config.rs#L11, bud since it has the `deserialize_default_provider` I can just ignore the provider 🤔 (however it's still exposed if set)

charts/super-agent-deployment/README.md
marcsanmi: I generated it with helm docs 🤔

paologallinaharbur: for example I've removed it from the other chart [here](https://github.com/newrelic/helm-charts/pull/1379/files#r1642502901). It was a @kang-makes idea to remove it so that we do not need to update it since it adds no real value

charts/super-agent-deployment/values.yaml
paologallinaharbur: 🤔 ah ok, so right now there is only "local" and we are going to have something else in the future...

your call I am fine having or not having it for now

charts/super-agent-deployment/templates/_helpers.tpl
marcsanmi: Updated to `shouldCreateAuthSecret`.

charts/super-agent-deployment/values.yaml
marcsanmi: I just removed `provider`.

charts/super-agent-deployment/values.yaml
```diff
@@ -85,7 +85,7 @@ proxy: ""
 
 # -- (bool) Send the metrics to the staging backend. Requires a valid staging license key. Can be configured also with `global.nrStaging`
 # @default -- `false`
-nrStaging:
+nrStaging:  # TODO: Update OpAMP `endpoint` and auth `token_url` for staging and Europe when nrStaging is enabled.
```
sigilioso, line 88: Nit: On top of adding the TODO, should we update the comment above to say that the other fields should be updated accordingly?

charts/super-agent-deployment/tests/auth_secret_test.yaml
```diff
@@ -0,0 +1,111 @@
...
+    set:
+      cluster: test
+      authSecret:
+        create: true
+        data:
+          auth_key:
+    asserts:
+      - template: templates/auth-secret.yaml
+        failedTemplate:
+          errorMessage: auth_key must be provided when authSecret.create is true
```
sigilioso, line 111: Should we add a test-case for when `authSecret.create` is not set but there is other fields in `authSecret`?

charts/super-agent-deployment/templates/deployment-superagent.yaml
```diff
@@ -90,6 +95,11 @@ spec:
         {{- with .Values.extraVolumes }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
+        {{- if eq (include "newrelic-super-agent.shouldCreateAuthSecret" .) "true" }}
+        - name: auth-secret-volume
+          secret:
+            secretName: {{ .Values.authSecret.name | default .Release.Name }}
```
sigilioso, line 101: Should the default has some sort of prefix? Would also make sense a default value in `values.yaml`?

sigilioso (commented):
Thanks! I've left some comments, mostly _nits_

sigilioso (approved):
🚀

PR #1402, opened 2024-06-19
[common-library] Add insights key helpers

### LINKED ISSUE #1324: Support `customSecretName` for `nri-statsd`

kang-makes:
#### Is this a new chart
No.

Issue #1324 ask for support to have the insights key out of the `values.yaml`. This is an opportunity to have the same UX for that key that we have for the rest of secrets/licenses.

This PR does not fix any issue yet but it is the first step to fix #1324.

Once this PR is merged, we have the tooling to add `common-library` to `nri-statsd`.

Some test were not passing because of updates of `helm unittest` so I have to fix them. 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/library/CHART-TEMPLATE/Chart.lock
+++ b/library/CHART-TEMPLATE/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common-library
   repository: file://../common-library
-  version: 1.1.0
-digest: sha256:810c8b54c04a83d06208becb3ee0c3268fbff165624ef8d0da5d75d5a3ae5074
-generated: "2022-08-31T10:40:37.947861+02:00"
+  version: 1.2.0
+digest: sha256:ca0a371fb6ce46b821593db9949d1942f2fdf40b54df1eff377ac094bc3d78fa
+generated: "2024-06-19T10:51:12.685283+02:00"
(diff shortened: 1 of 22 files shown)


juanjjaramillo (approved):

sigilioso (approved):
LGTM

PR #1409, opened 2024-06-26
[super-agent] Removing inter-chart dependency

paologallinaharbur:
#### Is this a new chart
no

The Idea is to move the super-agent-deployment dependency to the `chart` folder so that:
 - we have only one values
 - we have only one place for default and readme
 - it is easier to deploy since we can combine everything in one PR
- I tried as well to move it to the template folder, but I had issues with the common library and to access things behing super-agent-deployment

 - Focus on the agent-values changes, everything else is a copy paste

#### Checklist
- [x] Chart Version bumped


charts/super-agent/charts/super-agent-deployment/tests/__snapshot__/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
kang-makes, line 1:
```suggestion
**
```

kang-makes (commented):
Instead creating two PRs, one to copy the files (this one) and one to remove them (#1410) I would move everything on the same PR and rename the PR to something like "remove inter chart dependency" or something like that.

Doing it in the same PR not only allow to do a better `git blame` but also make the `git diff` way more easy to read. Most of the files would be a simple `git mv` that change the path from one side to another and change (at most) a few lines.

What do you think?

charts/super-agent/charts/super-agent-deployment/tests/__snapshot__/.gitignore
paologallinaharbur: yes the second has this as base, I would merge that into this, I am not sure how it would work with ci/cd, if it fails we can break it down again

paologallinaharbur: we can merge https://github.com/newrelic/helm-charts/pull/1410 here

paologallinaharbur: I ended up doing it directly

paologallinaharbur:
I ended up merging the other PR directly to improve readability of this, if the CI fails we can fix it later on

charts/super-agent/Chart.yaml
```diff
@@ -3,20 +3,20 @@ name: super-agent
...
   - name: flux2
     repository: https://fluxcd-community.github.io/helm-charts
     version: 2.13.0
     condition: flux2.enabled
+  - name: super-agent-deployment
+    version: 0.0.18-beta
+    condition: super-agent-deployment.enabled
+    # The following dependency is needed as sub-dependency of super-agent-deployment
   - name: common-library
     version: 1.1.1
```
kang-makes, line 18:
```suggestion
    version: 1.2.0
```

kang-makes (approved):
Thanks!

PR #1416, opened 2024-07-02
[super-agent] Making sure all resources are deleted

paologallinaharbur:
#### Is this a new chart
This PR:
 -  forces `CT` to use a separate naspace so that we are sure `kubectl delete ns ...` will wait that resources are garbage collected (`ct` is not doing so when running helm delete)
 - modifies the uninstall job to make sure the super agent is NOT running before deleting resources. This is needed to avoid the SA recreating it right after  

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/.github/workflows/lint_test_charts.yaml
+++ b/.github/workflows/lint_test_charts.yaml
@@ -86,8 +86,6 @@ jobs:
           github token: ${{ secrets.GITHUB_TOKEN }}
           driver: docker
           start args: "--container-runtime=containerd"
-      - name: Create test ns
-        run: kubectl create ns ct
 
       - name: Create certificate manager
         run: |
@@ -98,9 +96,9 @@ jobs:
             --set crds.enabled=true
 
       - name: Test install charts
-        run: ct install --namespace ct --config .github/ct-install.yaml
+        run: ct install --config .github/ct-install.yaml
       - name: Test upgrade charts
-        run: ct install --namespace ct --config .github/ct-install.yaml --upgrade
+        run: ct install --config .github/ct-install.yaml --upgrade
 
   bundle-integrity:
     name: Test nri-bundle dependencies has the same version of the common-library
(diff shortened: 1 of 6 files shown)


sigilioso (approved):
LGTM 👍

.github/workflows/lint_test_charts.yaml
```diff
@@ -86,8 +86,6 @@ jobs:
           github token: ${{ secrets.GITHUB_TOKEN }}
           driver: docker
           start args: "--container-runtime=containerd"
-      - name: Create test ns
-        run: kubectl create ns ct
```
kang-makes, line 90: The thing on testing in a namespace (`ct` in this pipeline) it to test that we are installing it a namespace that is neither `default` nor `newrelic` so you can test that all the objects are created in a different namespace and are able to boot.

I don't get why we need to remove the namespace. I don't get the race condition we are fixing... Why without it we remove the race condition?

paologallinaharbur: Without `--namespace ns` selects a random namespace that it is then deleted:
```
namespace "super-agent-6s2970ilqu" deleted
```

By doing so we are:
 - covering both that we are installing in a namespace different from default nor newrelic
 - having "for free" a `kubectl delete --wait` it waits for all resources to be deleted (sadly we cannot add it as a flag of `Helm delete`)

kang-makes: > selects a random namespace that it is then deleted

TIL. Sounds awesome then :D

kang-makes (approved):

.github/workflows/lint_test_charts.yaml
juanjjaramillo: Cool, I learned something new too

juanjjaramillo (approved):

PR #1439, opened 2024-07-24
[super-agent] feat: add license key as env from secret (NR-293501)

gsanchezgavier:

- Adds license key as env variable  from the secret for the SA deployment. This will allow to use env vars placeholders in configs rather than plain secrets for sub-agents.

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/super-agent/charts/super-agent-deployment/templates/deployment-superagent.yaml
+++ b/charts/super-agent/charts/super-agent-deployment/templates/deployment-superagent.yaml
@@ -54,8 +54,20 @@ spec:
           image: {{ include "newrelic.common.images.image" ( dict "imageRoot" .Values.image "context" .) }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
 
-          {{- with .Values.extraEnv }}
           env:
+            - name: NR_LICENSE_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "newrelic.common.license.secretName" . }}
+                  key: {{ include "newrelic.common.license.secretKeyName" . }}
+          {{- if .Values.config.superAgent.content.opamp }}
+            - name: NR_SA_OPAMP__HEADERS__API-KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "newrelic.common.license.secretName" . }}
+                  key: {{ include "newrelic.common.license.secretKeyName" . }}
+          {{- end }}
+          {{- with .Values.extraEnv }}
             {{- toYaml . | nindent 12 }}
           {{- end }}
           {{- with .Values.extraEnvFrom }}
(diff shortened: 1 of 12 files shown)


charts/super-agent/values.yaml
```diff
@@ -138,7 +138,7 @@ super-agent-deployment:
         #   endpoint: https://opamp.service.newrelic.com/v1/opamp
         #   headers:
         #     # TODO this value should be automatically injected
-        #     api-key: LICENSE_KEY
+        #     api-key: ${nr-env:NR_LICENSE_KEY}
```
paologallinaharbur, line 141: I think this is wrong

charts/super-agent/charts/super-agent-deployment/templates/deployment-superagent.yaml
```diff
@@ -54,8 +54,13 @@ spec:
...
           imagePullPolicy: {{ .Values.image.pullPolicy }}
 
-          {{- with .Values.extraEnv }}
           env:
+            - name: NR_LICENSE_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "newrelic.common.license.secretName" . }}
+                  key: {{ include "newrelic.common.license.secretKeyName" . }}
+          {{- with .Values.extraEnv }}
```
paologallinaharbur, line 63: I would add an extra env var NR_SA_OPAMP__HEADERS__API-KEY so that we van remove the "opamp:" entry in the sa config

charts/super-agent/values.yaml
paologallinaharbur: nice point 😕 I guess that also Juanma PR has the same issue of "enabling by default" opamp

paologallinaharbur: I am checking how complex it is to include enable/disable for opamp

charts/super-agent/charts/super-agent-deployment/tests/deployment_superagent_securityContext_test.yaml
```diff
@@ -3,6 +3,9 @@ templates:
   - templates/deployment-superagent.yaml
   - templates/configmap-superagent-config.yaml
   - templates/configmap-subagent-configs.yaml
+release:
```
paologallinaharbur, line 6: why this change is needed?

paologallinaharbur (changes requested):
All looks good, we need a release and to bump the image.

I would consider changing the cluster as well in a follow up PR if it is a few line change

charts/super-agent/charts/super-agent-deployment/tests/deployment_superagent_securityContext_test.yaml
gsanchezgavier: leftover, since i ended up moving the test to other file

charts/super-agent/charts/super-agent-deployment/templates/deployment-superagent.yaml
gsanchezgavier: done

paologallinaharbur (approved):

PR #1441, opened 2024-07-25
[newrelic-logging] Adjust order of env list in daemonset to fix dependency reference

### LINKED ISSUE #1408: [newrelic-logging] BREAKING Bug in env values order
Upon an attempted upgrade for the `nri-bundle` chart from `5.0.67` to `5.0.82`, we've started seeing one pod in the newrelic-logging deamonset running fine but others throwing `cannot execute insert file ...` and other db related errors.

For a bit of context, running on k8s 1.29 in EKS with :
```
  newrelic-logging:
    fluentBit:
      persistence:
        mode: persistentVolume
```

On further inspection, it was noted that there was only one db file being created on the volume, named `'$(NODE_NAME)-fb.db'`, which is the obvious reason why one pod would work fine, as other's wouldn't be able to get the lock on the same file.

We believe the issue was introduced with `newrelic-logging 1.21.0` , where [this change](https://github.com/newrelic/helm-charts/commit/f6c71610196899e6248fb0b641ace0be4d6c3df9#diff-dba7a505c5df68aac95df70ba65d1a43ac0349a7692e6a1ecd2899854ba90894L108-L128) was made. 
The reason why this is not working any more is because `NODE_NAME` is now defined after `FB_DB`. Here's the snippet from [k8s docs](https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/) that explains why this is an issue: 
```
Note that order matters in the env list. An environment variable is not considered "defined" if it is specified further down the list. 
```

At the moment, this is stopping us from applying updates Newrelic has notified us about resolving some vulnerabilities.

nluedema:
#### Is this a new chart
No

When using `persistentVolume` mode, the env variable `FB_DB` references `NODE_NAME`. This reference is currently not resolved, because `NODE_NAME` is defined after `FB_DB`. Therefore, only a single database file named `'$(NODE_NAME)-fb.db'` is created in the persistent volume, instead of one file for each node. #1408 traced the problem down to this [change](https://github.com/newrelic/helm-charts/commit/f6c71610196899e6248fb0b641ace0be4d6c3df9#diff-dba7a505c5df68aac95df70ba65d1a43ac0349a7692e6a1ecd2899854ba90894L108-L128), which moved the definition of `NODE_NAME` after the definition of `FB_DB`.

This PR fixes this, by moving the definition of `NODE_NAME` before the definition of `FB_DB` again.

The behavior of kubernetes for dependent environment variables is documented [here](https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/).

  - fixes #1408

I was thinking about a test that checks if `NODE_NAME` is defined before `FB_DB`, but I wasn't able to come up with a way to do this using `helm-unittest`.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/newrelic-logging/Chart.yaml
+++ b/charts/newrelic-logging/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart to deploy New Relic Kubernetes Logging as a DaemonSet, supporting both Linux and Windows nodes and containers
 name: newrelic-logging
-version: 1.23.0
+version: 1.23.1
 appVersion: 2.0.2
 home: https://github.com/newrelic/kubernetes-logging
 icon: https://newrelic.com/assets/newrelic/source/NewRelic-logo-square.svg
--- a/charts/newrelic-logging/templates/daemonset.yaml
+++ b/charts/newrelic-logging/templates/daemonset.yaml
@@ -96,6 +96,11 @@ spec:
               {{- else }}
               value: "docker,cri"
               {{- end }}
+            # NODE_NAME needs to be defined before FB_DB, because FB_DB references NODE_NAME in its value when using persistentVolume
+            - name: NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
             {{- if or (not .Values.fluentBit.persistence) (eq .Values.fluentBit.persistence.mode "hostPath") }}
             - name: FB_DB
               value: {{ .Values.fluentBit.db | quote }}
@@ -116,10 +121,6 @@ spec:
               value: {{ include "newrelic-logging.lowDataMode" . | default "false" | quote }}
             - name: RETRY_LIMIT
               value: {{ .Values.fluentBit.retryLimit | quote }}
-            - name: NODE_NAME
-              valueFrom:
-                fieldRef:
-                  fieldPath: spec.nodeName
             - name: HOSTNAME
               valueFrom:
                 fieldRef:


mickutz (approved):

jsubirat (changes requested):
Hi @nluedema ,

Thank you for identifying this issue. Indeed, I didn't realize that moving `NODE_NAME` would affect `FB_DB` when using persistent volumes due to ordering. The reason this is happening though is because that environment variable uses `$(NODE_NAME)` instead of `${NODE_NAME}`, which means that it is Helm resolving the value of these variables rather than Fluent Bit when it starts.

To fix this, I would suggest to leave `NODE_NAME` untouched and instead fix `FB_DB`. We don't want environment variable definition to be dependent on order. To achieve this, could you please:
- Move `NODE_NAME` back to where it was
- Modify this:
```
- name: FB_DB
   value: "/db/$(NODE_NAME)-fb.db"
```
  to this:
  ```
- name: FB_DB
   value: "/db/${NODE_NAME}-fb.db"
```

This will ensure that environment variable definitions stay order-insensitive and that it is actually the Fluent Bit process that resolves `NODE_NAME` upon start rather than Helm.

nluedema:
Hi @jsubirat,

Thanks for the feedback! I didn't know that Fluent Bit is able to resolve environment variables. Doing it that way is of course a lot more elegant. I updated the PR according to your suggestions.

nr-rkallempudi:
Hi @nluedema , We have validated your PR and found that Fluent Bit is not resolving the variable name ${NODE_NAME} as it doesn't support nested environment variable interpolation. So we agreed internally to do with your previous code changes for now even though it keeps order dependency. 

Can you update code to previous one with curved braces and also add a comment mentioning the order dependency to have that variable resolved. So we can proceed with merging your PR.

nr-rkallempudi (changes requested):
Hi @nluedema , We have validated your PR and found that Fluent Bit is not resolving the variable name ${NODE_NAME} as it doesn't support nested environment variable interpolation. So we agreed internally to do with your previous code changes for now even though it keeps order dependency.

Can you update code to previous one with curved braces and also add a comment mentioning the order dependency to have that variable resolved. So we can proceed with merging your PR.

nluedema:
@nr-rkallempudi @jsubirat 
I updated the PR to only include the initial changes. Let me know what you think! A pity that nested environment variable interpolation doesn't work.

nr-rkallempudi (approved):

jsubirat (approved):
Thanks folks, and apologies for the confusion!

PR #1463, opened 2024-08-14
[common-library] Add missing features to the common library (NR-281093)

kang-makes:
#### Is this a new chart
No

We thought that some of the features that we need for the super-agent would be awesome to have them at `common-library` level.

We need the chart to be able to calculate the region (or let the user to specify it) and a new API Key.

To read it easier, I would go commit by commit as I added a new feature with its own documentation on each commit.

#### Checklist
- [x] Chart Version bumped


library/CHART-TEMPLATE/tests/cl-region.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+  - it: Overrides global region with the local one
+    set:
+      global:
+        region: EU
+      region: local
+    asserts:
+      - equal:
+          path: data.region
+          value: Local
```
sigilioso, line 48: Should we add another test case to check that `EU` region is inferred from the api-key when region is not set but the the api-key is available?

library/common-library/templates/_region.tpl
```diff
@@ -0,0 +1,74 @@
...
+{{- /* Defaults */ -}}
+{{- $region := "us" -}}
+{{- if include "newrelic.common.nrStaging" . -}}
+  {{- $region = "staging" -}}
+{{- else if include "newrelic.common.region._isEULicenseKey" . -}}
+  {{- $region = "eu" -}}
+{{- end -}}
+
+{{- include "newrelic.common.region.validate" (include "newrelic.common.region._fromValues" . | default $region ) -}}
+{{- end -}}
```
sigilioso, line 18: Would it make sense to fail or add a warning when `include "newrelic.common.region._isEULicenseKey" .` returns `"true"` but `include "newrelic.common.region._fromValues" . | lower` is not `"eu"`?

sigilioso (commented):
The changes look really good to me. I've left some comments to be discussed

library/common-library/templates/_region.tpl
```diff
@@ -0,0 +1,74 @@
...
+
+
+
+{{/*
+Returns the region from the values if valid. This only return the value from the `values.yaml`.
+More intelligence should be used to compute the region.
+
+Usage: `include "newrelic.common.region.validate" "us"`
+*/}}
```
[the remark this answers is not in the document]
kang-makes: It is already documented: https://github.com/newrelic/helm-charts/pull/1463/files#diff-ead97ed97e34739841bc3318ba1886dbdeb49568c970a273c57acc8f7001c2a4R552-R564

library/common-library/templates/_region.tpl
kang-makes: I wanted to keep the way the functions works consistent with how the `common-library` works: It computes things for you but then you can footgun yourself if you want to force something.

I also preferred this way to be consistent with the parsing of the license key. We cannot distinguish between a prod and a staging key, and we cannot fail if the key is not valid so I should not fail if the key and the region and the key does not match.

library/common-library/templates/_region.tpl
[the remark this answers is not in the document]
sigilioso: Sorry I missed the _normalized_ part. Thanks!

library/CHART-TEMPLATE/tests/cl-region.yaml
kang-makes: Good point. I completely forgot. Done [#27eb921](https://github.com/newrelic/helm-charts/pull/1463/commits/27eb92198bd310c1c9136893e9d50a56173e4109)

library/CHART-TEMPLATE/tests/cl-apikey_secret_test.yaml
```diff
@@ -0,0 +1,56 @@
+suite: test insightsKey secret
+templates:
+  - templates/example-insights-secret.yaml
```
sigilioso, line 3: As discussed offline this file should check the `example-apiKey-secret.yaml` instead 📋

library/CHART-TEMPLATE/README.md
```diff
@@ -36,17 +36,27 @@ low data modes or any other quirk that it could have.
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | affinity | object | `{}` |  |
+| apiKey | string | `"barqux"` |  |
 | cluster | string | `"barfoo"` |  |
 | containerSecurityContext | object | `{}` |  |
+| customAPIKeySecretKey | string | `""` |  |
+| customAPIKeySecretName | string | `""` |  |
+| customInsightsKeySecretKey | string | `""` |  |
+| customInsightsKeySecretName | string | `""` |  |
```
paologallinaharbur, line 45: aren't these deprecated?
Screenshot 2024-08-23 at 10 50 45

library/CHART-TEMPLATE/README.md
```diff
@@ -36,17 +36,27 @@ low data modes or any other quirk that it could have.
...
 | customSecretLicenseKey | string | `""` |  |
 | customSecretName | string | `""` |  |
 | deploymentAnnotations | object | `{}` |  |
 | dnsconfig | object | `{}` |  |
 | fedRamp.enabled | string | `nil` |  |
 | fullnameOverride | string | `""` |  |
 | global.affinity | object | `{}` |  |
+| global.apiKey | string | `""` |  |
```
paologallinaharbur, line 53: aren't we missing apiKey without global?

library/CHART-TEMPLATE/README.md
```diff
@@ -36,17 +36,27 @@ low data modes or any other quirk that it could have.
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | affinity | object | `{}` |  |
+| apiKey | string | `"barqux"` |  |
```
paologallinaharbur, line 39: Is it called userKey in the docs? https://docs.newrelic.com/docs/apis/intro-apis/new-relic-api-keys/

I think that we should use the same name or add a explanatory description

library/common-library/templates/_region.tpl
```diff
@@ -0,0 +1,74 @@
...
+{{- end -}}
+
+
+
+{{/*
+Returns the region from the values. This only return the value from the `values.yaml`.
+More intelligence should be used to compute the region.
+This helper is for internal use.
+*/}}
+{{- define "newrelic.common.region._fromValues" -}}
```
paologallinaharbur, line 51: why this is not calling validate?

library/CHART-TEMPLATE/tests/cl-apikey_secret_test.yaml
kang-makes: Really good point. Some copy and paste that I forgot to change. Done in [#15a9b7b](https://github.com/newrelic/helm-charts/pull/1463/commits/15a9b7b7959af5198c2786fe11f58bec8a0f33a3)

library/CHART-TEMPLATE/README.md
kang-makes: They are, but they are needed for StatsD.

You see it in the diff because the last PR when this was added did not added it to `vaules.yaml` and they were not templated in the readme.

library/CHART-TEMPLATE/README.md
kang-makes: Good point. Changed on [#0486574](https://github.com/newrelic/helm-charts/pull/1463/commits/0486574ced2ba833ae6bd823d51ba4bbe2c4a6f5)

library/CHART-TEMPLATE/README.md
kang-makes: PR when this was added: newrelic/helm-charts#1402

sigilioso (approved):
LGTM

library/common-library/templates/_region.tpl
```diff
@@ -0,0 +1,74 @@
+{{/*
+Return the region that is being used by the user
+*/}}
+{{- define "newrelic.common.region" -}}
+{{- if and (include "newrelic.common.license._usesCustomSecret" .) (not (include "newrelic.common.region._fromValues" .)) -}}
```
dbudziwojskiNR, line 5: I may be mistaken, but can we not evaluate the value of the secret to check if it's an US/EU key? We did this in the OTEL chart: https://github.com/newrelic/helm-charts/blob/master/charts/nr-k8s-otel-collector/templates/_endpoint.tpl

dbudziwojskiNR (changes requested):

library/common-library/templates/_region.tpl
kang-makes: Let's read that line:
The implementation of `newrelic.common.license._usesCustomSecret` is [here](https://github.com/newrelic/helm-charts/pull/1463/files/15ad3fd74067f54c30ba6858ee580b1aa9989c1e#diff-0c8f64dcb17357dbcfe03584129bc756eacfa7dd46e7c61767e46b6098a65544R64-R68): Returns true if the uses is suing a custom secret for the license instead of writing it explicitly in the `values.yaml`.

Ht implementation of `newrelic.common.region._fromValues` is [here](https://github.com/newrelic/helm-charts/pull/1463/files/15ad3fd74067f54c30ba6858ee580b1aa9989c1e#diff-35ca0af1666eaed8b3d266a1be796dab098eadcbd1582bd095fe4007542e963cR51-R59): Return the region from `values.yaml` if it is specified by the user.

The line were you left the comment basically test that use case you are pointing: If the user is using a secret for the license and the user does not set a region, as we will not be able to compute the region from the licese, [we fail](https://github.com/newrelic/helm-charts/pull/1463/files/15ad3fd74067f54c30ba6858ee580b1aa9989c1e#diff-35ca0af1666eaed8b3d266a1be796dab098eadcbd1582bd095fe4007542e963cR6).

The point onf this is to use the same UX for both charts as we are going to use on the super-agent chart. Take a look at the draft that is waiting for this to be merged: https://github.com/newrelic/helm-charts/pull/1466/files#diff-f752c985c9b04f49290d4a210d58d0981e8197c5825b7747defdc701fa4e1652R36-R54

dbudziwojskiNR (approved):

PR #1466, opened 2024-08-20
[super-agent] Improve helm chart UX (NR-281093)

kang-makes:
#### Is this a new chart
No

super-agent chart got the config from a field called `content` that added any arbitrary data to a config map. This is supposed to control this configMap and render it using values from Helm's `values.yaml`.

It also adds some protections so a customer is not able to use it on unsupported environments.

This PR is blocked until #1463 is merged.

#### Checklist
- [x] Chart Version bumped


kang-makes:
I had to remove from the chart the test for cluster's Kubernetes version as it was colliding with the CI so I moved it to the PR #1471

kang-makes:
The upgrade path is failing with this error:
```

**Comment by kang-makes** [comment]
The upgrade path is failing with this error:
```
Error: UPGRADE FAILED: template: super-agent/charts/super-agent-deployment/templates/deployment-superagent.yaml:24:34: executing "super-agent/charts/super-agent-deployment/templates/deployment-superagent.yaml" at <include (print $.Template.BasePath "/configmap-superagent-config.yaml") .>: error calling include: template: super-agent/charts/super-agent-deployment/templates/configmap-superagent-config.yaml:12:8: executing "super-agent/charts/super-agent-deployment/templates/configmap-superagent-config.yaml" at <include "newrelic-super-agent.config.content" .>: error calling include: template: super-agent/charts/super-agent-deployment/templates/_helpers.tpl:137:33: executing "newrelic-super-agent.config.content" at <include "newrelic-super-agent.config.agents.yaml" .>: error calling include: template: super-agent/charts/super-agent-deployment/templates/_helpers.tpl:18:8: executing "newrelic-super-agent.config.agents.yaml" at <fail (printf "Agent %s does not have agent type" $subAgentName)>: error calling fail: Agent open-telemetry does not have agent type
```

The previous `values.yaml` had this hardcoded: https://github.com/newrelic/helm-charts/pull/1466/files#diff-3bd49393f47bdfcb35b90c33237f384109da730bc892fa317ba8db58d29d7975L149-L158

Now, instead of having it in the values we have it as a helper: https://github.com/newrelic/helm-charts/pull/1466/files#diff-f752c985c9b04f49290d4a210d58d0981e8197c5825b7747defdc701fa4e1652R25-R34

The schema changed from this:
```yaml
    superAgent:
      content:
        agents:
          open-telemetry:
            agent_type: newrelic/io.opentelemetry.collector:0.2.0
    subAgents:
      open-telemetry:
        content:
          chart_values:
            licenseKey: ${nr-env:NR_LICENSE_KEY}
            cluster: ${nr-env:NR_CLUSTER_NAME}
```
To this:
```yaml
    subAgents:
      open-telemetry:
        type: newrelic/io.opentelemetry.collector:0.2.0
        content:
          chart_values:
            licenseKey: ${nr-env:NR_LICENSE_KEY}
            cluster: ${nr-env:NR_CLUSTER_NAME}
```
So the new chart is complaining that the old chart does not have `type` on the subagent.

The test is ok to fail as this is a breaking change. It should not have any repercussion to any customer as we do not support upgrades with `--reuse-values`.

charts/super-agent/Chart.yaml
```diff
@@ -3,19 +3,23 @@ name: super-agent
 description: Bootstraps New Relic' Super Agent
 
 type: application
-version: 0.0.18-beta
+version: 0.0.19-beta
+
+# TODO: Can we set renovatebot here with a regex?
+# Ref: https://github.com/fluxcd/flux2/blob/cc87ffd66e243fb85fc275792fa3708e44048048/cmd/flux/check.go#L62-L64
+kubeVersion: ">=1.28.0-0"
```
paologallinaharbur, line 10: is this a leftover of the other PR?

charts/super-agent/values.yaml
```diff
@@ -242,3 +247,14 @@ flux2:
...
+# -- Set of experimental configurations for super-agent chart.
+# @default -- See `values.yaml`
+experimental:
+  # Forces which is the minimal supported Kubernetes version for this chart without failing.
+  # Useful for development, testing, or if the chart is being templated in a unsupported way.
+  forceMinimumSupportedVersion: ""
+  # In case use uses `helm template`, helm uses the latest version supported by `helm`. This allows to
+  # change that version. Useful for testing, developing and to change some templates from subcharts that
+  # change manifests depending on the version of the cluster.
+  forceKubeVersion: ""
```
paologallinaharbur, line 260: Also this I think it is part of a different PR

charts/super-agent/charts/super-agent-deployment/templates/secret-sa-auth.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if include "newrelic-super-agent.auth.secret.shouldTemplate" . -}}
```
paologallinaharbur, line 1: i would add a comment explaining that usually we do not render it

charts/super-agent/charts/super-agent-deployment/templates/preinstall-job-register-system-identity.yaml
```diff
@@ -1,22 +1,54 @@
...
+  namespace: {{ .Release.Namespace }}
+data:
+  {{ include "newrelic.common.userKey.secretKeyName" . }}: {{ include "newrelic.common.userKey._userKey" . | b64enc }}
+{{- end }}
+---
 apiVersion: batch/v1
 kind: Job
 metadata:
   annotations:
     helm.sh/hook: pre-install # TODO we cannot enable auth after installation, we should add pre-upgrade and check if the secret exists
```
paologallinaharbur, line 28: it would be nice to fix this

charts/super-agent/Chart.yaml
kang-makes: Fixed on rebase

charts/super-agent/values.yaml
kang-makes: Fixed on rebase

charts/super-agent/charts/super-agent-deployment/templates/secret-sa-auth.yaml
kang-makes: I don't know if we are so sure of not rendering this so often... Why would you expect that clients will not embed the keys in the values file?

charts/super-agent/charts/super-agent-deployment/templates/preinstall-job-register-system-identity.yaml
kang-makes: Done [#e9a9af8](https://github.com/newrelic/helm-charts/pull/1466/commits/e9a9af8741a542fa0cc39ccc41e80b8c85528749)

charts/super-agent/values.yaml
```diff
@@ -94,71 +94,76 @@ super-agent-deployment:
...
-        #   port: 51200
-
+      # -- It you are a PowerUser, you can use this to override the configuration that has been created automatically by the chart.
+      # This configuration here will be **MERGED** with the configuration specified above.
+      # If you need to have you own configuration, disabled the creation of this configMap and create your own.
+      content: {}
 
     # -- Values that the fleet is going to have in the deployment.
-    # @default -- See `values.yaml` for examples
+    # @default -- {} (Empty. That defaults to configure the `newrelic/io.opentelemetry.collector` subagent)
```
paologallinaharbur, line 122: I think that this is not clear if you do not know what it is going on

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -2,19 +2,80 @@
...
 */ -}}
 {{- define "newrelic-super-agent.config.name" -}}
-{{- (include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" "local-data" "suffix" "superagent-config" )) -}}
+{{- include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" "local-data" "suffix" "superagent-config" ) -}}
 {{- end -}}
 
+
+
+{{- /*
+Return the agents part that should go in the super agent config. It is created from `.Values.config.subAgents`.
```
paologallinaharbur, line 11: I think that the comment is wrong. It creates the list of subAgents that is used for both the SA config and the SubAgents ones

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -2,19 +2,80 @@
...
+*/ -}}
+{{- define "newrelic-super-agent.config.agents.yaml" -}}
+{{- if (.Values.config).subAgents -}}
+{{- $agents := dict -}}
+{{- range $subAgentName, $subAgentConfig := (.Values.config).subAgents -}}
```
paologallinaharbur, line 16: why the () in  `(.Values.config).subAgents`?

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -2,19 +2,80 @@
...
+  {{- if not ($subAgentConfig).type -}}
+    {{- fail (printf "Agent %s does not have agent type" $subAgentName) -}}
+  {{- end -}}
+  {{- $_ := dict $subAgentName (dict "agent_type" $subAgentConfig.type "content" $subAgentConfig.content) | mustMerge $agents -}}
```
paologallinaharbur, line 20: it always confuses me when in the functions we do `| variable` 
isn't easier to read `$agents = mustMerge $agents dict ...`

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -2,19 +2,80 @@
...
+open-telemetry:
+  type: newrelic/io.opentelemetry.collector:0.2.0
+  content:
+    chart_values:
+      global:
+        licenseKey: ${nr-env:NR_LICENSE_KEY}
+        cluster: ${nr-env:NR_CLUSTER_NAME}
+        nrStaging: ${nr-env:NR_STAGING}
+        verboseLog: ${nr-env:NR_VERBOSE}
+        region: ${nr-env:NR_REGION}
```
paologallinaharbur, line 34: Thanks for adding all of them here!

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -2,19 +2,80 @@
...
-{{- else if .Values.euEndpoints -}}
-  https://system-identity-oauth.service.eu.newrelic.com/oauth2/token
-{{- else -}}
+{{- else if eq $region "EU" -}}
+  https://system-identity-oauth.service.newrelic.com/oauth2/token
+{{- else if eq $region "US" -}}
   https://system-identity-oauth.service.newrelic.com/oauth2/token
+{{- else if eq $region "Local" -}}
+  {{- /* Accessing the value directly without protection. A developer should now how to read the error. */ -}}
+  {{ .Values.development.backend.tokenRenewal }}
```
paologallinaharbur, line 76: I am wondering if this is this used already or useful (have you leveraged it when creating the script)?

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -2,19 +2,80 @@
...
+{{- $region := include "newrelic.common.region" . -}}
+
+{{- if eq $region "Staging" -}}
+  https://opamp.staging-service.newrelic.com/v1/opamp
+{{- else if eq $region "EU" -}}
+  https://opamp.service.eu.newrelic.com/v1/opamp
+{{- else if eq $region "US" -}}
+  https://opamp.service.newrelic.com/v1/opamp
+{{- else if eq $region "Local" -}}
+  {{- /* Accessing the value directly without protection. A developer should now how to read the error. */ -}}
```
paologallinaharbur, line 53: same here

charts/super-agent/values.yaml
```diff
@@ -94,71 +94,76 @@ super-agent-deployment:
...
-      # @default -- See `values.yaml` for examples
-      content:
-        agents:
-          open-telemetry:
-            agent_type: newrelic/io.opentelemetry.collector:0.2.0
-
-        # opamp:
-        #   # TODO The endpoint should be set automatically based on the licenseKey and on the nrStaging option if opamp.enable=true
-        #   endpoint: https://opamp.service.newrelic.com/v1/opamp
-        #   endpoint: https://opamp.service.eu.newrelic.com/v1/opamp
```
paologallinaharbur, line 137: I no longer see opamp defined in the values.yaml. How does it work? is it enabled by default?

paologallinaharbur: I think that we are enabling opamp if auth is enabled. Should we review UX with Jose? I find it confusing now

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -90,23 +170,130 @@ readOnlyRootFilesystem: true
...
+{{- end -}}
+
+
+
+{{- /*
+Check if .Values.config.auth.secret.private_key.secret_key exists and use it for the key in the secret containing the private
+key needed for the system identity. Fallbacks to `private_key`.
+*/ -}}
+{{- define "newrelic-super-agent.auth.secret.privateKey.key" -}}
+{{- $key := (((((.Values.config).auth).secret).private_key).secret_key) -}}
```
paologallinaharbur, line 238: `key` is confusing since it seems the `private_key`
also `secret_key` is not clear that is the key of the secret

MAybe a comment would help

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -90,23 +170,130 @@ readOnlyRootFilesystem: true
...
+{{- /*
+Check if .Values.config.auth.secret.client_id.secret_key exists and use it for the key in the secret containing the client id
+needed for the system identity. Fallbacks to `client_id`.
+*/ -}}
+{{- define "newrelic-super-agent.auth.secret.clientId.key" -}}
+{{- $key := (((((.Values.config).auth).secret).client_id).secret_key) -}}
+{{- if $key -}}
+  {{- $key -}}
+{{- else -}}
+  CLIENT_ID
```
paologallinaharbur, line 277: I am not sure if it worth to make these secret keys configurable, the work is already done, but maybe it would have been easier to just force the Key name

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
kang-makes: If `.Values.config` is `nil` and you try to access `subAgents` if panics and does not template.

But if you use `(.Values.config)` it returns "empty object" so you can access safely to `subAgents` that will return `nil`.

It is a safe mechanism as we cannot use [`dig` filter](http://masterminds.github.io/sprig/dicts.html#dig).

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
kang-makes: I might/will be useful for development purposes for fleet team and also for us of have mock servers were we can pass tests without a valid license key/system identity.

They are going to change the API, It is a tools that does 0 cost to maintain and is useful.

It also covers all the regions that the go client supports: https://github.com/newrelic/newrelic-client-go/blob/2785c31e58443ef35d5f3872b3b930d6d595225d/pkg/region/region_constants.go#L8-L21

charts/super-agent/values.yaml
kang-makes: Endpoint for OpAmp is calculated the same way as `tokenRenewal` and `systemIdentityRegistration`.

I'll ping José so he knows.

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
kang-makes: There are people that want a secret per license and people that want one secret for all licenses.

We have to make it configurable to cover both use cases.

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
kang-makes: It is commented on the `values.yaml`: https://github.com/newrelic/helm-charts/blob/15207de405080f265cc3fec28af867a1f5b135e0/charts/super-agent/values.yaml#L141-L166

Do we need to comment it again?

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
kang-makes: You are right. And not only that is that it was creating the agents incorrectly.

Fixed on [#7bb25f2](https://github.com/newrelic/helm-charts/pull/1466/commits/7bb25f25b56aa0f16e96d8b4c94d9f75a0f405ac)

Let me know what you think.

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
kang-makes: Fair. I changed all of them on [#d7c11af](https://github.com/newrelic/helm-charts/pull/1466/commits/d7c11af329768d346ab361f224a45d0e5ce1944c).

charts/super-agent/values.yaml
kang-makes: After an offline conversation, you were right. I changed here: [#8984a70](https://github.com/newrelic/helm-charts/pull/1466/commits/8984a702336fc8bd11bcdc37d28ed0f210b9f0e5)

srinivas-rayaprolu (approved):

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
paologallinaharbur: no, it is fine 😄

charts/super-agent/values.yaml
paologallinaharbur: I am not sure if we should error out if opamp=true while auth is not configured 🤔

paologallinaharbur: what do you think?

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
paologallinaharbur: Thanks it is better!

charts/super-agent/charts/super-agent-deployment/tests/configmap_fleet_configs_test.yaml
```diff
@@ -9,12 +9,12 @@ tests:
     set:
       config:
         subAgents:
-          open-telemetry: null
+          open-telemetry: ~
```
paologallinaharbur, line 12: what does it mean? can you add a comment?

charts/super-agent/values.yaml
paologallinaharbur: I think that if we say something: "If empty the chart add automatically the `newrelic/io.opentelemetry.collector` subagent. On the other hand, if populated the list of agent created is the one specified overwriting the default"

charts/super-agent/charts/super-agent-deployment/tests/configmap_superagent_config_test.yaml
```diff
@@ -113,7 +113,35 @@ tests:
               auth_config:
                 private_key_path: /etc/newrelic-super-agent/keys/from-secret.key
                 provider: local
-                token_url: https://system-identity-oauth.service.eu.newrelic.com/oauth2/token
+                token_url: https://system-identity-oauth.service.newrelic.com/oauth2/token
```
paologallinaharbur, line 139: This was changed already since they pushed for it, you will need to rebase 😕

charts/super-agent/values.yaml
kang-makes: From my point of view, auth is coupled to opamp. There is no way to use opamp without authenticating so the only toggle that is being used is the one at `.config.opamp.enabled`.

charts/super-agent/charts/super-agent-deployment/tests/configmap_fleet_configs_test.yaml
kang-makes: It is a left over. I was testing if `null` was being unmarshalled as a string instead of literally `null`.

`~` is the literal `null` in YAML as the spec says: https://yaml.org/type/null.html

> Shorthand:
> > `!!null`
> 
> Kind:
> > `Scalar`
> 
> Canonical:
>  > `~`
> 
> Regexp:
> > ```
> >      ~ # (canonical)
> >     |null|Null|NULL # (English)
> >     | # (Empty)
> > ```

charts/super-agent/charts/super-agent-deployment/tests/configmap_superagent_config_test.yaml
kang-makes: Rebased :)

charts/super-agent/values.yaml
kang-makes: Done [#f11ce5a](https://github.com/newrelic/helm-charts/pull/1466/commits/f11ce5aa8c6e8b9def8f699df1103b6f11ac4f96). Let me know what you think :)

charts/super-agent/values.yaml
```diff
@@ -94,71 +94,84 @@ super-agent-deployment:
...
+    #         verboseLog: ${nr-env:NR_VERBOSE}
+    #         region: ${nr-env:NR_REGION}
+    #         # you can set here modifications to the open telemetry chart
+
+    opamp:
+      # -- Enables or disables the auth against fleet control. It implies to disable any fleet communication and running the agent
+      # in stand alone mode where only the agents specified on `.config.subAgents` will be launched.
+      enabled: true
+
+    auth:
```
gsanchezgavier, line 146: i guess this was already discussed but shouldn't be auth inside opamp config?

charts/super-agent/values.yaml
```diff
@@ -94,71 +94,84 @@ super-agent-deployment:
...
-        # -- Port-forward it `$ kubectl port-forward pod/{pod-name} 51200:51200`
-        # -- And query it as `$ curl localhost:51200/status`
-        server:
-          enabled: true
-        #   port: 51200
-
-
-    # -- Values that the fleet is going to have in the deployment.
-    # @default -- See `values.yaml` for examples
+      # -- It you are a PowerUser, you can use this to override the configuration that has been created automatically by the chart.
```
gsanchezgavier, line 116: what is a PowerUser ? this might be misleading

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -2,19 +2,80 @@
...
+{{- range $subAgentName, $subAgentConfig := (.Values.config).subAgents -}}
+  {{- if not ($subAgentConfig).type -}}
+    {{- fail (printf "Agent %s does not have agent type" $subAgentName) -}}
+  {{- end -}}
+  {{- $agents = mustMerge $agents (dict $subAgentName $subAgentConfig) -}}
+{{- end -}}
+{{- $agents | toYaml -}}
+{{- else -}}
+{{- /* Default agents for Kubernetes */ -}}
+open-telemetry:
```
gsanchezgavier, line 25: out of curiosity , this default is on the helper to avoid the merge when custom ones are added? if so , isn't that against the expected behaviour of helm values ?

charts/super-agent/Chart.yaml
```diff
@@ -25,8 +25,6 @@ keywords:
 maintainers:
   - name: sigilioso
     url: https://github.com/sigilioso
-  - name: gsanchezgavier
```
gsanchezgavier, line 28: :)

charts/super-agent/values.yaml
kang-makes: No, it was not discussed. the discuss was that the `enabled` part should be in `opamp` instead of `auth`: https://github.com/newrelic/helm-charts/pull/1466#discussion_r1749977943

I am a bit reluctant to think that `auth` is going to be used only for `opamp` given the future I see about downloading packages and such.

This is just a feeling, if you don't feel the same way I have no issues on changing it.

charts/super-agent/values.yaml
kang-makes: I wanted to keep the possibility to create a config directly from the values that is passed to the super agent.

That is something that only somebody that know what is doing should use that value. I called it power user because it is __kind of a hacker__ let's say.

I am open to any suggestion on which would you like to have in that comment.

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
kang-makes: In Helm, objects are merged and lists are overridden. In that sense you are right.

But if we do not allow to override it this way, then the `open-telemetry` subagent is enabled always and that is no what we want.

This is a common behavior while creating complex defaults and we already do this with the `common-library` in [many](https://github.com/newrelic/helm-charts/blob/ddd7eb10b7530d3282f5469ddd3321a60192e524/library/common-library/templates/_nodeselector.tpl), [many](https://github.com/newrelic/helm-charts/blob/ddd7eb10b7530d3282f5469ddd3321a60192e524/library/common-library/templates/_low-data-mode.tpl), [many](https://github.com/newrelic/nri-kubernetes/blob/4545f562851f739b71ac86f9dc3d49cfd3a85b06/charts/newrelic-infrastructure/templates/kubelet/_security_context_helper.tpl#L25-L29) cases.

charts/super-agent/Chart.yaml
kang-makes: I used the users from `coreint` instead of `caos`. I'll add all caos team. Sorry.

charts/super-agent/values.yaml
gsanchezgavier: i would perhaps put it inside opamp just to be consistent with the SA config but not an strong opinion.

charts/super-agent/values.yaml
gsanchezgavier: make sense, but i would just mention what it does and maybe add some warns. 
```suggestion
      # -- Overrides the configuration that has been created automatically by the chart.
```

gsanchezgavier (approved):
I just left nits, feel free to discard them,  LGTM thanks!

charts/super-agent/Chart.yaml
kang-makes: Done in [#d598ded](https://github.com/newrelic/helm-charts/pull/1466/commits/d598dedf47d32be2ea9e055031faf859b75386fe)

charts/super-agent/values.yaml
kang-makes: Done in [#d598ded](https://github.com/newrelic/helm-charts/pull/1466/commits/d598dedf47d32be2ea9e055031faf859b75386fe)

charts/super-agent/charts/super-agent-deployment/templates/preinstall-job-register-system-identity.yaml
```diff
@@ -62,7 +101,7 @@ spec:
 
               ERROR_MESSAGE=$(jq -r '.errors[0].message // "NOERROR"' "$TEMPORAL_FOLDER/response.json")
               if [ "$ERROR_MESSAGE" != "NOERROR" ]; then
-                echo "Error creating an identity: $ERROR_MESSAGE"
+                echo "Failed to create a New Relic System Identity for OpAMP communication authentication. Please verify that your User Key is valid and that your Account Organization has the necessary permissions to create a System Identity: $ERROR_MESSAGE"
```
gsanchezgavier, line 104: cool thanks!

gsanchezgavier (approved):

PR #1472, opened 2024-09-03
[nr-k8s-otel-collector] Honor the LowDataMode toogle from the common-library

kang-makes:
#### Is this a new chart
No

The value for `lowDataMode` was hard coded at values level so a user cannot set the value if you not go deep down to the chart and patch it.

This change has its own implementation of the lowDataMode helper that defaults to `true` while honoring the behaviour of the common-library of setting a value a `global` level.

#### Checklist
- [x] Chart Version bumped


csongnr (approved):

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -153,5 +153,5 @@ receivers:
     enabled: true
 
 # -- (bool) Send only the [metrics required](https://github.com/newrelic/helm-charts/tree/master/charts/nr-k8s-otel-collector/docs/metrics-lowDataMode.md) to light up the NR kubernetes UI, this agent defaults to setting lowDataMode true, but if this setting is unset, lowDataMode will be set to false
```
dbudziwojskiNR, line 155: After this change, wouldn't `lowDataMode` be true if unset? If so, we should update this wording.

dbudziwojskiNR (changes requested):

charts/nr-k8s-otel-collector/values.yaml
dbudziwojskiNR: @kang-makes Bumping this :) Thanks!

dbudziwojskiNR (dismissed):

dbudziwojskiNR (approved):

PR #1484, opened 2024-09-23
[super-agent] Add tests to the pre-install job

kang-makes:
#### Is this a new chart
No

Chart with missing values is redering while it should fail.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
+++ b/charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
@@ -175,9 +175,9 @@ Return .Values.config.auth.organizationId and fails if it does not exists
 */ -}}
 {{- define "newrelic-super-agent.auth.organizationId" -}}
 {{- if (((.Values.config).opamp).auth).organizationId -}}
-  {{- .Values.config.auth.organizationId -}}
+  {{- .Values.config.opamp.auth.organizationId -}}
 {{- else -}}
-  {{- fail ".config.auth.organizationId is required." -}}
+  {{- fail ".config.opamp.auth.organizationId is required" -}}
 {{- end -}}
 {{- end -}}
 
--- a/charts/super-agent/Chart.lock
+++ b/charts/super-agent/Chart.lock
@@ -4,9 +4,9 @@ dependencies:
   version: 2.13.0
 - name: super-agent-deployment
   repository: ""
-  version: 0.0.24-beta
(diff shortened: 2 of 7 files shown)


paologallinaharbur:
I am adding as well two small fixes:
 - the job serviceAccount and roleBinding are missing the hook pre-updagrade
 - the way we are mounting the volume now is leading to an issue:
```
        message: 'failed to create task for container: failed to create shim task:
          OCI runtime create failed: runc create failed: unable to start container
          process: error during container init: error mounting "/var/lib/kubelet/pods/d7822316-18c0-450d-8f8c-b3d6017dd0b8/volume-subpaths/auth-secret-private-key/super-agent-deployment/2"
          to rootfs at "/etc/newrelic-super-agent/keys/from-secret.key": mount /var/lib/kubelet/pods/d7822316-18c0-450d-8f8c-b3d6017dd0b8/volume-subpaths/auth-secret-private-key/super-agent-deployment/2:/etc/newrelic-super-agent/keys/from-secret.key
          (via /proc/self/fd/6), flags: 0x5001: not a directory: unknown: Are you
          trying to mount a directory onto a file (or vice-versa)? Check if the specified
          host path exists and is the expected type'
```
 - `organization_id` -> `organizationId` in values.yaml

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -29,9 +29,9 @@ open-telemetry:
       global:
         licenseKey: ${nr-env:NR_LICENSE_KEY}
         cluster: ${nr-env:NR_CLUSTER_NAME}
-        nrStaging: ${nr-env:NR_STAGING}
-        verboseLog: ${nr-env:NR_VERBOSE}
-        region: ${nr-env:NR_REGION}
+      {{- if include "newrelic.common.nrStaging" . }}
+        nrStaging: true
+      {{- end -}}
```
sigilioso, line 34: Should we add a comment explaining why env-vars with boolean values cannot be used here?

sigilioso (approved):
LGTM

I simply left a question in case we'd like to add a further explanation.

Thanks!

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
paologallinaharbur: I do not think it is the proper place 😕

PR #1490, opened 2024-10-02
[nr-ebpf-agent] Add option to configure push period.

benkilimnik:
~This PR has a hard dependency on updating the ebpf client version -- need cut a new release for that first.~ Done in #1493 

#### Is this a new chart
No

Adds option to configure push period from the ebpf agent. The eBPF agent applies a request path clustering algorithm to reduce cardinality in exported HTTP data. The algorithm only looks for similar request paths within data of the same push period. This change enables customers to increase the window under consideration for cardinality reduction if needed.

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/nr-ebpf-agent/values.yaml
+++ b/charts/nr-ebpf-agent/values.yaml
@@ -23,7 +23,7 @@ protocols:
   cass: true
   dns: true
   pgsql: true
-# -- The periodicity in seconds at which the eBPF client pushes data to the OTel collector for export to NR.
+# -- The periodicity in seconds at which the eBPF agent pushes data to the OTel collector for export to NR.
 # The eBPF agent applies a request path clustering algorithm to reduce cardinality in exported HTTP data.
 # The algorithm only looks for similar request paths within data of the same push period.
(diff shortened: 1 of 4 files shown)


charts/nr-ebpf-agent/values.yaml
```diff
@@ -23,6 +23,11 @@ protocols:
   cass: true
   dns: true
   pgsql: true
+# -- The periodicity in seconds at which the eBPF client pushes data to the OTel collector for export to NR.
```
kpattaswamy, line 26: minor but should it be this?
```suggestion
# -- The periodicity in seconds at which the eBPF agent pushes data to the OTel collector for export to NR.
```

kpattaswamy (approved):

charts/nr-ebpf-agent/values.yaml
benkilimnik: Applied. A little ambiguous from a customer perspective (since it is marketed as a single ebpf "agent") but technically correct.

kpattaswamy (dismissed):

kpattaswamy (approved):
Is this ready to land?

PR #1523, opened 2024-11-07
[nr-ebpf-agent] Support nri-bundle supplied secrets. 

Philip-R-Beckwith:
#### Is this a new chart
No.

This chart adds support for some environment variables/secrets that are documented in the NRI-bundle. 
This removes the need for the user to supply separate secrets specifically for the eBPF agent. 
This change will allow for a more seamless install for the customers who choose to install the eBPF agent as part of the nri bundle. (When the eBPF agent is added to the bundle.)

  - fixes # N/A

Support for global ingest keys is documented in the nri-bundle/read.me. 
If users are installing this chart by itself, they should use the values.licenseKey not global vars supplied by the bundle.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/nr-ebpf-agent/templates/secrets.yaml
+++ b/charts/nr-ebpf-agent/templates/secrets.yaml
@@ -1,10 +1,10 @@
 {{- $licenseKey := include "nr-ebpf-agent.licenseKey" . -}}
 {{- $customSecretLicenseKey := include "nr-ebpf-agent.customSecretKey" . -}}
-
 apiVersion: v1
 kind: Secret
 metadata:
   name: nr-ebpf-agent-secrets
+type: Opaque
 data:
   {{- if .Values.licenseKey}}
   NR_LICENSE_KEY: {{ .Values.licenseKey | b64enc }}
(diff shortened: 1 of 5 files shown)


charts/nr-ebpf-agent/templates/secrets.yaml
```diff
@@ -1,7 +1,17 @@
...
   name: nr-ebpf-agent-secrets
+type: Opaque
 data:
+  {{- if .Values.licenseKey}}
+  NR_LICENSE_KEY: {{ .Values.licenseKey | b64enc }}
+  {{- else if $licenseKey}}
+  NR_LICENSE_KEY: {{ $licenseKey | b64enc }}
+  {{ else if $customSecretLicenseKey}}
+  NR_LICENSE_KEY: ""
+  {{- else }}
```
kpattaswamy, line 15: From my understanding, the first and last condition are identical with the exception of the last one throwing a log expressing that the key is required. Perhaps we could remove the last one if that's that case?

Philip-R-Beckwith: If we get to the last case it should always throw. If we managed to get there it means we didn't have a license key a global license key or a custom secrets path. 

The Since a pre condition to get to this step is that the license key is empty, required will always throw. 
We could reduce one of the control blocs by changing the order, but that'd make it so that the local license key isn't the prefered secret if it's available. 

We can't move the required statement up a level since we don't want to insert a value for NR_LICENSE_KEY when we skip throwing when get the secret from customSecretLicenseKey values.

kpattaswamy (approved):

PR #1542, opened 2024-12-10
feat: exposed addituional env variables in newrelic logging charts

voorepreethi:
  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/newrelic-logging/Chart.yaml
+++ b/charts/newrelic-logging/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart to deploy New Relic Kubernetes Logging as a DaemonSet, supporting both Linux and Windows nodes and containers
 name: newrelic-logging
-version: 1.23.5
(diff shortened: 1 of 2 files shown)


charts/newrelic-logging/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart to deploy New Relic Kubernetes Logging as a DaemonSet, supporting both Linux and Windows nodes and containers
 name: newrelic-logging
-version: 1.23.5
+version: 1.23.6
```
rohit-bandlamudi-nr, line 4: Why is the version bumped up to 1.23.6. We want to have a version suffixed with something like "-beta" to the version name. And we should tag this helm chart with the same. So as to be able to install it with something like **helm install --version "some_version-beta"**

rohit-bandlamudi-nr (changes requested):

PR #1546, opened 2024-12-12
[super-agent] L2 identity creation

rubenruizdegauna:
#### Is this a new chart
No
This PR adds L2 identity creation from L1 ClientID and ClientSecret. 
It maintains backwards compatibility so `userKey` can still be used, but if ClientID and ClientSecret are present, the latter will be used.

As with `userKey`, a custom secret can be used to pass the ClientID and ClientSecret.

  - fixes https://new-relic.atlassian.net/browse/NR-343393

#### Checklist
- [x] Chart Version bumped


charts/super-agent/README.md
```diff
@@ -53,74 +61,85 @@ super-agent-deployment:
...
+|------------------------------------------------------------------------|--------|------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
paologallinaharbur, line 81: Something is weird you have a super long sentence I think

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -283,3 +283,116 @@ value is provided, it defaults to `""` (empty string) so this helper can be used
   {{- /* Empty string */ -}}
 {{- end -}}
 {{- end -}}
+
+{{/* check if both L1 ClientID and ClientSecret are provided */}}
+{{- define "newrelic-super-agent.auth.l1Identity" -}}
+{{- if and (include "newrelic-super-agent.auth.l1IdentityClientId" .) (include "newrelic-super-agent.auth.l1IdentityClientSecret" .) -}}
+    true
```
paologallinaharbur, line 290: you can decide if you want to fail or show any message if just one is provided.
Or write a mesasge to the cusotmer depending on which auth is enabled

charts/super-agent/values.yaml
```diff
@@ -19,6 +19,15 @@ super-agent-deployment:
   # -- In case you don't want to have the license key in you values, this allows you to point to which secret key is the license key located. Can be configured also with `global.customSecretLicenseKey`
   customSecretLicenseKey: ""
 
+  # -- L1 identity client_id to use. Can be configured also with `global.l1IdentityClientId`
```
paologallinaharbur, line 22: why supporting also "global"? I think that no other chart have it!

Globals are the ones supported by the common library

charts/super-agent/values.yaml
```diff
@@ -19,6 +19,15 @@ super-agent-deployment:
...
+  l1IdentityClientId: ""
+  # -- L1 identity client_secret to use. Can be configured also with `global.l1IdentityClientSecret`
+  l1IdentityClientSecret: ""
+  # -- In case you don't want to have the client_id and client_secret in you values, this allows you to point to a user created secret to get the key from there. Can be configured also with `global.customL1IdentityClientSecretKey`
+  customL1IdentitySecretName: ""
+  customL1IdentityClientIdSecretKey: ""
+  customL1IdentityClientSecretSecretKey: ""
```
paologallinaharbur, line 29: below we have already:
```
      auth:
        # -- Organization ID where fleets will live.
        organizationId: ""
        secret:
          create: true
          # -- Name auth' secret provided by the user. If the creation of this secret is set to `true`, this is the same the secret
          # will have.
          # @default -- release name suffixed with "-auth"
          name:
          private_key:
            # -- Key inside the secret containing the private key.
            # @default -- `private_key`
            secret_key:
            # -- In case `.config.auth.secret.create` is true, you can set these keys to set private key directly as base64.
            # This options is mutually exclusive with `plain_pem`.
            base64_pem:
            # -- In case `.config.auth.secret.create` is true, you can set these keys to set private key directly as plain text.
            # This options is mutually exclusive with `base64_pem`.
            plain_pem:
          client_id:
```

why it is not in the auth section?

paologallinaharbur: is it deprecated such section? if so we should mark it accordingly

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
paologallinaharbur: we can also decide to fail if old and new auth is configured at the same time

charts/super-agent/README.md
rubenruizdegauna: The sentence was already there, I just formatted it 🤷‍♂️ 
I'll take a look

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
rubenruizdegauna: I added the check in the template, not in the helper. So basically is:
* at least one method should be provided: userKey or clientId/clientSecrert, it will fail if none
* if both are provided env variables for both are created and only clientId/clientSecret will be used in the shell script

sdyt?

charts/super-agent/values.yaml
rubenruizdegauna: The client_id in that section is the L2 client_id, and I thought that it was used for configuration. The other ones are in the same level of userKey, that are only used for L1 creation. I can move it but I'm not sure as userKey was already out of this section.

Quick sync discussion?

charts/super-agent/README.md
rubenruizdegauna: fixed in d455a3a6c181a915d7c093fdf4d3b3bc0dceddcf

charts/super-agent/values.yaml
paologallinaharbur: yes, otherwise it is really confusing, someone could think that such global would be available in all charts

charts/super-agent/charts/super-agent-deployment/templates/preinstall-job-register-system-identity.yaml
```diff
@@ -79,45 +112,130 @@ spec:
...
-                        publicKey: \"'$(openssl enc -base64 -A -in "$TEMPORAL_FOLDER/pub")'\"
-                      ) {
-                        clientId,
-                        name
-                      }
-                    }"
-                  }' | tr -d $'\n' | \
-                  curl \
-                    -w "%{http_code}" \
+              if [ "${NEW_RELIC_AUTH_CLIENT_ID:-}" != "" ] && [ "${NEW_RELIC_AUTH_CLIENT_SECRET:-}" != "" ]; then
```
paologallinaharbur, line 115: I am not reviewing this snippet, I understand it is a copy/paste from onhost solution

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -283,3 +283,116 @@ value is provided, it defaults to `""` (empty string) so this helper can be used
...
+{{- else if .Values.global -}}
+  {{- if .Values.global.customL1IdentityClientSecretSecretKey }}
+    {{- .Values.global.customL1IdentityClientSecretSecretKey -}}
+  {{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "newrelic-super-agent.auth.generatedIdentityCredentialsSecretName" -}}
+{{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "preinstall-client-credentials" ) }}
+{{- end -}}
```
paologallinaharbur, line 387: nit this is used by the function below, I would place it just below that

charts/super-agent/values.yaml
rubenruizdegauna: There you have 69b6a282d20432dc152b74f82da0d37a6e759c2c :)

charts/super-agent/values.yaml
```diff
@@ -19,6 +19,15 @@ super-agent-deployment:
   # -- In case you don't want to have the license key in you values, this allows you to point to which secret key is the license key located. Can be configured also with `global.customSecretLicenseKey`
   customSecretLicenseKey: ""
 
+  # -- L1 identity client_id to use. Can be configured also with `global.l1IdentityClientId`
+  l1IdentityClientId: ""
+  # -- L1 identity client_secret to use. Can be configured also with `global.l1IdentityClientSecret`
+  l1IdentityClientSecret: ""
+  # -- In case you don't want to have the client_id and client_secret in you values, this allows you to point to a user created secret to get the key from there. Can be configured also with `global.customL1IdentityClientSecretKey`
```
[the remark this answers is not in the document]
rubenruizdegauna: 🕵️  thanks 6f93111c28fa65fea415ca52a48a970b34aff504

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
rubenruizdegauna: I dedicated this commit to you 😗  c3cdee864ca2b501b5f1b945abdb8d2b90be10cb

charts/super-agent/values.yaml
paologallinaharbur: there is still this comment!

paologallinaharbur: and the ones below

charts/super-agent/README.md
```diff
@@ -94,6 +100,9 @@ As of the creation of the chart, it has no particularities and this section can
...
+| super-agent-deployment.customL1IdentitySecretName | string | `""` | L1 ClientID and ClientSecret holding Secret name. Can be configured also with `global.customL1IdentitySecretName`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```
paologallinaharbur, line 105: also all these

charts/super-agent/charts/super-agent-deployment/templates/_helpers.tpl
```diff
@@ -283,3 +283,100 @@ value is provided, it defaults to `""` (empty string) so this helper can be used
...
+
+{{/* return L1 ClientID with global fallback */}}
+{{- define "newrelic-super-agent.auth.l1IdentityClientId" -}}
+{{- if .Values.l1IdentityClientId -}}
+  {{- .Values.l1IdentityClientId -}}
+{{- end -}}
+{{- end -}}
+
+{{/* return L1 ClientSecret with global fallback */}}
```
paologallinaharbur, line 301: and these

charts/super-agent/values.yaml
rubenruizdegauna: Thanks! :wasn't-me: 20e5d87d66e392c29c4f5a2caa9594de9d66e435

also remved from README

paologallinaharbur (approved):

PR #1520, opened 2024-11-04
Adding ebpf agent to bundle

Philip-R-Beckwith:
#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [ ] Chart Version bumped
- [ ] Variables are documented in the README.md
- [ ] Title of the PR starts with chart name (e.g. `[mychartname]`)

PR #1579, opened 2025-01-22
[nri-bundle] Add eAPM to NRI Bundle

kpattaswamy:
#### Is this a new chart
No
Adds the changes from https://github.com/newrelic/helm-charts/pull/1520
  - fixes #

#### Checklist
- [x] Chart Version bumped


ramkrishankumarN (approved):

charts/nri-bundle/Chart.yaml
```diff
@@ -65,6 +66,11 @@ dependencies:
     condition: newrelic-pixie.enabled
     version: 2.1.6
 
+  - name: nr-ebpf-agent
+    version: 0.1.14
+    condition: newrelic-eapm-agent.enabled
```
dbudziwojskiNR, line 71: Are we sure we want to stray away from the chart name? In other words, are we sure we want to use `newrelic-eapm-agent` over `nr-ebpf-agent`? Having a different chart vs config name has led to confusion in the past.

kpattaswamy: I believe this was an intentional decision to have the underlying chart named `nr-ebpf-agent` but represented as `newrelic-eapm-agent` externally.

Tagging @ramkrishankumarN to confirm

ramkrishankumarN: I think that's right -- confirmed with product. `newrelic-eapm-agent` (eAPM) is the external name.


we can keep the ebpf naming as it does more than APM (in future) internally.

ramkrishankumarN (approved):

PR #1589, opened 2025-01-28
[ebpf-agent] Remove unnecessary selector labels

Philip-R-Beckwith:
#### Is this a new chart
NO.
Removes selector labels that we don't need. 

Installed the chart without the new selectors and got data: 
Screenshot 2025-01-28 at 12 24 33 PM

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/nr-ebpf-agent/templates/nr-ebpf-agent-daemonset.yaml
+++ b/charts/nr-ebpf-agent/templates/nr-ebpf-agent-daemonset.yaml
@@ -10,8 +10,6 @@ metadata:
 spec:
   selector:
     matchLabels:
-      app: nr-ebpf-agent
-      component: agent
       {{- include "newrelic.common.labels.selectorLabels" . | nindent 6 }}
   template:
     metadata:


charts/nr-ebpf-agent/templates/nr-ebpf-agent-daemonset.yaml
```diff
@@ -10,8 +10,6 @@ metadata:
 spec:
   selector:
     matchLabels:
-      app: nr-ebpf-agent
```
kpattaswamy, line 13: Do we need to remove these labels from the [nr-ebpf-agent-service.yaml](https://github.com/newrelic/helm-charts/blob/07a0fc171c3207df0571ca496f45c7626da329d7/charts/nr-ebpf-agent/templates/nr-ebpf-agent-service.yaml)? I believe the service/labels were added to configure TLS comm. between the agent/collector

Philip-R-Beckwith: The other option is to rename the daemonset and make any necessary changes throughout the Helm chart to facilitate that change. 
However, considering that this is a daemonset and we're already using the selector 
`{{- include "newrelic.common.labels" . | nindent 4 }}` I do not believe we needed to include these selectors here. 
Also considering that the service that was created along with the implementation of TLS uses the selector `{{- include "newrelic.common.labels" . | nindent 4 }}` not  `app: nr-ebpf` is evidence that these labels/ selectors were added without any real functionality. 

Since both the service and this daemonset both use the selector 
`{{- include "newrelic.common.labels" . | nindent 4 }}` we should be good to delete this selector.

kpattaswamy (approved):

PR #1617, opened 2025-02-12
[agent-control] fix: Added backofflimit and ignore-not-found to the uninstall-job

alvarocabanas:
- Agent-control : Added backofflimit and ignore-not-found to the uninstall-job

### DIFF
--- a/charts/agent-control/charts/agent-control-deployment/templates/uninstall-job.yaml
+++ b/charts/agent-control/charts/agent-control-deployment/templates/uninstall-job.yaml
@@ -19,6 +19,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   ttlSecondsAfterFinished: 120
+  backoffLimit: 3
   template:
     spec:
       restartPolicy: Never
@@ -49,6 +50,6 @@ spec:
       
               {{- if include "newrelic-agent-control.auth.secret.shouldRunJob" . }}
               # Delete the secrets created in the cluster
-              kubectl -n {{ $.Release.Namespace }} delete secret {{ include "newrelic-agent-control.auth.secret.name" . }}
+              kubectl delete secrets -n {{ .Release.Namespace }} -l "nr-generic-secret=true"
               {{- end }}
 {{- end -}}
--- a/charts/agent-control/Chart.lock
+++ b/charts/agent-control/Chart.lock
@@ -4,9 +4,9 @@ dependencies:
   version: 2.14.1
 - name: agent-control-deployment
   repository: ""
-  version: 0.0.39-beta
+  version: 0.0.40-beta
 - name: common-library
   repository: https://helm-charts.newrelic.com
   version: 1.3.1
-digest: sha256:9f32bf1c7e1c5e5fe4bf0ca1d48ee89155816667971393d7d9ad147dc8bb7768
-generated: "2025-02-17T08:34:45.995478+01:00"
+digest: sha256:81728940ef40bb132f63a5beb9f419397a370191fdd3025332bf819c2159443c
+generated: "2025-02-17T10:50:05.543595+01:00"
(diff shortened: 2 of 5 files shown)


charts/agent-control/charts/agent-control-deployment/templates/uninstall-job.yaml
```diff
@@ -49,6 +50,6 @@ spec:
       
               {{- if include "newrelic-agent-control.auth.secret.shouldRunJob" . }}
               # Delete the secrets created in the cluster
-              kubectl -n {{ $.Release.Namespace }} delete secret {{ include "newrelic-agent-control.auth.secret.name" . }}
+              kubectl -n {{ $.Release.Namespace }} delete secret --ignore-not-found {{ include "newrelic-agent-control.auth.secret.name" . }}
```
rubenruizdegauna, line 53: What about the other resources that we try to delete above? Would it be worth to add the `--ignore-not-found`?

alvarocabanas: The other ones use labels so they don't fail.

gsanchezgavier: couldn't be this the secret provided ? I would expect we label any resource we create and then remove only those ones, but not the ones referenced

alvarocabanas: I don't understand what you comment, do you mean to also add a label to the secret? I am just adding this ignore-not-found because this was already created like this.

alvarocabanas: mmm... I see, then we should see if we can label it if not created by the user and only delete it in that case, and since we would be using a label then we also avoid having to add the flag, I will give it a look.

alvarocabanas: I fixed it in last commit.

gsanchezgavier: Is this needed? i see that we are rendering this secret, why isn't being remove on uninstall?https://github.com/newrelic/helm-charts/blob/1ad80e482b932c880eabbd730fb3dbfb55c1b5cc/charts/agent-control/charts/agent-control-deployment/templates/secret-sa-auth.yaml#L5

gsanchezgavier: perhaps in the past the script was creating the secret and that is why we removed manually? or maybe i'm missing something

sigilioso: I believe we are still creating it manually (depending on how values are set up).

The secret is only rendered when it is specified in values

https://github.com/newrelic/helm-charts/blob/36a8f5bf7bfc0ce9aea441537d982f577d1f5a8b/charts/agent-control/charts/agent-control-deployment/templates/_helpers.tpl#L208-L209

Otherwise, it is created manually

https://github.com/newrelic/helm-charts/blob/36a8f5bf7bfc0ce9aea441537d982f577d1f5a8b/charts/agent-control/charts/agent-control-deployment/templates/preinstall-job-register-system-identity.yaml#L281-L287

alvarocabanas: The templated one is only if specified on the values and is either that one or the one by the script https://github.com/newrelic/helm-charts/blob/36a8f5bf7bfc0ce9aea441537d982f577d1f5a8b/charts/agent-control/values.yaml#L166.
So this will only delete the one created by the script.

gsanchezgavier: thanks for clarifying this i missed that in some cases the job is creating the secret 👍 , then it make sense to remove it from the unninstall job

gsanchezgavier (approved):
thanks

PR #1629, opened 2025-02-21
fix: fixed test cases and empty cluster name

voorepreethi:
This Pull Request addresses a bug where the cluster name was not included in the values file, resulting in an inability to append an empty cluster name when adding labels. This fix ensures that an empty cluster name is appended by default, thereby allowing labels to be correctly appended in the Prometheus scraper configuration.

Details
Issue: The absence of a cluster name in the values file prevented the addition of an empty value for the cluster name during the label configuration in Prometheus scrapers.
Fix: The values file has been updated to include an empty cluster name by default. This update facilitates the proper addition of labels, even when no specific cluster name is provided.
Impact: Ensures that Prometheus scrapers can append labels successfully, thereby maintaining consistent and accurate monitoring metrics.
  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/newrelic-logging/templates/_helpers.tpl
+++ b/charts/newrelic-logging/templates/_helpers.tpl
@@ -205,7 +205,12 @@ Returns fluentbit config to collect and forward its metrics to New Relic
     add_label            source kubernetes
     add_label            pod_name ${HOSTNAME}
     add_label            node_name ${NODE_NAME}
-    {{- printf "add_label            cluster_name %s" (include "newrelic-logging.cluster" .) | nindent 4 -}}
+    {{- $clusterName := (include "newrelic-logging.cluster" .) -}}
+    {{- if $clusterName -}}
+    {{- printf "add_label            cluster_name %s" $clusterName | nindent 4 -}}
+    {{- else -}}
+    {{- printf "add_label            cluster_name \"%s\"" $clusterName | nindent 4 -}}
+    {{- end -}}
     {{- printf "add_label            namespace %s" .Release.Namespace | nindent 4 -}}
     {{- printf "add_label            daemonset_name %s" (include "newrelic-logging.fullname" .) | nindent 4 -}}
 {{- end -}}
--- a/charts/newrelic-logging/tests/fluentbit_monitoring_config_test.yaml
+++ b/charts/newrelic-logging/tests/fluentbit_monitoring_config_test.yaml
@@ -5,10 +5,17 @@ release:
   name: my-release
   namespace: my-namespace
 tests:
-  - it: "additional labels for entity synthesis are included by default"
-    templates:
-      - templates/configmap.yaml
+  - it: "additional labels for entity synthesis are included by default with cluster name empty"
+    set:
+      cluster: ""
     asserts:
       - matchRegex:
           path: data["fluent-bit.conf"]
-          pattern: (?s)add_label\s+cluster_name \n\s*add_label\s+namespace my-namespace\n\s*add_label\s+daemonset_name my-release-newrelic-logging
+          pattern: (?s)add_label\s+cluster_name \"\"\n\s*add_label\s+namespace my-namespace\n\s*add_label\s+daemonset_name my-release-newrelic-logging
+  - it: "additional labels for entity synthesis are included by default when cluster name is set"
+    set:
+      cluster: my-cluster
+    asserts:
+      - matchRegex:
+          path: data["fluent-bit.conf"]
+          pattern: (?s)add_label\s+cluster_name my-cluster\n\s*add_label\s+namespace my-namespace\n\s*add_label\s+daemonset_name my-release-newrelic-logging
--- a/charts/newrelic-logging/tests/fluentbit_monitoring_config_test.yaml
+++ b/charts/newrelic-logging/tests/fluentbit_monitoring_config_test.yaml
@@ -5,19 +5,21 @@ release:
   name: my-release
   namespace: my-namespace
 tests:
-  - it: "additional labels for entity synthesis are included by default with cluster name empty"
+  - it: "includes additional labels with cluster name empty"
     set:
       cluster: ""
     asserts:
       - matchRegex:
           path: data["fluent-bit.conf"]
           pattern: (?s)add_label\s+cluster_name \"\"\n\s*add_label\s+namespace my-namespace\n\s*add_label\s+daemonset_name my-release-newrelic-logging
-  - it: "additional labels for entity synthesis are included by default with cluster name null"
+
+  - it: "includes additional labels with cluster name null"
     asserts:
       - matchRegex:
           path: data["fluent-bit.conf"]
           pattern: (?s)add_label\s+cluster_name \"\"\n\s*add_label\s+namespace my-namespace\n\s*add_label\s+daemonset_name my-release-newrelic-logging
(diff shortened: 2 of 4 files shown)


charts/newrelic-logging/tests/fluentbit_monitoring_config_test.yaml
```diff
@@ -5,10 +5,38 @@ release:
   name: my-release
   namespace: my-namespace
 tests:
-  - it: "additional labels for entity synthesis are included by default"
-    templates:
-      - templates/configmap.yaml
+  - it: "includes additional labels with cluster name empty"
```
rohit-bandlamudi-nr, line 8: nit: Update test cases names to include info about cluster value and the resulting cluster_name

1. cluster_name is "" when cluster is not set
2. cluster_name is "" when cluster is set with empty value
3. cluster_name is "" when cluster is set with null value
4. cluster_name is "" when cluster is set with empty string
5. cluster_name is set properly when cluster is set with non-empty

rohit-bandlamudi-nr (changes requested):

charts/newrelic-logging/tests/fluentbit_monitoring_config_test.yaml
rohit-bandlamudi-nr: Also include another case for quoted cluster name. for eg. cluster: "test"

maya-jha: Maybe instead of saying 
`cluster_name is "" when cluster is not set` we  can say its empty `cluster_name is empty when cluster is not set`

charts/newrelic-logging/templates/_helpers.tpl
```diff
@@ -205,7 +205,12 @@ Returns fluentbit config to collect and forward its metrics to New Relic
     add_label            source kubernetes
     add_label            pod_name ${HOSTNAME}
     add_label            node_name ${NODE_NAME}
-    {{- printf "add_label            cluster_name %s" (include "newrelic-logging.cluster" .) | nindent 4 -}}
+    {{- $clusterName := (include "newrelic-logging.cluster" .) -}}
```
maya-jha, line 208: Where does it get deployed if we don't provide cluster name? Does it take it from kube config? How about other params , can they be null too?

charts/newrelic-logging/tests/fluentbit_monitoring_config_test.yaml
```diff
@@ -5,10 +5,38 @@ release:
...
+    set:
+      cluster: ""
```
maya-jha, line 10: should we test based on `[newrelic-logging.cluster](https://github.com/newrelic/helm-charts/blob/85037884f7e3c5a975a47063ab724e20f9c4e124/charts/newrelic-logging/templates/_helpers.tpl#L63)` instead of cluster? Also, I think we may need to just test based on empty. We can add a test for `newrelic-logging.cluster` itself if it does not exist.

charts/newrelic-logging/templates/_helpers.tpl
```diff
@@ -205,7 +205,12 @@ Returns fluentbit config to collect and forward its metrics to New Relic
...
+    {{- if $clusterName -}}
+    {{- printf "add_label            cluster_name %s" $clusterName | nindent 4 -}}
+    {{- else -}}
+    {{- printf "add_label            cluster_name \"%s\"" $clusterName | nindent 4 -}}
+    {{- end -}}
     {{- printf "add_label            namespace %s" .Release.Namespace | nindent 4 -}}
```
maya-jha, line 214: can any other label be also null/empty?

maya-jha (changes requested):

charts/newrelic-logging/templates/_helpers.tpl
voorepreethi: charts won't get deployed without release name . So other parameters are mandatory . Cluster name is something which we are fetching not from the deployed cluster but from values

charts/newrelic-logging/templates/_helpers.tpl
rohit-bandlamudi-nr: other params, cant be null as they will fetched from the generated k8s manifests. Cluster name would be picked up from values file

charts/newrelic-logging/tests/fluentbit_monitoring_config_test.yaml
rohit-bandlamudi-nr: The value `newrelic-logging.cluster` is defined by us in the helpers.tpl file and its value depends on the value given in values file. This test is sufficient as we can know the resultant config based on value  of cluster given in values file. Separate test for the value of newrelic-logging.cluster can be added but that's not added till now and we tested it manually anyways.

charts/newrelic-logging/templates/_helpers.tpl
rohit-bandlamudi-nr: > Where does it get deployed if we don't provide cluster name? 

Value of cluster name is only a label we could use later for entity synthesis and relationship synthesis. Doesn't affect location of deployment. If that is empty, it would not be easy to query the fluent bit entity based on the empty cluster name

charts/newrelic-logging/tests/fluentbit_monitoring_config_test.yaml
maya-jha: Thats not a good idea. We should test based on `newrelic-logging.cluster` and then also add a test for `newrelic-logging.cluster`. Otherwise, we will need to add a test here based on both local `cluster` and `global.cluster`

charts/newrelic-logging/templates/_helpers.tpl
maya-jha: can we get cluster name from the deployed cluster?

charts/newrelic-logging/tests/fluentbit_monitoring_config_test.yaml
voorepreethi: > then also add a test for newrelic-logging.cluster

we can have a separate test for this to see if newrelic-logging.cluster is properly set (whether global or local one is chosen). But leave the existing test (based on `cluster` from values file) as it is, as we want to check if  cluster_name is properly set based on the value of "cluster" from values file

rohit-bandlamudi-nr: @voorepreethi see if its possible to test with newrelic-logging.cluster variable.

charts/newrelic-logging/templates/_helpers.tpl
rohit-bandlamudi-nr: AFAIK, it's not possible. We have to check. But the idea here is to not restrict the cluster to take the value of the cluster name that the release is deployed on. The user can give any value he wants here. This depends on what is the name of the cluster entity that the user want to create / associate related entities to.

maya-jha: create a JIRA ticket to address other comments related to test and using cluster name even if its empty.

maya-jha (approved):

hrai-nr (approved):

rohit-bandlamudi-nr (approved):

PR #1647, opened 2025-03-20
[nr-k8s-otel-collector] Add DaemonSet and Deployment container port configuration options

bpschmitt:
#### Is this a new chart

No

Adds the ability for users to specify what container ports need to be open for both the DaemonSet and Deployment.  For example, if a user wants to collect OTel internal telemetry, port 8888 needs to be open.

N/A

#### Checklist
- [X] Chart Version bumped

### DIFF
--- a/charts/nr-k8s-otel-collector/values.yaml
+++ b/charts/nr-k8s-otel-collector/values.yaml
@@ -64,7 +64,21 @@ daemonset:
   # -- Sets additional environment variable sources for the daemonset.
   envsFrom: []
   # -- Settings for daemonset configmap
-  # @default -- See `values.yaml`
+  # @default -- See `values.yaml`\
+  ports: 
+    metrics:
+      enabled: true
+      containerPort: 8888
+      servicePort: 8888
+      protocol: TCP 
+    http:
+      enabled: true
+      containerPort: 4318
+      protocol: TCP
+    grpc:
+      enabled: true
+      containerPort: 4317
+      protocol: TCP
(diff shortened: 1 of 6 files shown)


charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -64,7 +64,21 @@ daemonset:
   # -- Sets additional environment variable sources for the daemonset.
   envsFrom: []
   # -- Settings for daemonset configmap
-  # @default -- See `values.yaml`
+  # @default -- See `values.yaml`\
+  ports:
+    metrics:
+      enabled: true
+      containerPort: 8888
+      servicePort: 8888
```
dbudziwojskiNR, line 72: `servicePort` is being ignored by the tpl function. Do we need to support this parameter?

bpschmitt: @dbudziwojskiNR we should probably adopt what the OTel Collector chart is doing.  

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/38fb9903a04c1578f3ec498708e22dcf29510926/charts/opentelemetry-collector/values.yaml#L306-L353

The `servicePort` just gets added to the associated service for the OTel Collector from what I can tell.

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/38fb9903a04c1578f3ec498708e22dcf29510926/charts/opentelemetry-collector/templates/_config.tpl#L305-L322

PR #1649, opened 2025-03-21
[nr-k8s-otel-collector] OTEL improvements

dbudziwojskiNR:
  - fixes #

#### Checklist
- [x] Chart Version bumped


charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -227,7 +227,7 @@ data:
 
       metricstransform/kubeletstats:
         transforms:
-          - include: container\.(cpu\.utilization|cpu\.usage|filesystem\.(capacity|usage)|memory\.usage)
+          - include: container\.(cpu\.usage|filesystem\.(capacity|usage)|memory\.usage)
```
TmNguyen12, line 230: Just to make sure I am understanding this, why are we marking low.data.mode true if these metrics exist? do they only show up in lowdata mode?

dbudziwojskiNR: It's a little bit unintuitive (and I'd like to change it before GA via the extensibility work we are doing this quarter). The rundown is that we stamp all metrics as `low.data.mode` as false and then set the ones we want to keep to "true". We then drop all metrics that are `low.data.mode=false` thus preserving the metrics we want.

TmNguyen12 (approved):
lgtm

ramkrishankumarN (approved):
LGTM

PR #1670, opened 2025-04-03
[agent-control] multiple improvements of delete job

paologallinaharbur:
#### Is this a new chart
no
improves the way the delete jo is created and managed. IT remove the dependency on the main service account

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent-control/charts/agent-control-deployment/templates/rbac.yaml
+++ b/charts/agent-control/charts/agent-control-deployment/templates/rbac.yaml
@@ -30,19 +30,6 @@ rules:
       - get
       - list
       - watch
-  - apiGroups: [ "apps" ]
-    resources:
-      - deployments/scale
-    verbs:
-      - get
-      - patch
-      - update
-  - apiGroups: [ "" ]
-    resources: ["secrets"]
-    verbs:
-      - delete
-    resourceNames:
-    - {{ include "newrelic-agent-control.auth.secret.name" . }}
   ###
   # AC is capable of manage following resources defined as part of an Agent Type.
   - apiGroups:
--- a/charts/agent-control/charts/agent-control-deployment/templates/uninstall-job.yaml
+++ b/charts/agent-control/charts/agent-control-deployment/templates/uninstall-job.yaml
@@ -8,6 +8,7 @@ Notice that instrumentations needs to go first since it depends on a chart insta
 */ -}}
 {{- $acCRList := (list "instrumentations.newrelic.com" "helmreleases.helm.toolkit.fluxcd.io" "helmrepositories.source.toolkit.fluxcd.io") -}}
 {{- $acResourcesLabelSelector := "app.kubernetes.io/managed-by=newrelic-agent-control" -}}
+{{- $authJobResourcesLabelSelector := "app.kubernetes.io/created-by=ac-auth-job" -}}
 
 apiVersion: batch/v1
 kind: Job
@@ -58,7 +59,7 @@ spec:
       
               {{- if include "newrelic-agent-control.auth.secret.shouldRunJob" . }}
               # Delete the secrets created in the cluster
-              kubectl delete secrets -n {{ .Release.Namespace }} -l {{ $acResourcesLabelSelector }}
+              kubectl delete secrets -n {{ .Release.Namespace }} -l {{ $authJobResourcesLabelSelector }}
               {{- end }}
 
 {{ if .Values.rbac.create }}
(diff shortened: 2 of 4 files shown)


charts/agent-control/charts/agent-control-deployment/templates/rbac.yaml
```diff
@@ -30,19 +30,6 @@ rules:
       - get
       - list
       - watch
-  - apiGroups: [ "apps" ]
-    resources:
-      - deployments/scale
-    verbs:
-      - get
```
gsanchezgavier, line 37: isn't it used by the health checkers?

charts/agent-control/charts/agent-control-deployment/templates/rbac.yaml
```diff
@@ -30,19 +30,6 @@ rules:
...
-      - patch
-      - update
-  - apiGroups: [ "" ]
-    resources: ["secrets"]
-    verbs:
-      - delete
```
gsanchezgavier, line 43: same but for GC?

charts/agent-control/charts/agent-control-deployment/templates/uninstall-job.yaml
```diff
@@ -45,18 +46,105 @@ spec:
...
               if kubectl api-resources -o name |grep {{ $cr }}; then
-                kubectl -n {{ $.Release.Namespace }} delete {{ $cr }} -l {{ $saResourcesLabelSelector }}
+                kubectl -n {{ $.Release.Namespace }} delete {{ $cr }} -l {{ $acResourcesLabelSelector }}
               fi
               {{ end }}
       
               {{- if include "newrelic-agent-control.auth.secret.shouldRunJob" . }}
               # Delete the secrets created in the cluster
-              kubectl delete secrets -n {{ .Release.Namespace }} -l "nr-generic-secret=true"
+              kubectl delete secrets -n {{ .Release.Namespace }} -l {{ $acResourcesLabelSelector }}
```
gsanchezgavier, line 61: isn't line 50 already removing all secretes without any condition ?

charts/agent-control/charts/agent-control-deployment/templates/uninstall-job.yaml
```diff
@@ -1,29 +1,30 @@
...
+  name: {{ $uninstallName }}
   namespace: {{ .Release.Namespace }}
 spec:
   ttlSecondsAfterFinished: 120
   backoffLimit: 3
   template:
     spec:
       restartPolicy: Never
-      serviceAccountName: {{ include "newrelic.common.serviceAccount.name" . }}
+      serviceAccountName: {{ $uninstallServiceAccountName }}
```
gsanchezgavier, line 27: for a moment i thought this was going to cause issues in the future when we add new resources but we still need to manually add the removal of those in this script until we apply the ownership mechanism. So is good to have separate rbac 👍

charts/agent-control/charts/agent-control-deployment/templates/rbac.yaml
paologallinaharbur: a few lines above there is
```
  - apiGroups: [ "apps" ]
    resources:
      - deployments
      - daemonsets
      - statefulsets
    verbs:
      - get
      - list
      - watch
```

That is only to do a scale down of the ac POD when running the delete

charts/agent-control/charts/agent-control-deployment/templates/rbac.yaml
paologallinaharbur: this was for the `{{ include "newrelic-agent-control.auth.secret.name" . }}` only

We have a few lines below the new condition:
```
  - apiGroups: [ "" ]
    resources: ["secrets"]
    verbs:
      - get
      - list
      - watch
      - create
      - delete
      - deletecollection
      - patch
      - update
```

charts/agent-control/charts/agent-control-deployment/templates/uninstall-job.yaml
paologallinaharbur: Yes..
Your question makes me wonder if the CG now tried to delete the L1 secret 🤔

paologallinaharbur: no, but it causes a warning 😕 
```
sub_thread{thread: "Garbage collector"}: executing garbage collection: garbage collector fetched resources without required labels
```

I think it is better to use two different labels

gsanchezgavier (approved):
LGTM

PR #1682, opened 2025-04-16
[nr-k8s-otel-collector] feat: More OTEL Ingest Improvements

dbudziwojskiNR:
  - fixes #

#### Checklist
- [ ] Chart Version bumped


ramkrishankumarN (approved):

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -486,6 +493,7 @@ data:
               - delete_key(attributes, "id")
               - delete_key(attributes, "name")
               - delete_key(attributes, "interface")
+              - delete_key(attributes, "cpu")
```
ramkrishankumarN, line 496: Nice!

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -495,6 +503,10 @@ data:
             action: delete
           - key: net.host.port
             action: delete
+          - key: url.scheme
+            action: delete
+          - key: server.address
+            action: delete
```
ramkrishankumarN, line 509: We are removing these two attributes from every metric right?

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -68,6 +68,13 @@ data:
 
     processors:
       groupbyattrs:
+        keys:
+          - k8s.namespace.name
+          - k8s.node.name
+          - low.data.mode
+          - job_label
+
+      groupbyattrs/ksm:
```
ramkrishankumarN, line 77: Should we get rid of `groupbyattrs/ksm` since it's not used anymore.

ramkrishankumarN: Also, I don't think grouping by such large number of attributes will do any good?

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -606,9 +618,7 @@ data:
       cumulativetodelta:
 
       batch:
-        send_batch_max_size: 1000
         timeout: 30s
```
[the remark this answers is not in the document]
dbudziwojskiNR: Yep. Did we have some other parameters in mind? I do think the larger `send_batch_size` is helpful. But perhaps we should set `send_batch_max_size`

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
dbudziwojskiNR: Agreed. Will do.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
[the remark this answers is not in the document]
ramkrishankumarN: Yeah, we need to set these values lower. 
send_batch_max_size (default = 0): The upper limit of the batch size. 0 means no upper limit of the batch size. 

I remember grpc has around 4MB limit for acceptance. I suggest to keep the   send_batch_max_size to around 3.2MB (80%)  and since send_batch_max_size must be greater than or equal to send_batch_size, keep the send_batch_size to lower e.g. 1MB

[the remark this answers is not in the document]
ramkrishankumarN: With the current settings, it is possible, the data will be dropped if the total batch size, i.e. if 8192 metrics had a size of 8MB, they would be dropped ( exceeded the grpc limit.)

PR #1695, opened 2025-04-29
[nr-k8s-otel-collector] Adding NR Event Attribute identifier to event logs

Philip-R-Beckwith:
#### Is this a new chart
No.

We're using this attribute to identify which logs are actually representative of events. 

And just like that we're already seeing infra events instead of logs. 
Screenshot 2025-04-29 at 1 51 32 PM

Also had to change more '|-' to '|' in the unit tests. My theory is that the unit tests aren't running unless a change that would impact them has been made. So then they do run and the |- makes them fail. Which is why I keep having to remove one or two - from unit tests every time I PR. 

  - fixes #

#### Checklist
- [x] Chart Version bumped


kpattaswamy (approved):

maxa101:
let's also update the ReadMe to point to the new Otlp namespace https://github.com/newrelic/helm-charts/tree/master/charts/nr-k8s-otel-collector OtlpInfrastructureEvent

PR #1703, opened 2025-05-07
[nr-k8s-otel-collector] Add custom config support

kpattaswamy:
#### Is this a new chart
No
TSIA
  - fixes #

#### Checklist
- [ ] Chart Version bumped


charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -91,12 +86,1192 @@ deployment:
...
   envsFrom: []
-  # -- Settings for deployment configmap
-  # @default -- See `values.yaml`
-  configMap:
-    # -- OpenTelemetry config for the deployment. If set, overrides default config and disables configuration parameters for the deployment.
-    config: {}
 
+otel:
+  receivers:
+    deployment:
```
Philip-R-Beckwith, line 92: Not a blocker, 
Looks like we're separating configs into deployment, daemonset and shared. 
When I looked into it, it looks like otel discards configs that aren't applied to a pipeline. 
So we could remove the distinction between configs at this level. 
Would make this file a little easier to read and remove some exists checks in our templating.

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -7,8 +7,28 @@ metadata:
     {{- include "newrelic.common.labels" . | nindent 4 }}
 data:
   deployment-config.yaml: |
-    {{- with include "nrKubernetesOtel.deployment.configMap.config" . }}
-    {{- . | nindent 4 }}
+    {{- if .Values.otel }}
+    receivers:
+      {{- include "deployment-receivers" . | nindent 6 }}
```
Philip-R-Beckwith, line 12: If we're sticking with the deployment, daemonset, shared paradigm, all sections should include a way to include shared components.

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -7,8 +7,28 @@ metadata:
     {{- include "newrelic.common.labels" . | nindent 4 }}
 data:
   deployment-config.yaml: |
-    {{- with include "nrKubernetesOtel.deployment.configMap.config" . }}
-    {{- . | nindent 4 }}
+    {{- if .Values.otel }}
```
Philip-R-Beckwith, line 10: Not a blocker: 
This huge if/else might be better suited in a tpl file.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -7,16 +7,36 @@ metadata:
...
+    connectors:
+      {{- include "daemonset-connector" . | nindent 6 }}
+
+    service:
+      {{- if include "newrelic.common.verboseLog" . }}
+      telemetry:
+        logs:
+          level: "debug"
+      {{- end }}
+      pipelines:
```
Philip-R-Beckwith, line 30: Blocker.
It looks like we're not grabbing the "Service" configs from the yaml.
These sections ^ define all of the components but if we don't reference them in the service.pipelines these components won't do anything.

Philip-R-Beckwith (changes requested):
Looking pretty good so far. 👍

charts/nr-k8s-otel-collector/values.yaml
kpattaswamy: My preference would be to separate by deployment/daemonset/shared since templating logic in `_collector_config.tpl `could target things easier, currently this is what it does for certain configs under each type

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -7,8 +7,28 @@ metadata:
...
+    connectors:
+      {{- include "deployment-connector" . | nindent 6 }}
+
+    service:
+      {{- if include "newrelic.common.verboseLog" . }}
+      telemetry:
+        logs:
+          level: "debug"
+      {{- end }}
+      pipelines:
```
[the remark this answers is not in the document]
kpattaswamy: Extensions and pipelines configs should be exposed in `Values.yaml` which would be piped here now - wdyt?

charts/nr-k8s-otel-collector/values.yaml
dbudziwojskiNR, line 1: I know that we had a discussion earlier about storing the default config in the values file vs in the ConfigMap and I think that we should have another discussion about it. The recent work on the prometheus agent has resurfaced the issues that using the `values.yaml` has.

PR #1715, opened 2025-05-19
[agent-control-deployment] secret ownership

paologallinaharbur:
#### Is this a new chart
no
Changes the secret ownership so that it gets deleted by the agent-control-cli
Adding the agent-id to specify which component is leveraging leveraging the secret and to avoid the garbage collector to complain

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent-control-deployment/templates/preinstall-job-register-system-identity.yaml
+++ b/charts/agent-control-deployment/templates/preinstall-job-register-system-identity.yaml
@@ -290,7 +290,7 @@ spec:
                 {{ include "newrelic-agent-control.auth.secret.name" . }} \
                 --from-literal=CLIENT_ID=$CLIENT_ID \
                 --from-file="private_key=$TEMPORAL_FOLDER/key" | \
-              jq '.metadata.labels += ({{ include "newrelic.common.labels" . | fromYaml | toJson }} + {"app.kubernetes.io/created-by": "ac-auth-job"})' | \
+              jq '.metadata.labels += ({{ include "newrelic.common.labels" . | fromYaml | toJson }} + {"app.kubernetes.io/managed-by": "newrelic-agent-control"} + {"newrelic.io/agent-id": "agent-control"} )' | \
               kubectl apply -n "{{ .Release.Namespace }}" -f -
 ---
 {{ if .Values.rbac.create }}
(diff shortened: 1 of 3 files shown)


charts/agent-control-deployment/templates/preinstall-job-register-system-identity.yaml
```diff
@@ -290,7 +290,7 @@ spec:
                 {{ include "newrelic-agent-control.auth.secret.name" . }} \
                 --from-literal=CLIENT_ID=$CLIENT_ID \
                 --from-file="private_key=$TEMPORAL_FOLDER/key" | \
-              jq '.metadata.labels += ({{ include "newrelic.common.labels" . | fromYaml | toJson }} + {"app.kubernetes.io/created-by": "ac-auth-job"})' | \
+              jq '.metadata.labels += ({{ include "newrelic.common.labels" . | fromYaml | toJson }} + {"app.kubernetes.io/managed-by": "newrelic-agent-control"} + {"newrelic.io/agent-id": "agent-control"} )' | \
```
danielorihuela, line 293: Context: The remote update can be enabled and disabled. At least, https://new-relic.atlassian.net/jira/software/c/projects/NR/boards/61?selectedIssue=NR-383640 had this requirement.

Question: If we merge this change, people not using remote update will have this secret forever?

paologallinaharbur: no the uninstall job is removing it via bash

paologallinaharbur: I mean.. should we keep this uninstall job for the no-remote update use-case?

paologallinaharbur: if we always use the uninstall job it will be deleted anyway

danielorihuela (approved):
I understand the idea is to always use the "new" uninstallation job. Regardless of the installation method. I don't see any problem then.

I might be lacking context thought. Another pair of eyes from @sigilioso I think would be beneficial.

charts/agent-control-deployment/templates/preinstall-job-register-system-identity.yaml
paologallinaharbur: yes exactly!

sigilioso:
I agree with the approach of always using the new uninstallation job, but then we shouldn't merge this until the job is actually in place, right? 🤔

PR #1717, opened 2025-05-19
[agent-control] agent control deployment installation job

danielorihuela:
#### Is this a new chart

no

Uses the new agent control cli to install the `agent-control-deployment` chart. Part of #NR-383640.

#### Checklist
- [x] Chart Version bumped


charts/agent-control/templates/install.yaml
```diff
@@ -0,0 +1,40 @@
...
+  namespace: {{ .Release.Namespace }}
+spec:
+  template:
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" .Release.Name "suffix" "install-job") }}
+      containers:
+        - name: install-agent-control-deployment
+          # TODO: Pin a version to avoid problems with breaking changes.
+          image: newrelic/newrelic-agent-control-cli
```
danielorihuela, line 20: I think we should pin the cli version and update it manually. If not, an update to the cli could break the code of existing clients. However, we don't have any published version yet.

charts/agent-control/values.yaml
```diff
@@ -6,8 +6,9 @@ fullnameOverride: ""
 # -- Values for the agent-control-deployment chart. Ref.: https://github.com/newrelic/helm-charts/blob/master/charts/agent-control-deployment/values.yaml
 # @default -- See `values.yaml`
 agent-control-deployment:
-  # -- Enable the installation of the Agent Control.
-  enabled: true
+  # -- Disables remote update for Agent Control.
+  # When enabled, Agent Control won't be able to update itself from the remote configuration.
+  disableRemoteUpdate: false
```
sigilioso, line 11: This should be enabled by default, shouldn't it? (At very least, while it is under active development!) I would also add a comment specifying that setting it to false is not supported yet.

sigilioso: I really like the approach of using `disableRemoteUpdate` to manage the `agent-control-deployment` chart installation because it is simple and easy to understand. However, I'm a bit worried about some potential drawbacks:

1.- Currently, the flag `agent-conrol-deployment.enabled=false` is used in agent's control test to install flux (checkout the corresponding [Tiltfile](https://github.com/newrelic/newrelic-agent-control/blob/10c1f3b939816c9c35de0998df9cad1623cf832f/agent-control/tests/k8s/Tiltfile#L39-L53). We will need to find out to install the same version of flux we are recommending in the chart. I'd create a branch in the agent-control repository pointing to this one in order to double check that everything is working as expected.

2.- The remote update features is under active development and, for now, it shouldn't be enabled unless you are actively testing/development such feature. I'd say the flag is currently too exposed. We could add some documentation, explicitly name the flag as experimental (Eg: `agent-control-deployment.experimental.disableRemoteUpdate`) or even add a second temporarily flag to check that this is disabled on purpose. Something similar to the snippet bellow:

```
{{- if or  (not (hasKey .Values.agent-control-deployment "disableRemoteUpdate")) (eq .Values.agent-control-deployment.disableRemoteUpdate false) }}
  {{- if not .Values.agent-control-deployment.experimentalRemoteUpdateEnabled -}}
    {{- fail "Remote Update is an experimental feature under active development that should not be enabled unless involved in its development. Enable it at your own risk through .Values.agent-control-deployment.experimentalRemoteUpdateEnabled  " -}}
  {{- end -}}
```

We could also be less _dramatic_ and use a warning instead. Anyway, I'll try to set very explicitly how experimental this feature is right now.

sigilioso (commented):
Great job so far!  I've left some comments

charts/agent-control/README.md
```diff
@@ -72,7 +72,7 @@ agent-control-deployment:
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | agent-control-deployment | object | See `values.yaml` | Values for the agent-control-deployment chart. Ref.: https://github.com/newrelic/helm-charts/blob/master/charts/agent-control-deployment/values.yaml |
-| agent-control-deployment.enabled | bool | `true` | Enable the installation of the Agent Control. |
+| agent-control-deployment.disableRemoteUpdate | bool | `false` | Disables remote update for Agent Control. When enabled, Agent Control won't be able to update itself from the remote configuration. |
```
rubenruizdegauna, line 75: IMHO there are many negations regarding this flag. It's called `disable`, it defaults to `false`, `when enabled, Agent Control won't be able...`

Also in conditions below `if (not (...disableRemoteUpdate)` 

Wouldn't be easier to convert it to positive? `remoteUpdateEnabled:true`

charts/agent-control/templates/rbac.yaml
```diff
@@ -0,0 +1,44 @@
+{{- if (not (index .Values "agent-control-deployment").disableRemoteUpdate) -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  labels:
+    {{- include "newrelic.common.labels" . | nindent 4 }}
+  name: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "install-job-resources") }}
```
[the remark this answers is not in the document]
danielorihuela: I wanted to create a service account with the least amount of privileges.  I don't know if we are more secure using two service accounts for the installation and uninstallation job. I assume that an attacker that could gain access the the "installation service account" will also have access to the "uninstallation service account", but I created it that way to leave it open for whoever implements the uninstall job.

charts/agent-control/templates/ac-deployment-secret.yaml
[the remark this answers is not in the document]
danielorihuela: is there any formatting tool for those golang templates?

charts/agent-control/templates/install.yaml
```diff
@@ -0,0 +1,40 @@
+{{- if (not (index .Values "agent-control-deployment").disableRemoteUpdate) -}}
+{{- $installJobName := include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "install-job" ) -}}
+{{- $releaseName := printf "%s-deployment" .Release.Name -}}
+apiVersion: batch/v1
+kind: Job
```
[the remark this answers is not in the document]
danielorihuela: Didn't thought about that. Is it worth it to implement the retry behaviour in the cli? I would say we can rely on k8s and avoid the extra implementation, testing and maintenance cost. I assume we won't use this cli outside the installation job, that's why I'm saying that.

About the limit... The cli can fail if it can't download the chart or if the installation check fails. I don't think we should use a high number. The installation check by default takes 5 minutes at most. Maybe a `backoffLimit` of 2 or 3 should be enough. If we decrease the "installation check default time" we could increase the backoff limit.

Wdyt? Should we let k8s handle the retry and decrease it to 2 or 3? Just as a starting point, and we iterate over it as we see fit?

charts/agent-control/templates/install.yaml
```diff
@@ -0,0 +1,40 @@
...
+            requests:
+              cpu: 50m
+              memory: 64Mi
+          command:
+            - newrelic-agent-control-cli
+            - install-agent-control
+            - --release-name
+            - "{{ $releaseName }}"
+            - --chart-version
+            - "*"
```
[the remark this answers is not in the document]
danielorihuela: Makes sense. Is it necessary though? Won't be update the chart version in the installation job every time we update the `agent-control-deployment` chart? 

I understand we want to avoid situations in which we update the "child" chart and break the "parent" chart. Should we instead make this field configurable from the values? And allow a range of versions? That would avoid having to bump up the "parent" version every time we increase the "child" one. I don't know if this is desirable or worth it. Asking out of curiosity.

charts/agent-control/README.md
danielorihuela: Yes, this is super shitty. I'll with thought about it again. The problem is that I couldn't find a way to negate the value in the dependencies list.

https://helm.sh/docs/topics/charts/#tags-and-condition-fields-in-dependencies

> Condition - The condition field holds one or more YAML paths (delimited by commas). If this path exists in the top parent's values and resolves to a boolean value, the chart will be enabled or disabled based on that boolean value. Only the first valid path found in the list is evaluated and if no paths exist then the condition has no effect.

```
dependencies:
  - name: flux2
    version: 2.15.0
    condition: flux2.enabled
  - name: agent-control-deployment
    repository: https://helm-charts.newrelic.com
    version: 0.0.46
    condition: agent-control-deployment.disableRemoteUpdate <- Can't negate it
  - name: common-library
    version: 1.3.1
    repository: https://helm-charts.newrelic.com
```

charts/agent-control/templates/ac-deployment-secret.yaml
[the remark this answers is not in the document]
sigilioso: When I read [this post](https://stackoverflow.com/a/729795) I set up my editors to always include the new line at the end of files (regardless of the file type). Example for vscode: `"files.insertFinalNewline": true,`.

Regarding formatting, also with vscode, the Kubernetes extension adds some syntax and _Intellisense_ for helm templates.

charts/agent-control/templates/install.yaml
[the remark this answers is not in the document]
sigilioso: Discussed _offline_: we would be honoring the value defined in the dependency definition (Chart.yaml) that way a user could always choose a different version from the latest.

Regarding a range of version, this seems a feature for the remote-upgrade itself (specifically for choosing the version to send as part of the remote configuration). But the agent-control chart could set ranges (Eg: *, 1, ...), but for now we are setting the version explicitly.

charts/agent-control/templates/install.yaml
```diff
@@ -0,0 +1,40 @@
...
+            - --secrets
+            - "{{ $releaseName }}=agent-control-deployment.yaml,{{ $releaseName }}=global.yaml"
```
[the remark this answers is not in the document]
danielorihuela: done https://github.com/newrelic/helm-charts/pull/1717/commits/fc4c83a3508fc41be04a9f5303f146b5b4846c89

charts/agent-control/templates/install.yaml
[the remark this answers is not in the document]
danielorihuela: done in https://github.com/newrelic/helm-charts/pull/1717/commits/fe8613f945fc8698fe9a488ead859bdcfed7cad8

charts/agent-control/README.md
rubenruizdegauna: Oh, sad to see that limitation, thanks for the explanation!

I see that there is an offline discussion, let's follow it up there.

charts/agent-control/templates/install.yaml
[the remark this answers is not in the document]
danielorihuela: As suggested by @sigilioso. I tried to get the version from the dependency list. It can be done, but only from enabled dependencies. In our case, since enabling remote update disables the dependency list, we can't get it that way.

```
- '{{ (index .Chart.Dependencies 1).Version }}'
```

charts/agent-control/values.yaml
danielorihuela: I created a branch where I'll perform the testing now and future changed regarding naming. Pasting the link here for myself https://github.com/newrelic/newrelic-agent-control/tree/do/test/k8s-keeps-working

charts/agent-control/README.md
danielorihuela: Discussed in https://newrelic.atlassian.net/wiki/spaces/INST/pages/4377182302/DACI+Remote+update+configuration+for+Agent+Control+chart

charts/agent-control/Chart.yaml
```diff
@@ -13,8 +13,7 @@ dependencies:
   - name: agent-control-deployment
     repository: https://helm-charts.newrelic.com
     version: 0.0.46
-    condition: agent-control-deployment.enabled
-    # The following dependency is needed as sub-dependency of agent-control-deployment
+    condition: agent-control-deployment.disableRemoteUpdate
```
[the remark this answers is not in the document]
danielorihuela: Won't do because implementation changed

charts/agent-control/templates/install.yaml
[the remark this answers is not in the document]
danielorihuela: Solved using "appVersion"

charts/agent-control/templates/rbac.yaml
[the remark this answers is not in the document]
danielorihuela: I will leave this open for whoever takes the uninstallation part (me or someone else). He should dig a bit more to know if it makes sense to separate the permissions or not.

danielorihuela:
After discussing offline. We decided to support a single method of installation. Check out https://newrelic.atlassian.net/wiki/spaces/INST/pages/4377182302/DACI+Remote+update+configuration+for+Agent+Control+chart.

Long story short, this changed the PR in the following ways:

* `agent-control-deployment` removed from the dependency list
* `appVersion` added with the `agent-control-deployment` version
* `enabled` flag recovered
* no more double negations in templates

This also implies that the agent control k8s integration tests will keep working without any modifications. You can check it out https://github.com/newrelic/newrelic-agent-control/commit/5381e009d8498150c3e84b634860a3734ff735b6.

charts/agent-control/templates/install.yaml
danielorihuela: Should a line in https://newrelic.atlassian.net/wiki/spaces/INST/pages/4056809473/version+Release be enough? Under Helm Charts.

```
- [ ] Update installation job image version
```

charts/agent-control/templates/ac-deployment-secret.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if (index .Values "agent-control-deployment").enabled -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "agent-control.release.name" . }}
```
gsanchezgavier, line 5: should be  `agent-control.secret.name` instead, i see that is referenced that way in the cli?

charts/agent-control/templates/_helpers.tpl
```diff
@@ -0,0 +1,7 @@
+{{- define "agent-control.release.name" -}}
+  {{- printf "%s-deployment" .Release.Name -}}
+{{- end -}}
+
+{{- define "agent-control.secret.name" -}}
+  {{- printf "%s-deployment" .Release.Name -}}
+{{- end -}}
```
paologallinaharbur, line 7: is it intended that these two functions do exactly the same? Should we unify them somehow?

charts/agent-control/templates/ac-deployment-secret.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if (index .Values "agent-control-deployment").enabled -}}
```
paologallinaharbur, line 1: same here, why not doing .Values.agent-control-deployment.enabled? if you are scared of nulls you could do the 'a(.b(.c))' trick

charts/agent-control/templates/rbac.yaml
```diff
@@ -0,0 +1,44 @@
+{{- if (index .Values "agent-control-deployment").enabled -}}
```
paologallinaharbur, line 1: same here (and in other parts of the PR)

charts/agent-control/templates/rbac.yaml
```diff
@@ -0,0 +1,44 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
```
paologallinaharbur, line 4: aren't you missing:
```
    helm.sh/hook: post-install,post-upgrade
    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
```

charts/agent-control/templates/rbac.yaml
```diff
@@ -0,0 +1,44 @@
...
+    resources: [ "helmrepositories", "helmreleases" ]
+    verbs:
+      - get
+      - create
+      - list
+      - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
```
paologallinaharbur, line 31: same here:
```
    helm.sh/hook: post-install,post-upgrade
    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
```
you might want to add weight to make sure they are created in order

charts/agent-control/templates/service-account.yaml
paologallinaharbur, line 1: the name of the file should reflect that the sa if for the install-job only.
or we can add all rbac related template in a single file called "rbacs-install"

An other option if having a SA that is alwars there and simply create/remove rbacs when needed

charts/agent-control/templates/ac-deployment-secret.yaml
danielorihuela: You are right. It works because the templates contain the same printf function, but this should be the secret. Thanks!!!

charts/agent-control/templates/install.yaml
```diff
@@ -0,0 +1,41 @@
...
+              memory: 64Mi
+          command:
+            - newrelic-agent-control-cli
+            - install-agent-control
+            - --release-name
+            - '{{ include "agent-control.release.name" . }}'
+            - --chart-version
+            - "{{ .Chart.AppVersion }}"
+            - --secrets
+            - "{{ $secretName }}=agent-control-deployment.yaml,{{ $secretName }}=global.yaml"
```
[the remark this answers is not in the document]
danielorihuela: Various agent types do it that way https://github.com/newrelic/newrelic-agent-control/blob/main/agent-control/agent-type-registry/newrelic/com.newrelic.k8s_agent_operator-0.1.0.yaml#L103. I assumed we wanted the same order in every case.

Should we deviate for agent control?

charts/agent-control/templates/_helpers.tpl
danielorihuela: I did that on purpose. However, names could change if someone has a better idea. Thus, I decided to copy paste instead of reuse.

charts/agent-control/templates/ac-deployment-secret.yaml
danielorihuela: As commented on another thread, the problem is with dashes in the name.

charts/agent-control/templates/rbac.yaml
danielorihuela: As commented on another thread, the problem is with dashes in the name.

charts/agent-control/templates/_helpers.tpl
danielorihuela: If you want I could call the release one from the secret one. It wouldn't add nor remove any value, I think.

charts/agent-control/Chart.yaml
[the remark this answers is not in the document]
danielorihuela: I don't know what the cleaning job is. It fails on update, when it tries to creates CRDs that already exist and are owned by the old namespace.

```
- name: Test install charts
  run: ct install --config .github/ct-install.yaml
- name: Test upgrade charts
  run: ct install --config .github/ct-install.yaml --upgrade
```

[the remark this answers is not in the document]
paologallinaharbur: 🤔 the old CRDs? The flux ones are still installed by the main chart, no change there.

I was referring to this https://github.com/newrelic/helm-charts/blob/c4db1a31c1a5e02cff8419029f3e997223878276/charts/agent-control-deployment/values.yaml#L98

charts/agent-control/templates/_helpers.tpl
paologallinaharbur: no strong opinion regarding having two functions, but I still prefer one or reuse the code as you suggested 😅 your call

charts/agent-control/templates/ac-deployment-secret.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "agent-control.release.name" . }}
+  namespace: {{ .Release.Namespace }}
+type: Opaque
+data:
+  agent-control-deployment.yaml: {{ (index .Values "agent-control-deployment") | toYaml | b64enc }}
```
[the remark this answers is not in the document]
gsanchezgavier: oh that's a tricky one , i think we use the chart name for that variable as standard. Anyway i was just curious i imagine that it didn't work some how, thanks for clarifying !

charts/agent-control/templates/install.yaml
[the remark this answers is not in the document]
gsanchezgavier: nono, you are right, there is even a [comment](https://github.com/newrelic/newrelic-agent-control/blob/main/agent-control/agent-type-registry/newrelic/com.newrelic.prometheus-0.1.0.yaml#L55) confirming it

charts/agent-control/templates/rbac.yaml
danielorihuela: True. I added it to the role, the role binding and the service account. They are all used only for the installation job.

charts/agent-control/Chart.yaml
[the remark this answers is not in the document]
danielorihuela: Yes, I haven't tried what you said. I don't think how this would solve the issue. Let me try to explain it in detail.

1. The pipeline starts
2. It installs the chart (with a random namespace)

```
- name: Test install charts
  run: ct install --config .github/ct-install.yaml
```

3. Chart gets installed in a namespace
4. Namespace gets removed (but helm repository and helm release CRDs already exist

Captura de pantalla 2025-05-22 a las 18 40 38 (1)

5. Pipeline moves to upgrade

```
- name: Test upgrade charts
  run: ct install --config .github/ct-install.yaml --upgrade
```

6. Test fails because CRDs already exist and are owned by another namespace

```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: CustomResourceDefinition "helmreleases.helm.toolkit.fluxcd.io" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "agent-control-zc7mxpp7w6": current value is "agent-control-23izzp36fi"; annotation validation error: key "meta.helm.sh/release-namespace" must equal "agent-control-zc7mxpp7w6": current value is "agent-control-23izzp36fi"
```

---

I'm not sure how `cleanupManagedResources: true` might help.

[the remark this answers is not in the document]
danielorihuela: I tried it just in case. Got the same error.

Captura de pantalla 2025-05-27 a las 17 44 10

charts/agent-control/templates/_helpers.tpl
danielorihuela: I like having two. Just because they represent different things. The secret name and the release name. It's coincidental that the value is equivalent. Let me call one from the other though.

charts/agent-control/templates/service-account.yaml
danielorihuela: I'll leave this for @rubenruizdegauna. He is already working on uninstall and he will decide or at least propose whether we should use one or two service accounts.

paologallinaharbur (dismissed):

charts/agent-control/templates/install.yaml
danielorihuela: You mean for integration tests and all that? Okay.

sigilioso: Not only for integration tests but also for user needed a private registry/repository.

danielorihuela: I was thinking about prefixing with "dev", but if some users might need to use their private repositories, I'll leave it. Why would clients need the private repository?

I don't like that now the user can shoot his foot by putting whatever tag they want in there. When it was hardcoded, we could enforce a correct relation between the cli version and the chart version.

danielorihuela: I will still need to modify https://newrelic.atlassian.net/wiki/spaces/INST/pages/4056809473/version+Release to update the "tag" in the chart values, I assume.

sigilioso: It actually depends on their requirements. I can imagine a security requirement where all images are mirrored into a private repository after some internal and independent security checks. I believe we have already received a work request to support such as use case (in regards of Agent Types where the repository is not configurable for now)

charts/agent-control/README.md
```diff
@@ -72,7 +72,10 @@ agent-control-deployment:
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | agent-control-deployment | object | See `values.yaml` | Values for the agent-control-deployment chart. Ref.: https://github.com/newrelic/helm-charts/blob/master/charts/agent-control-deployment/values.yaml |
+| agent-control-deployment.chartRepositoryUrl | string | `"https://helm-charts.newrelic.com"` | The repository URL from where the `agent-control-deployment` chart will be installed. |
 | agent-control-deployment.enabled | bool | `true` | Enable the installation of the Agent Control. |
+| agent-control-deployment.installerImage | object | `{"pullPolicy":"IfNotPresent","pullSecrets":[],"registry":null,"repository":"newrelic/newrelic-agent-control-cli","tag":"nightly"}` | The image used to install Agent Control. |
```
paologallinaharbur: I think that this name is misleading since we use it as well for the uninstall

paologallinaharbur (dismissed):
I would change the installer image name

charts/agent-control/templates/install.yaml
```diff
@@ -0,0 +1,48 @@
...
+      serviceAccountName: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" .Release.Name "suffix" "install-job") }}
+      containers:
+        - name: install-agent-control-deployment
+          {{- if (index .Values "agent-control-deployment").toolkitImage }}
+          image: {{ include "newrelic.common.images.image" ( dict "imageRoot" (index .Values "agent-control-deployment").toolkitImage "context" .) }}
+          imagePullPolicy: {{ (index .Values "agent-control-deployment").toolkitImage.pullPolicy }}
+          {{- else }}
+          image: "newrelic/newrelic-agent-control-cli:nightly"
+          imagePullPolicy: "IfNotPresent"
+          {{- end }}
```
paologallinaharbur, line 27: I would remove it right after merging this

danielorihuela: okey, let's go

PR #1723, opened 2025-05-26
[agent-control-deployment] feat: setup liveness/readiness probes

sigilioso:
#### Is this a new chart
No

This PR includes readiness/liveness probes for the Agent Control container. Additionally includes checks to assure that the Agent Control status server is enabled with the expected port set.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent-control-deployment/templates/_helpers.tpl
+++ b/charts/agent-control-deployment/templates/_helpers.tpl
@@ -96,7 +96,7 @@ cluster name, licenses, and custom attributes
 {{- define "newrelic-agent-control.config.content" -}}
 
 {{- /* config set here so we can populate it as we enable and disable snippets. */ -}}
-{{- $statusServerPort := int (include "newrelic-agent-control.config.statusServer.port" .) -}}
+{{- $statusServerPort := ((.Values.config).status_server).port -}}
 {{- $statusServerHost := "0.0.0.0" -}}
 {{- $config := dict "server" (dict "enabled" true "port" $statusServerPort "host" $statusServerHost) -}}
 
@@ -140,10 +140,6 @@ cluster name, licenses, and custom attributes
 {{- end -}}
 
 {{- $config | toYaml -}}
-{{- end -}}
-
-{{- define "newrelic-agent-control.config.statusServer.port" -}}
-{{ ((.Values.config).status_server).port | default 51200 }}
 {{- end -}}
 
 {{- /* These are the defaults that are used for all the containers in this chart */ -}}
--- a/charts/agent-control-deployment/templates/deployment-agentcontrol.yaml
+++ b/charts/agent-control-deployment/templates/deployment-agentcontrol.yaml
@@ -119,13 +119,13 @@ spec:
           livenessProbe:
             httpGet:
               path: /status
-              port: {{ include "newrelic-agent-control.config.statusServer.port" . }}
+              port: {{ .Values.config.status_server.port }}
             failureThreshold: 1
             periodSeconds: 10
           startupProbe:
             httpGet:
               path: /status
-              port: {{ include "newrelic-agent-control.config.statusServer.port" . }}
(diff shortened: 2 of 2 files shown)


charts/agent-control-deployment/templates/deployment-agentcontrol.yaml
```diff
@@ -116,6 +116,20 @@ spec:
...
+              port: {{ .Values.config.status_server.port }}
+            failureThreshold: 1
+            periodSeconds: 10
+          startupProbe:
+            httpGet:
+              path: /status
+              port: {{ .Values.config.status_server.port }}
+            failureThreshold: 5
+            periodSeconds: 5
+            initialDelaySeconds: 5
```
sigilioso, line 131: I'm not completely sure about the best periods and thresholds. Should we make them configurable?

charts/agent-control-deployment/templates/deployment-agentcontrol.yaml
```diff
@@ -116,6 +116,20 @@ spec:
 
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+          livenessProbe:
```
paologallinaharbur, line 119: what's about readiness probe? We do not receive traffic, but could it be useful anyway?

paologallinaharbur, line 119: Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(Deployment.spec.template.spec): unknown field "livenessProbe" in io.k8s.api.core.v1.PodSpec, ValidationError(Deployment.spec.template.spec): unknown field "readinessProbe" in io.k8s.api.core.v1.PodSpec]

charts/agent-control-deployment/templates/deployment-agentcontrol.yaml
paologallinaharbur: we can start simply and adjust if needed in the future 😄

charts/agent-control-deployment/templates/deployment-agentcontrol.yaml
sigilioso: Is it an old error? I messed up with indentation at first but I have already updated it 😅 
I'm not sure if you are referring to the same

sigilioso: I started with livenessProbe + readinessProbe but I switched to startupProbe + livenessProbe. I'll double check and share the explanation

charts/agent-control-deployment/templates/_helpers.tpl
```diff
@@ -125,10 +127,24 @@ cluster name, licenses, and custom attributes
...
+{{- if ne (printf "%v" $config.server.port) (printf "%v" $statusServerPort) -}}
+  {{- fail "Setting up the status server port is `.Values.config.agentControl.content` is not supported because it would conflict with container probes. Use `.Values.config.status_server.port` instead" -}}
 {{- end -}}
 
+{{- $config | toYaml -}}
+{{- end -}}
 
+{{- define "newrelic-agent-control.config.statusServer.port" -}}
+{{ ((.Values.config).status_server).port | default 51200 }}
+{{- end -}}
```
paologallinaharbur, line 147: I am a bit confused, the default is already in the values.yaml... why do we need this function?

paologallinaharbur: It should update the defaults values.yaml. 

You can try manually to do the upgrade, but I believe it should work. Maybe it is `ct` misbehaving

paologallinaharbur (approved):

sigilioso:
As pointed out by @paologallinaharbur (thanks!), the update test is failing because `ct` is using `--reuse-values` which takes previous values not including the new defaults added in the PR.

PR #1736, opened 2025-06-04
add nr_ prefix to new relic default components

dpacheconr:
Add nr_ prefix to default new relic components

This decreases the amount times collector does not start when configuring components with the same name when using this new functionality 
https://github.com/newrelic/helm-charts/blob/master/charts/nr-k8s-otel-collector/values.yaml#L75
https://github.com/newrelic/helm-charts/blob/master/charts/nr-k8s-otel-collector/values.yaml#L107

### DIFF
--- a/charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
+++ b/charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
@@ -12,7 +12,7 @@ data:
     {{- else }}
     receivers:
       {{- include "nrKubernetesOtel.daemonset.configMap.extraConfig.receivers" . | nindent 6 }}
-      hostmetrics:
+      hostmetrics/nr:
         # TODO (chris): this is a linux specific configuration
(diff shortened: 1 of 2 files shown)


dpacheconr:
Steps to reproduce
Add custom/extra component to https://github.com/newrelic/helm-charts/blob/master/charts/nr-k8s-otel-collector/values.yaml#L167 that has the same name as one of our existing components, AFAIK we don't do any transformation on those via the helpers (.i.e. merge, concat), so the values will be added as is to configmap
Common example, is to add more prometheus jobs, ours is called prometheus, if you add extraconfig receiver named prometheus
The collector will fail to start


This is because receivers, processors, exporters and pipelines are defined through component identifiers following the type[/name] format, for example otlp or otlp/2. You can define components of a given type more than once as long as the identifiers are unique. Having quite generic names, will cause issues when customers try to add the same component with the same generic name
https://opentelemetry.io/docs/collector/configuration/

Hence changing our components to have unique names with nr prefix

PR #1742, opened 2025-06-09
[nr-k8s-otel-collector] Apply groupbyattr and transform ksm with correct k8sattributes in daemonset

ramkrishankumarN:
1. Fixes a typo for the openshift inside processors.
2. Removes the attributes/self because it is not needed anymore.
3. Uses k8sattribute processor which is being used inside deployment configmap
4. Factors the prometheus into a new pipeline

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
+++ b/charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
@@ -542,89 +542,6 @@ data:
           - key: net.host.port
             action: delete
 
-      attributes/self:
-        actions:
-          - key: k8s.node.name
-            action: upsert
-            from_attribute: node
-          - key: k8s.namespace.name
-            action: upsert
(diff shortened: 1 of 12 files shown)


charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -668,11 +673,37 @@ data:
...
+            - resourcedetection/cloudproviders
+            {{- end }}
+            - resource
+            {{- if include "nrKubernetesOtel.lowDataMode" . }}
+            - transform/low_data_mode_inator
+            - resource/low_data_mode_inator
+            {{- end }}
+            - k8sattributes/ksm
+            - cumulativetodelta
+            - batch
```
ramkrishankumarN, line 688: The idea here is to match "exactly" what we did for KSM in deployments.

PR #1744, opened 2025-06-10
[agent-control] uninstallation job

rubenruizdegauna:
#### Is this a new chart
No
This PR adds the uninstallation job for the AC.
* Uninstallation SA
* RBAC
* Uninstallation job

#### Checklist
- [x] Chart Version bumped


charts/agent-control/values.yaml
```diff
@@ -2,6 +2,8 @@
 nameOverride: ""
 # -- Override the full name of the release
 fullnameOverride: ""
+# -- Override the url of the AC helm repository
+repositoryUrl: ""
```
danielorihuela: Are this and the `chartRepositoryUrl` I created below the same? I think they are.
I think it makes sense to have that option below `agent-control-deployment` since it's used there.

I don't know for sure. Can an expert chime in?

charts/agent-control/templates/install.yaml
```diff
@@ -36,6 +36,8 @@ spec:
             - "{{ $secretName }}=agent-control-deployment.yaml,{{ $secretName }}=global.yaml"
             - --namespace
             - "{{ .Release.Namespace }}"
+            {{- if .Values.repositoryUrl }}
             - --repository-url
-            - '{{ (index .Values "agent-control-deployment").chartRepositoryUrl }}'
+            - "{{ .Values.repositoryUrl }}"
+            {{- end }}
```
danielorihuela: I know the cli has the default value already. However, I think it's cool having it here to have a second layer of defense for free. That way, we won't have to change the chart to make mandatory if we ever change the cli to not have a default.

I might be to picky. Not a strong opinion anyway. We can leave it as it is now.

charts/agent-control/templates/rbac.yaml
```diff
@@ -46,7 +46,69 @@ roleRef:
...
+  - kind: ServiceAccount
+    name: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" .Release.Name "suffix" "install-job") }}
+    namespace: {{ .Release.Namespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  annotations:
+    helm.sh/hook: post-install,post-upgrade
+    helm.sh/hook-weight: "0"
```
danielorihuela: Are we missing `helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded`?

charts/agent-control/templates/install.yaml
```diff
@@ -18,8 +18,13 @@ spec:
       serviceAccountName: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" .Release.Name "suffix" "install-job") }}
       containers:
         - name: install-agent-control-deployment
+          {{- if (index .Values "agent-control-deployment").installerImage }}
           image: {{ include "newrelic.common.images.image" ( dict "imageRoot" (index .Values "agent-control-deployment").installerImage "context" .) }}
           imagePullPolicy: {{ (index .Values "agent-control-deployment").installerImage.pullPolicy }}
+          {{- else }}
+          image: "newrelic/newrelic-agent-control-cli:nightly"
+          imagePullPolicy: "IfNotPresent"
+          {{- end }}
```
danielorihuela, line 27: Workaround for "ct issue". Remove before merging once tests work. Or let it there and remove in a future release.

charts/agent-control/templates/install.yaml
```diff
@@ -36,6 +41,8 @@ spec:
             - "{{ $secretName }}=agent-control-deployment.yaml,{{ $secretName }}=global.yaml"
             - --namespace
             - "{{ .Release.Namespace }}"
+            {{- if (index .Values "agent-control-deployment").chartRepositoryUrl }}
             - --repository-url
-            - '{{ (index .Values "agent-control-deployment").chartRepositoryUrl }}'
+            - "{{ (index .Values "agent-control-deployment").chartRepositoryUrl }}"
+            {{- end -}}
```
danielorihuela, line 47: Workaround for "ct issue". Remove before merging once tests work. Or let it there and remove in a future release.

charts/agent-control/templates/rbac.yaml
```diff
@@ -46,7 +46,63 @@ roleRef:
...
   name: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "install-job-resources") }}
 subjects:
-- kind: ServiceAccount
-  name: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" .Release.Name "suffix" "install-job") }}
+  - kind: ServiceAccount
+    name: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" .Release.Name "suffix" "install-job") }}
+    namespace: {{ .Release.Namespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
danielorihuela, line 55: This role is owned by `Helm`. We don't need hooks. It will be deleted when the release is deleted.

charts/agent-control/templates/rbac.yaml
```diff
@@ -46,7 +47,63 @@ roleRef:
...
+    resources: [ "configmaps","secrets" ]
+    verbs:
+      - delete
+      - list
+      - watch
+      - deletecollection
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
```
paologallinaharbur, line 100: I would follow the same approach as above, in the installation job we have the hooks, no?

charts/agent-control/templates/service-account-uninstallation.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if (index .Values "agent-control-deployment").enabled -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
+    helm.sh/hook: pre-delete
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+    helm.sh/hook-weight: "-20"
```
paologallinaharbur, line 8: IMO if we delete the SA it makes sense to delete relate RBACs as well

charts/agent-control/templates/uninstall.yaml
```diff
@@ -0,0 +1,35 @@
...
+  namespace: {{ .Release.Namespace }}
+spec:
+  backoffLimit: 2
+  template:
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" .Release.Name "suffix" "uninstall-job") }}
+      containers:
+        - name: uninstall-agent-control-deployment
+          image: newrelic/newrelic-agent-control-cli:nightly
```
paologallinaharbur: why is it hardcoded? I think they are templated in the values.yaml

charts/agent-control/templates/install.yaml
```diff
@@ -36,6 +41,8 @@ spec:
             - "{{ $secretName }}=agent-control-deployment.yaml,{{ $secretName }}=global.yaml"
             - --namespace
             - "{{ .Release.Namespace }}"
+            {{- if (index .Values "agent-control-deployment").chartRepositoryUrl }}
             - --repository-url
-            - '{{ (index .Values "agent-control-deployment").chartRepositoryUrl }}'
+            - "{{ (index .Values "agent-control-deployment").chartRepositoryUrl }}"
```
paologallinaharbur, line 46: is this related to the other PR?

charts/agent-control/templates/uninstall.yaml
danielorihuela: Yes. It was hardcoded by @rubenruizdegauna and I didn't change it. Sorry.

charts/agent-control/templates/rbac.yaml
danielorihuela: Yes, but these are used for the uninstallation job. They are automatically removed when we delete the helm release. So... Do we need the hooks at all?

danielorihuela: Ah, I should use the same hooks as in the service account, right?

charts/agent-control/templates/install.yaml
danielorihuela: kind of. This variable is new and comes from the insallation job. We can discuss the name and the variable there.
Regarding the "if" condition. We need this workaround here to avoid the issue with the new variables not existing in the previous version.

charts/agent-control/templates/rbac.yaml
```diff
@@ -29,6 +29,7 @@ rules:
       - create
       - list
       - watch
+      - update
```
paologallinaharbur, line 32: ? why do we need update for the uninstallation?

paologallinaharbur: but I guess it is for the parent PR

charts/agent-control/templates/rbac.yaml
```diff
@@ -46,7 +47,71 @@ roleRef:
...
+  labels:
+    {{- include "newrelic.common.labels" . | nindent 4 }}
+  name: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "uninstall-job-resources") }}
   namespace: {{ .Release.Namespace }}
+rules:
+  - apiGroups: [ "" ]
+    resources: [ "namespaces" ]
+    verbs:
+      - get
+  - apiGroups: [ "apps" ]
```
paologallinaharbur, line 70: do we need this?

paologallinaharbur: I think we are not checking any of that

charts/agent-control/templates/rbac.yaml
danielorihuela: Yes, if I don't add this permissions the tests from `ct` fail. I think it's because of the update test.

charts/agent-control/templates/rbac.yaml
danielorihuela: mmmm, I'll double check

danielorihuela: If I remove it, tests fail

PR #1760, opened 2025-06-22
[agent-control-deployment] feat: use newrelic-auth-cli to manage system id creation

DavSanchez:
  - fixes #

#### Checklist
- [x] Chart Version bumped


charts/agent-control-deployment/templates/preinstall-job-register-system-identity.yaml
```diff
@@ -111,185 +111,65 @@ spec:
...
-                      sleep 2
-                    else
-                      echo -n "Error getting system identity auth token. The API endpoint returned $HTTP_CODE: " && cat "$TEMPORAL_FOLDER/response_token.json" | tr -d '\n' && echo " Retrying ($RETRY/3)..."
-                      sleep 2
-                      continue
-                    fi
-                  fi
-                done
+              ACCESS_TOKEN=""
+              for RETRY in 1 2 3; do
```
danielorihuela, line 120: We need a new flag in the cli I think. A `--retries` flag. The `TokenRetrieverWithCache` has a retries field. Would it simplify this loop here?

danielorihuela: Not a big deal. In fact, we don't have the same behaviour in place for "identity creation". So maybe it's better to let it as it is for consistency.

charts/agent-control/Chart.yaml
```diff
@@ -3,9 +3,9 @@ name: agent-control
 description: Bootstraps New Relic' Agent Control
 
 type: application
-version: 0.0.68
+version: 0.0.69
```
danielorihuela: While working on the uninstallation and installation job I asked @paologallinaharbur to know if we could have a single PR updating both the `agent-control` and `agent-control-deployment` chart versions. I don't remember if he told me that "it's not possible" or that "he was not sure".

Do you have more knowledge on that? Maybe the tests fail because of that.

paologallinaharbur: We can try, there are two things that give me hope:
 - I believe that it was no supported by the tool we were using, but it was 3-4 years ago, therefore maybe now it works.

 - moreover, it is possible that since the way we manage child charts is different (via the job and not direct dependency) it could just work without publishing first the child chart

paologallinaharbur: we had this requirement, but I do not think that it is really needed in our case
> ### Requirements
> * Open separate pull requests to submit changes to multiple charts.

paologallinaharbur: Now for sure it is supported https://github.com/helm/chart-releaser/issues/411

charts/agent-control-deployment/templates/preinstall-job-register-system-identity.yaml
DavSanchez: I also thought on that... I think the ideal state would be to just cover everything in the CLI so we can just replace the whole script with a single call, but we are not there yet.

charts/agent-control/Chart.yaml
danielorihuela: I think it is

> {"level":"error","ts":"2025-06-25T12:39:10.328Z","msg":"reconciliation stalled","controller":"helmchart","controllerGroup":"source.toolkit.fluxcd.io","controllerKind":"HelmChart","HelmChart":{"name":"agent-control-26t5nzkrop-agent-control-deployment","namespace":"agent-control-26t5nzkrop"},"namespace":"agent-control-26t5nzkrop","name":"agent-control-26t5nzkrop-agent-control-deployment","reconcileID":"5f528d93-98ea-4d1b-b0ba-ad90b395442b","error":"invalid chart reference: failed to get chart version for remote reference: no 'agent-control-deployment' chart with version matching '0.0.53' found"}

DavSanchez: So if I'm still finding this on the CI means I need to open separate PRs anyways then? Ok, give me a couple minutes and I'll do!

DavSanchez: Done

danielorihuela: I'm not sure how `ct` works, but I think it's not smart enough to install all the modified charts from the PR instead of the registry. Thus, maybe we can update both, but tests will fail.

paologallinaharbur: if we want we could disable the healthcheck during the installation

PR #1767, opened 2025-06-25
[agent-control] feat: update agent-control-deployment version

DavSanchez:
#### Is this a new chart

No

Updates the `agent-control` chart to use the latest `agent-control-deployment` chart version added in #1760.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent-control/Chart.yaml
+++ b/charts/agent-control/Chart.yaml
@@ -3,9 +3,9 @@ name: agent-control
 description: Bootstraps New Relic' Agent Control
 
 type: application
-version: 0.0.69
+version: 0.0.70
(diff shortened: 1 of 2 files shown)


danielorihuela:
Can we take the opportunity to pin the version of the cli? https://github.com/newrelic/helm-charts/blob/248c731d53ef107c2aebe5eeaf70c78500f91a67/charts/agent-control/values.yaml#L20

danielorihuela:
If we can, it would be great to merge this after https://github.com/newrelic/helm-charts/pull/1773. That way every version is pinned now that the new cli is used.

gsanchezgavier:
updating it in #1776

PR #1770, opened 2025-06-25
[nr-k8s-otel-collector] Fixing naming for rendered files.

Philip-R-Beckwith:
#### Is this a new chart
No

The rendered files were full of names like "example-nr-k8s-otel-collector" 
And tags that included "example".
This removes the "example" from the rendered files giving a more polished chart for those choosing to use the pre-rendered files. 

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ generate-examples:
 			VALUES=$$(find $${EXAMPLES_DIR}/$${example} -name *values.yaml); \
 			rm -rf "$${EXAMPLES_DIR}/$${example}/rendered"; \
 			for value in $${VALUES}; do \
-				helm template example charts/$${chart_name} --namespace newrelic --set licenseKey='<NR_licenseKey>' --set cluster='<cluser_name>' --values $${value} --output-dir "$${EXAMPLES_DIR}/$${example}/rendered"; \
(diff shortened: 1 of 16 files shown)


danielstokes (approved):

Makefile
```diff
@@ -13,7 +13,7 @@ generate-examples:
 			VALUES=$$(find $${EXAMPLES_DIR}/$${example} -name *values.yaml); \
 			rm -rf "$${EXAMPLES_DIR}/$${example}/rendered"; \
 			for value in $${VALUES}; do \
-				helm template example charts/$${chart_name} --namespace newrelic --set licenseKey='<NR_licenseKey>' --set cluster='<cluser_name>' --values $${value} --output-dir "$${EXAMPLES_DIR}/$${example}/rendered"; \
+				helm template "$${chart_name}" charts/$${chart_name} --namespace newrelic --set licenseKey='<NR_licenseKey>' --set cluster='<cluser_name>' --set fullnameOverride="$${chart_name}" --values $${value} --output-dir "$${EXAMPLES_DIR}/$${example}/rendered"; \
```
dbudziwojskiNR: Do we need the full name override? It should just work with `nr-k8s-otel-collector` as the chart name.

Philip-R-Beckwith: Apparently we dont... When I was looking at the logic I was sure we'd end up with names like... 
`nr-k8s-otel-collector-nr-k8s-otel-collector-daemonset` if we didn't add the full name override. 
Not sure why we don't without it TBH. 
Since using example renders `example-nr-k8s-otel-collector-daemonset`... 
But testing it, we do not need it... So I removed it... 🤷

dbudziwojskiNR (approved):

PR #1783, opened 2025-07-02
feat: improvements over split namespaces

DavSanchez:
  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/agent-control-deployment/templates/_helpers.tpl
+++ b/charts/agent-control-deployment/templates/_helpers.tpl
@@ -104,7 +104,7 @@ cluster name, licenses, and custom attributes
 {{- $k8s := (dict
   "cluster_name" (include "newrelic.common.cluster" .)
   "namespace" .Release.Namespace
-  "namespace_agents" ((.Values.config).subAgentsNamespace)
+  "namespace_agents" (default .Release.Namespace .Values.config.subAgentsNamespace)
 ) -}}
 {{- $config = mustMerge $config (dict "k8s" $k8s) -}}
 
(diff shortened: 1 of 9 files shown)


charts/agent-control-deployment/templates/_helpers.tpl
```diff
@@ -101,7 +101,11 @@ cluster name, licenses, and custom attributes
 {{- $config := dict "server" (dict "enabled" true "port" $statusServerPort "host" $statusServerHost) -}}
 
 {{- /* Add to config k8s cluster and namespace config */ -}}
-{{- $k8s := (dict "cluster_name" (include "newrelic.common.cluster" .) "namespace" .Release.Namespace "namespace_agents" .Release.Namespace) -}}
+{{- $k8s := (dict
+  "cluster_name" (include "newrelic.common.cluster" .)
+  "namespace" .Release.Namespace
+  "namespace_agents" (default .Release.Namespace .Values.config.subAgentsNamespace)
```
danielorihuela: I'm aware this is still a draft, but I had a doubt while working on the public docs for this new feature. Which will be the default name for the `namespace_agents` variable?

I see that it will take the value from the `values.yaml` and if not, it will take a default. Will this default be the release namespace? I guess the idea is to have a different one, right? Do we know already which will be it?

DavSanchez: You're right that this is not exactly the final form, this was part of a round of testing that I was doing making sure everything was working as expected. Yes, the idea is that namespace where we install agent-control will be a different one that we indicate when running the `helm` command, and the namespace where the agents are installed will be `newrelic` by default (as is the typical namespace used by all NR agents that might assume it) unless it's changed somehow. CC @paologallinaharbur

DavSanchez:
Unblocking PR with minimal changes, will refine the `ClusterRole`s and `Role`s on a separate one.

DavSanchez:
> I believe this PR is not relevant anymore

You're right, closing in favor of https://github.com/newrelic/helm-charts/pull/1788 and https://github.com/newrelic/helm-charts/pull/1789.

PR #1788, opened 2025-07-07
[agent-control-deployment] feat: split namespaces

DavSanchez:
  - fixes #

#### Checklist
- [x] Chart Version bumped


DavSanchez:
Requesting @paologallinaharbur 's review because this contains some of his work from https://github.com/newrelic/helm-charts/pull/1777

charts/agent-control-deployment/templates/preinstall-job-register-system-identity.yaml
```diff
@@ -67,8 +67,6 @@ spec:
         - name: register-system-identity
           image: "{{ (.Values.systemIdentityRegistration.image).repository }}:{{ (.Values.systemIdentityRegistration.image).tag }}"
           imagePullPolicy: {{ .Values.systemIdentityRegistration.image.pullPolicy }}
-          securityContext:
-            allowPrivilegeEscalation: false
```
danielorihuela: Why was this removed?

I added it because it's a good practice https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-4-prevent-in-container-privilege-escalation. Not against removing it, but I want to understand the reason.

charts/agent-control-deployment/templates/rbac.yaml
```diff
@@ -73,6 +55,56 @@ roleRef:
...
+  labels:
+    {{- include "newrelic.common.labels" . | nindent 4 }}
+  name: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "watch") }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "watch") }}
+subjects:
 - kind: ServiceAccount
   name: {{ include "newrelic.common.serviceAccount.name" . }}
```
danielorihuela, line 91: The service account above includes the same template. Does it give the same name or a different one? If it's the same name, isn't that an issue?

charts/agent-control-deployment/templates/preinstall-job-register-system-identity.yaml
DavSanchez: Hmm, might have been an accidental removal due to rebase. I'll recover it!

DavSanchez: Done in https://github.com/newrelic/helm-charts/pull/1788/commits/16ce24f8c9b2bce99cddf4632eb819127a56819c

charts/agent-control-deployment/templates/rbac.yaml
DavSanchez: The service accounts specified here are the "destinations" of the role bindings, this means that the same account will get the roles specified in both rolebindings. As the roles are additive (i.e. no "deny" roles), I think the roles can overlap without issues. CC @paologallinaharbur for confirmation.

Does this answer your question or was it some other concern?

paologallinaharbur: you can have multiple roleBindings binding different roles to the same service account!

charts/agent-control-deployment/templates/rbac.yaml
```diff
@@ -73,6 +55,56 @@ roleRef:
...
+  - apiGroups: [ "newrelic.com" ]
+    resources: ["*"]
+    verbs: ["*"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  labels:
+    {{- include "newrelic.common.labels" . | nindent 4 }}
+  name: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "watch") }}
```
paologallinaharbur: I think this is not corrext, after the reflector change, AC does not need anymore to list and watch resources across the whole cluster, but only in 2 namespaces

paologallinaharbur: in this way you are allowing AC to get secrets across the whole cluster

charts/agent-control-deployment/Chart.yaml
```diff
@@ -4,8 +4,8 @@ description: A Helm chart to install New Relic Agent Control on Kubernetes
 
 type: application
```
paologallinaharbur, line 5: you need to specify the namespace also in the values.yaml of this chart:
```
templates/: template: agent-control-deployment/templates/namespace-agents.yaml:2:39: executing "agent-control-deployment/templates/namespace-agents.yaml" at <$subAgentsNamespace>: invalid value; expected string
```

charts/agent-control-deployment/templates/namespace-agents.yaml
```diff
@@ -0,0 +1,9 @@
+{{- $subAgentsNamespace := index (include "newrelic-agent-control.config.content" . | fromYaml) "k8s" "namespace_agents" -}}
```
paologallinaharbur: I am not following why you are not simply getting it from .Values.subAgentsNamespace

paologallinaharbur (commented):
RBACs looks ok now! I think we need to change the way the namespace is fetched when creating it. Moreover, it should be present in the values.yaml with a default

charts/agent-control-deployment/templates/rbac.yaml
danielorihuela: TIL. I didn't know that.

I was asking because of "security" reasons. But I was thinking now that we don't want two separate service accounts because it would be cumbersome to properly handle everything right?

PR #1792, opened 2025-07-08
[agent-control-deployment] cleaning unused code and features

paologallinaharbur:
#### Is this a new chart
no
It removes:
 - the possibility to configure the key of the secret where some config are fetched from. 
    This feature it adds complexity without a real benefit.
 - it is no longer templating the secret for userKey and mounting it as an environment variables
 -  related code functions and tests

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent-control-deployment/templates/deployment-agentcontrol.yaml
+++ b/charts/agent-control-deployment/templates/deployment-agentcontrol.yaml
@@ -69,7 +69,7 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: {{ include "newrelic-agent-control.auth.secret.name" . }}
-                  key: {{ include "newrelic-agent-control.auth.secret.clientId.key" . }}
+                  key: CLIENT_ID
           {{- end }}
           {{- if include "newrelic.common.verboseLog" . }}
             - name: NR_AC_LOG__LEVEL
@@ -150,7 +150,7 @@ spec:
           secret:
             secretName: {{ include "newrelic-agent-control.auth.secret.name" . }}
             items:
-              - key: {{ include "newrelic-agent-control.auth.secret.privateKey.key" . }}
+              - key: private_key
                 path: from-secret.key
         {{- end }}
         {{- with .Values.extraVolumes }}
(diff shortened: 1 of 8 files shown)


charts/agent-control-deployment/templates/deployment-agentcontrol.yaml
```diff
@@ -69,7 +69,7 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: {{ include "newrelic-agent-control.auth.secret.name" . }}
-                  key: {{ include "newrelic-agent-control.auth.secret.clientId.key" . }}
+                  key: CLIENT_ID
```
danielorihuela, line 72: Before we could specify the key inside the configmap where the client id was, right?
This is something internal to AC. We don't need to make it configurable. That's the point, am I right?

Just for my understanding.

charts/agent-control-deployment/templates/preinstall-job-register-system-identity.yaml
```diff
@@ -43,8 +26,8 @@ metadata:
   name: {{ include "newrelic-agent-control.auth.generatedIdentityCredentialsSecretName" . }}
   namespace: {{ .Release.Namespace }}
 data:
-  {{ include "newrelic-agent-control.auth.l1IdentityCredentialsKey.clientIdKeyName" . }}: {{ include "newrelic-agent-control.auth.identityClientId" . | b64enc }}
-  {{ include "newrelic-agent-control.auth.l1IdentityCredentialsKey.clientSecretKeyName" . }}: {{ include "newrelic-agent-control.auth.identityClientSecret" . | b64enc }}
+  clientIdKey: {{ include "newrelic-agent-control.auth.identityClientId" . | b64enc }}
```
danielorihuela, line 29: Again, this was configurable before, but it's only used by AC. Thus, we can hardcode the names. Is that it?

Just for my understanding.

charts/agent-control-deployment/templates/deployment-agentcontrol.yaml
paologallinaharbur: yes, that is the idea. 
we provided that possibility, but I think that it was actually a feature we did not need and that created a lot of complexity

paologallinaharbur: I believe that a customer could create a secret with any key in Nr namespace, I cannot think to a common scenario where you do not have that possibility (if so we can change it again in the future)

charts/agent-control-deployment/templates/preinstall-job-register-system-identity.yaml
paologallinaharbur: yes, that is the idea

danielorihuela (dismissed):

PR #1795, opened 2025-07-09
feat(clientIDAuthToken): support managing locally the token creation

paologallinaharbur:
#### Is this a new chart
no
It adds a new way to authenticate `identityClientAuthToken`. It leverage the same clientID and it allows using a local L2 to authenticate and pass just the token

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent-control-deployment/Chart.yaml
+++ b/charts/agent-control-deployment/Chart.yaml
@@ -4,7 +4,7 @@ description: A Helm chart to install New Relic Agent Control on Kubernetes
 
 type: application
 
-version: 0.0.59
+version: 0.0.60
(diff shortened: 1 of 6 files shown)


danielorihuela (commented):
The idea is the the client should do something like

```
# pseudo-bash
token = cli authenticate
helm install with token
```

Am I right?

danielorihuela (approved):

danielorihuela (approved):

paologallinaharbur:
> The idea is the the client should do something like
# pseudo-bash
token = cli authenticate
helm install with token
Am I right?

Yes! you are right!

PR #1804, opened 2025-07-11
[agent-control] fix: do not render namespace manifest if target is release namespace

DavSanchez:
#### Is this a new chart
No

Addresses an issue where, if the AC namespace and the sub-agents namespace were the same, the chart wasn't able to be installed.

This would happen due to Helm rendering the sub-agents namespace manifest and then --create-namespace for a namespace with the same name taking into effect, so Helm would later attempt to apply the rendered namespace manifest to create a namespace that already exists.

The fix introduces an additional check that disables the render of the namespace manifest if the namespace it will attempt to create is the same as the release manifest. This way Helm can be passed the --create-namespace as expected for the AC installation command.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent-control/templates/namespace-agents.yaml
+++ b/charts/agent-control/templates/namespace-agents.yaml
@@ -1,7 +1,10 @@
-{{- if not (lookup "v1" "Namespace" "" (index .Values "agent-control-deployment" "subAgentsNamespace")) -}}
 {{- /* 
-
+    We check both that the namespace does not exist and that it is not the same as the release namespace.
+    This is due to how we instruct the users to install the chart, using the options `-n <NAMESPACE> --create-namespace`,
+    that will make Helm create the release namespace if it does not exist, so we prevent the rendering to take place.
+    Otherwise, the namespace manifest would be rendered and later applied, which would fail as the namespace was created by Helm.
 */ -}}
+{{- if not (lookup "v1" "Namespace" "" (index .Values "agent-control-deployment" "subAgentsNamespace")) -}}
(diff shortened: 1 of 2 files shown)


vjripoll (dismissed):
LGTM

charts/agent-control/templates/namespace-agents.yaml
```diff
@@ -1,4 +1,11 @@
+{{- /* 
+    We check both that the namespace does not exist and that it is not the same as the release namespace.
+    This is due to how we instruct the users to install the chart, using the options `-n <NAMESPACE> --create-namespace`,
+    that will make Helm create the release namespace if it does not exist, so we prevent the rendering to take place.
+    Otherwise, the namespace manifest would be rendered and later applied, which would fail as the namespace was created by Helm.
+*/ -}}
 {{- if not (lookup "v1" "Namespace" "" (index .Values "agent-control-deployment" "subAgentsNamespace")) -}}
+{{- if not (eq .Release.Namespace (index .Values "agent-control-deployment" "subAgentsNamespace")) -}}
```
paologallinaharbur, line 8: we need the same also in the other chart, no?

paologallinaharbur (dismissed):

vjripoll (dismissed):

charts/agent-control/templates/namespace-agents.yaml
DavSanchez: It is added there as well!

vjripoll (approved):

PR #1806, opened 2025-07-11
feat: remove default privileged and update collecter.md

TmNguyen12:
  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/nr-k8s-otel-collector/collector.md
+++ b/charts/nr-k8s-otel-collector/collector.md
@@ -243,7 +243,7 @@ kubeletstats:
 ```
     kubeletstats:
         collection_interval: {{ .Values.receivers.kubeletstats.scrapeInterval }}
-        {{- if include "newrelic.common.privileged" . }}
+        {{- if not (include "newrelic.common.gkeAutopilot" .) }}
         endpoint: "${KUBE_NODE_NAME}:10250"
         auth_type: "serviceAccount"
         insecure_skip_verify: true
--- a/charts/nr-k8s-otel-collector/values.yaml
+++ b/charts/nr-k8s-otel-collector/values.yaml
@@ -96,7 +96,7 @@ daemonset:
   podSecurityContext: {}
   # -- Sets security context (at container level) for the daemonset. Overrides `containerSecurityContext` and `global.containerSecurityContext`
   containerSecurityContext:
-    privileged: true
+    privileged: false
   # -- Sets resources for the daemonset.
   resources: {}
   # -- Sets additional environment variables for the daemonset.


charts/nr-k8s-otel-collector/collector.md
```diff
@@ -243,7 +243,7 @@ kubeletstats:
 ```
     kubeletstats:
         collection_interval: {{ .Values.receivers.kubeletstats.scrapeInterval }}
-        {{- if include "newrelic.common.privileged" . }}
+        {{- if not (include "newrelic.common.gkeAutopilot" .) }}
```
danielstokes, line 246: What's this for?

TmNguyen12: I was just updating this readme to match what was in [here](https://github.com/newrelic/helm-charts/blob/8d23a25590e075e93edf4102762096a3781e19f9/charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml#L70)

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -95,8 +95,8 @@ daemonset:
   # -- Sets security context (at pod level) for the daemonset. Overrides `podSecurityContext` and `global.podSecurityContext`
   podSecurityContext: {}
   # -- Sets security context (at container level) for the daemonset. Overrides `containerSecurityContext` and `global.containerSecurityContext`
-  containerSecurityContext:
-    privileged: true
+  containerSecurityContext: {}
```
dbudziwojskiNR: Do we not prefer to set this as false explicitly for certain flavors of K8s?

dbudziwojskiNR:
Also added the release notes section to the PR's description. Feel free to edit it as needed.

charts/nr-k8s-otel-collector/collector.md
danielstokes: The term `gkeAutopilot` is too overloaded, meaning we're using this for way more than should be intended.  Which I'd say the same for `privileged`, because this has nothing to do with that.

We should either use some other include (yet to be defined) just use something like `secureKubeletPort` or __PLACE_BETTER_NAME_HERE__ .  AKS has similar features, as does EKS, and can be enabled (disabled by default since v1.32 in GKE with or without autopilot) on self managed clusters or disabled in AKS, GKE, prob more.

It's also recommended to disable the read only port.

Anyway, it's already in *common*, so we can just merge anyway without changing the name, since it is just a name, or create a new PR on common to add the include

TmNguyen12: Acknowledged, will update this on new PR when addressing gkeAutopilot

danielstokes (approved):

PR #1817, opened 2025-07-21
[agent-control-deployment] feat: add custom repo url config (NR-436882)

gsanchezgavier:
Adds custom chart repo urls helper

### DIFF
--- a/charts/agent-control-deployment/values.yaml
+++ b/charts/agent-control-deployment/values.yaml
@@ -103,6 +103,9 @@ config:
     # @default -- See `values.yaml`
     port: 51200
 
+  # -- List of allowed chart repository URLs. The Agent Control will only allow to deploy agents from these repositories.
+  # @default -- `[]`(Only newrelic chart repositories allowed: ["https://helm-charts.newrelic.com","https://newrelic.github.io/<>"])
+  allowedChartRepositoryUrl: []
   # -- List of managed agents that will be deployed. The key represents the name of the agent and the value holds the configuration.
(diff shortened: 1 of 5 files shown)


charts/agent-control-deployment/values.yaml
```diff
@@ -103,6 +103,9 @@ config:
     # @default -- See `values.yaml`
     port: 51200
 
+  # -- List of allowed chart repository URLs. The Agent Control will only allow to deploy agents from these repositories.
+  # @default -- `[]`(Only newrelic chart repositories allowed: ["https://helm-charts.newrelic.com","https://newrelic.github.io/<>"])
```
danielorihuela, line 107: How do you know that these are the default values? They change based on the agent type right? This can't be enforced from here.

gsanchezgavier: yes they are defined on each agent type

danielorihuela: Won't this documentation become obsolete once we allow anyone to write their own agent type? They might use chart repositories other than newrelic ones by default.

danielorihuela (approved):

charts/agent-control-deployment/values.yaml
gsanchezgavier: yes it might, once we are there and we understand how this feature interacts with custom agent types we can iterate. I think is still worth to mention, until that happen, which are the defaults or what is the expected behaviour if not define any.

PR #1850, opened 2025-08-05
[agent-control] update proxy test

paologallinaharbur:
#### Is this a new chart
no
it introduce extra variables  and volumes for the install job of flux
Uppgrade is breaking since I am changing the installerJob var name

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent-control/templates/install-flux.yaml
+++ b/charts/agent-control/templates/install-flux.yaml
@@ -32,10 +32,6 @@ spec:
           env:
             {{- toYaml . | nindent 12 }}
           {{- end }}
-          {{- with (index .Values "agent-control-cd").installer.extraVolumeMounts }}
-          volumeMounts:
-            {{- toYaml . | nindent 12 }}
-          {{- end }}
           args:
             - |
               helm repo add newrelic {{ (index .Values "agent-control-cd").chartRepositoryUrl }}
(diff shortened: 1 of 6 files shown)


vjripoll (approved):

charts/agent-control/tests/install_flux_test.yaml
```diff
@@ -3,9 +3,10 @@ tests:
   - it: should leverage correct image tag
     set:
       agent-control-cd:
-        installerImage:
-          tag: 123
-          repository: test
+        installer:
+          image:
+            tag: 123
+            repository: test
```
sigilioso, line 9: I'm not sure if it is worth it: we could extend the uni-tests with the recently introduced `extraVolumes` and `extraEnv`

charts/agent-control/templates/install-flux.yaml
```diff
@@ -28,6 +28,14 @@ spec:
...
               memory: 64Mi
           command: ["/bin/sh", "-c"]
+          {{- with (index .Values "agent-control-cd").installer.extraEnv }}
+          env:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          {{- with (index .Values "agent-control-cd").installer.extraVolumeMounts }}
+          volumeMounts:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
```
sigilioso: I'm not sure if I've got it right but I think we are defining `volumeMounts` twice 🤔

charts/agent-control/tests/install_flux_test.yaml
paologallinaharbur: there is a e2e in agent-control testing it nighlty

sigilioso (approved):
LGTM 🚀

charts/agent-control/templates/install-flux.yaml
paologallinaharbur: you were right!

PR #1857, opened 2025-08-08
[nr-k8s-otel-collector] Routing based on metric not datapoint

Philip-R-Beckwith:
#### Is this a new chart
NO. 

Routing for the otel collector was using data point. 
The issue here is that it seems that only the data points were being routed to the down stream pipelines. 
This means that the rest of the metric's metadata was being chopped off. 
This is _mostly_ not an issue because the data point has access to the scope/resource anyway. 
Thee's honestly not a ton of data attached to the metric context itself. 
Metrics don't even have their own attributes. 

However there are two specific fields that are important for SUMs 

aggregation_temporality
is_monotonic 

By default we're only sending one sum kube_pod_container_status_restarts_total. 
Due to the missing attributes this sum was being rejected at the NR Otel Endpoint. 

There are a couple other SUMS that this agent will send if enabled. 
otelcol_receiver_accepted_log_records
otelcol_process_runtime_total_alloc_bytes 

And others. These were seeming to come through just fine surprisingly. 
However turns out the reason this was is because these metrics are reported from both the daemonset and the deployment. The daemonset uses the metric context for routing not datapoint. 

After fix 
kube_pod_container_status_restarts_total is being reported 
Screenshot 2025-08-07 at 9 39 54 PM

When enabled otelcol_process_runtime_total_alloc_bytes is being reported for both the deployment and daemonset. 
Screenshot 2025-08-07 at 9 39 22 PM

No events are being rejected from the NR otel metric endpoint. 
Screenshot 2025-08-07 at 9 38 30 PM
 
 
 Before the fix... 
 No kube_pod_container_status_restarts_total
Screenshot 2025-08-07 at 10 11 44 PM

When enabled otelcol_process_runtime_total_alloc_bytes is being reported only from for both the daemonset. 
Screenshot 2025-08-07 at 10 11 57 PM

NR otel endpoint rejects 
Screenshot 2025-08-07 at 10 12 09 PM

This investigation was initiated in order to fix the kube_pod_container_status_restarts_total so I'm also updating the cumulativetodelta processor to exclude this metric as well. This metric should stay a cumulative sum instead of a delta. 
 
Screenshot 2025-08-11 at 1 12 24 PM

Metric count before/ after this change. 
We can see that we're getting 1 extra metric after the change (`kube_pod_container_status_restarts_total`)
Screenshot 2025-08-11 at 1 28 48 PM

https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/ottl/contexts/ottlmetric
  - fixes Issue dropping sums due to routing table using datapoints instead of metric. 
  - kube_pod_container_status_restarts_total should be a cumulative sum not a delta sum

#### Checklist
- [x] Chart Version bumped

## 🚀 What's Changed
* Fixes Issue where sums were being dropped due to missing temporality data. 
* `kube_pod_container_status_restarts_total` is ignored by `cumulativetodelta` processor so it remains a cumulative metric.

### DIFF
--- a/charts/newrelic-logging/Chart.yaml
+++ b/charts/newrelic-logging/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart to deploy New Relic Kubernetes Logging as a DaemonSet, supporting both Linux and Windows nodes and containers
 name: newrelic-logging
-version: 1.30.0
+version: 1.31.0
 appVersion: 2.4.0
 home: https://github.com/newrelic/kubernetes-logging
 icon: https://newrelic.com/assets/newrelic/source/NewRelic-logo-square.svg
--- a/charts/newrelic-logging/templates/configmap-lua.yaml
+++ b/charts/newrelic-logging/templates/configmap-lua.yaml
@@ -2,6 +2,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
+  namespace: {{ .Release.Namespace }}  
   name: {{ include "newrelic-logging.fullname" . }}-lua
 data:
   payload.lua: |-
--- a/charts/nr-ebpf-agent/Chart.lock
+++ b/charts/nr-ebpf-agent/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common-library
   repository: https://helm-charts.newrelic.com
-  version: 1.3.1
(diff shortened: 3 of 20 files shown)


ramkrishankumarN (approved):
🎉 🚢 LGTM

dbudziwojskiNR (changes requested):
Lgtm! Please don't forget to update the release notes in the PR notes prior to merging this in.

dbudziwojskiNR (approved):

PR #1858, opened 2025-08-08
[agent-control] feat: Flux self-update

DavSanchez:
#### Is this a new chart

No

This is a feature branch that will aggregate the changes required in the `agent-control` chart to enable its Flux dependency to manage and upgrade itself.

The contents of this PR will also need other additional changes:

- [x] Bumping the AC version in `agent-control-deployment` to one that includes changes in the CLI: #1851
- [x] Add Flux uninstallation functionality. #1864 
- [x] Improve values structure (breaking change)
- [x] Bump agent-control-deployment version with its own values structure improvements (breaking change)
- [x] e2e green on Agent Control feature-branch: https://github.com/newrelic/newrelic-agent-control/pull/1560

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent-control/Chart.yaml
+++ b/charts/agent-control/Chart.yaml
@@ -5,7 +5,7 @@ description: Bootstraps New Relic' Agent Control
 type: application
 version: 0.0.86
 # agent-control-deployment chart default version.
-appVersion: 0.0.68
+appVersion: 0.0.69
 annotations:
   # agent-control-cd chart default version.
   agentControlCdVersion: 0.0.2


charts/agent-control/values.yaml
```diff
@@ -65,8 +87,8 @@ agent-control-cd:
   installer:
     image:
       registry:
-      repository: alpine/helm
-      tag: "3.17.1"
+      repository: newrelic/newrelic-agent-control-cli
```
paologallinaharbur: why do we have this and at the same time installationJob.agentControlCd?

Is it still WIP?

DavSanchez: It's still used on a couple places in `install-flux.yaml`. I think merging the two image configurations, under e.g. `toolkitImage`, might not be desirable yet, at least until we merge the two installation jobs in [NR-448615](https://new-relic.atlassian.net/browse/NR-448615).

I can take that task next and get rid of this. WDYT?

paologallinaharbur: we can sync live so that I get the full context 😄

DavSanchez: Created task to clean up the values: https://new-relic.atlassian.net/browse/NR-450510

charts/agent-control/templates/install.yaml
```diff
@@ -1,6 +1,7 @@
 {{- if (index .Values "agent-control-deployment").enabled -}}
 {{- $installJobName := include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "install-job" ) -}}
 {{- $secretName := include "agent-control.secret.name" . -}}
+{{- $fluxSecretName := printf "%s-flux" (include "agent-control.secret.name" .) -}}
```
paologallinaharbur, line 4: I believe this is wrong, right now I see it in the cluster as ac-deployment-flux

the suffix deployment should not be there

DavSanchez: The referenced secret has the same name, as created by the manifest: https://github.com/newrelic/helm-charts/blob/dadea23bd960235687500f65f545c35c1763c319/charts/agent-control/templates/ac-flux-deployment-secret.yaml#L6

I guess we could create a different helper for this secret and use a different name if it might cause problems. Should we change it?

charts/agent-control/README.md
```diff
@@ -71,29 +71,30 @@ agent-control-deployment:
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
```
paologallinaharbur, line 73: have we generated them one last time?

sigilioso: Yes, It is updated.

charts/agent-control/templates/install-flux.yaml
```diff
@@ -18,8 +17,8 @@ spec:
       serviceAccountName: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" .Release.Name "suffix" "flux-install-job") }}
       containers:
         - name: install-flux
-          image: {{ include "newrelic.common.images.image" ( dict "imageRoot" (index .Values "agent-control-cd").installer.image "context" .) }}
-          imagePullPolicy: {{ (index .Values "agent-control-cd").pullPolicy }}
+          image: {{ include "newrelic.common.images.image" ( dict "imageRoot" .Values.toolkitImage "context" .) }}
+          imagePullPolicy: {{ .Values.toolkitImage.pullPolicy }}
```
DavSanchez, line 21: Thanks!!

charts/agent-control/tests/secret_override_test.yaml
```diff
@@ -3,53 +3,55 @@ chart:
...
-          #   acRemoteUpdate: true
-          #   cdRemoteUpdate: false
-          #   enabled: true
-          #   subAgentsNamespace: newrelic
-          # ------------------------------------------
-          # To debug a failing test, copy the actual Base64 output and decode it with:
-          # echo "BASE64_STRING_HERE" | base64 --decode
-        equal:
+        matchRegex:
+          decodeBase64: true
```
DavSanchez, line 30: TIL

charts/agent-control/values.yaml
```diff
@@ -36,49 +66,33 @@ installationJob:
...
-  # -- Log level for the uninstallation job.
-  # @default -- info
-  logLevel: "info"
+agentControlCd:
 
-# -- Values for the agent-control-deployment chart. Ref.: https://github.com/newrelic/helm-charts/blob/master/charts/agent-control-deployment/values.yaml
-# @default -- See `values.yaml`
-agent-control-deployment:
-  # -- Enable the installation of the Agent Control.
+  # -- Enable the installation of a Continuous Deployment system that can be managed by Agent Control.
```
DavSanchez, line 71: Did we already discuss if we're going to make explicit that we use Flux or should we use a more neutral language like in this example? For me, this particular wording suggest that the CD system has different alternatives.

```suggestion
  # -- Enable the installation of the Continuous Deployment system (Flux) to be managed by Agent Control.
```

DavSanchez (commented):
It's looking great so far!

charts/agent-control/values.yaml
sigilioso: Well, this enables/disables it regardless of supporting only Flux or supporting something else, right? 🙂

PR #1860, opened 2025-08-11
[agent-control] fix: add auth mechanisms for Flux helm commands

DavSanchez:
#### Is this a new chart

No

Enables authentication when adding the Helm repo for the Agent Control CD. Previously, the configuration was hardcoded and did not allowed setting up auth.

  - fixes [NR-447488](https://new-relic.atlassian.net/browse/NR-447488)

#### Checklist
- [ ] Chart Version bumped

## 🚀 What's Changed

[NR-447488]: https://new-relic.atlassian.net/browse/NR-447488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/agent-control/templates/install-flux.yaml
+++ b/charts/agent-control/templates/install-flux.yaml
@@ -1,8 +1,6 @@
 {{- if index .Values "agent-control-cd" }}
 {{- if (index .Values "agent-control-cd" "flux2").enabled -}}
 {{- $installJobName := include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "flux-install-job" ) -}}
-{{- $basicAuthSecret := .Values.installationJob.agentControlCd.repositorySecretReferenceName }}
-{{- $certAuthSecret := .Values.installationJob.agentControlCd.repositoryCertificateSecretReferenceName -}}
 apiVersion: batch/v1
 kind: Job
 metadata:
@@ -32,7 +30,7 @@ spec:
           command: ["/bin/sh", "-c"]
           envFrom:
             - secretRef:
-                name: {{ $basicAuthSecret }}
+                name: {{ .Values.installationJob.agentControlCd.repositorySecretReferenceName }}
                 optional: true
           {{- with (index .Values "agent-control-cd").installer.extraEnv }}
           env:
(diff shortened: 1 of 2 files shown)


charts/agent-control/templates/install-flux.yaml
```diff
@@ -28,6 +28,10 @@ spec:
               cpu: 50m
               memory: 64Mi
           command: ["/bin/sh", "-c"]
+          envFrom:
+            - secretRef:
+                name: {{ .Values.installationJob.agentControlCd.repositorySecretReferenceName }}
+                optional: true
```
paologallinaharbur: why not leveraging {{ if - }} here instead of marking it as optional?

I fear it could create confugsion

charts/agent-control/templates/install-flux.yaml
```diff
@@ -36,10 +40,19 @@ spec:
...
+                helm repo add newrelic-agent-control-cd {{ (index .Values "agent-control-cd").chartRepositoryUrl }} \
+{{- if .Values.installationJob.agentControlCd.repositoryCertificateSecretReferenceName }}
+                  # Use TLS options
+                  --ca-file /etc/flux-cert/ca.crt \
+                  --cert-file /etc/flux-cert/tls.crt \
+                  --key-file /etc/flux-cert/tls.key
+{{- else if .Values.installationJob.agentControlCd.repositorySecretReferenceName }}
+                  # Use basic auth options
+                  --username $username \
+                  --password $password
```
paologallinaharbur: are we assuming the secret to have these two fields?
If so we could mount themdirectly to be sure they are present and create a proper env var:

`pod.spec.containers.env.valueFrom`

DavSanchez: I guess so, basic auth data should have them when it's used with [Flux's HelmRepositories](https://fluxcd.io/flux/components/source/helmrepositories/#secret-reference), so if we are going to use the same value for a `helm` call both fields will be available to us.

DavSanchez: > we could mount themdirectly to be sure they are present and create a proper env var:
> 
> `pod.spec.containers.env.valueFrom`

Isn't that what I'm already doing with `spec.containers[0].envFrom`? The secret keys will be created as proper env vars, so I get to use them in the script.

charts/agent-control/templates/install-flux.yaml
DavSanchez: Good catch! I think it's better. Changed in https://github.com/newrelic/helm-charts/pull/1860/commits/769eb37ee43f29900dd0d00ae0700213ff091fd5!

charts/agent-control/templates/install-flux.yaml
paologallinaharbur: now you are loading whatever it is in the secret, with `eod.spec.containers.env.valueFrom` you specify each variable by name

paologallinaharbur: (random example)
```yaml
      env:
        - name: MY_NODE_NAME
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
        - name: MY_POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
        - name: MY_POD_NAMESPACE
```

DavSanchez: Thanks for the feedback and insights! Changed in https://github.com/newrelic/helm-charts/pull/1860/commits/79c6717b89854d10451c8062d0d2139a81ca12ee

paologallinaharbur (approved):

PR #1862, opened 2025-08-11
[agent-control-deployment] rework system identities

paologallinaharbur:
 - it removes the possibility to pass a pre-existing system identity via values. No one is using this feature and we should not suggest to store the private key in the values.yaml 
 - it re-organise all systemIdentity related values 
 - org_id has nothing to do with the fleet_control configuration and it has been moved
    
#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent-control-deployment/templates/_helpers.tpl
+++ b/charts/agent-control-deployment/templates/_helpers.tpl
@@ -182,11 +182,11 @@ Helper to toggle the creation of the job that creates and registers the system i
 {{- end -}}
 
 {{- if and (include "newrelic-agent-control.auth.customIdentitySecretName" .) (include "newrelic-agent-control.auth.parentIdentity" .) -}}
-  {{- fail "You should not specify in .Values.systemIdentity both a customSecretName and identityClientId identityClientSecret/identityClientAuthToken" -}}
+  {{- fail "You should not specify in .Values.systemIdentity both a secretName and identityClientId identityClientSecret/identityClientAuthToken" -}}
 {{- end -}}
 
 {{- if and (not (include "newrelic-agent-control.auth.customIdentitySecretName" .)) (not (include "newrelic-agent-control.auth.parentIdentity" .)) -}}
-  {{- fail "You must specify in .Values.systemIdentity a customSecretName or identityClientId identityClientSecret/identityClientAuthToken" -}}
+  {{- fail "You must specify in .Values.systemIdentity a secretName or identityClientId identityClientSecret/identityClientAuthToken" -}}
 {{- end -}}
 
 {{- if not (.Values.systemIdentity).organizationId -}}
--- a/charts/agent-control-deployment/tests/preinstall_job_test.yaml
+++ b/charts/agent-control-deployment/tests/preinstall_job_test.yaml
@@ -11,7 +11,7 @@ tests:
   - it: by default it fails with missing values
     asserts:
       - failedTemplate:
-          errorMessage: You must specify in .Values.systemIdentity a customSecretName or identityClientId identityClientSecret/identityClientAuthToken
+          errorMessage: You must specify in .Values.systemIdentity a secretName or identityClientId identityClientSecret/identityClientAuthToken
 
 
   - it: it fails with no organizationId
@@ -46,7 +46,7 @@ tests:
         organizationId: test
     asserts:
       - failedTemplate:
-          errorMessage: You should not specify in .Values.systemIdentity both a customSecretName and identityClientId identityClientSecret/identityClientAuthToken
+          errorMessage: You should not specify in .Values.systemIdentity both a secretName and identityClientId identityClientSecret/identityClientAuthToken
 
   - it: with fleet disabled, the job should template correctly 0 documents.
     set:


charts/agent-control-deployment/values.yaml
```diff
@@ -150,55 +150,44 @@ config:
...
-          # This options is mutually exclusive with `base64`.
-          plain:
-
-# These options provides the required info to run the pre-install job that creates the system identity that are used communicating via OpAMP.
-# You can either authenticate via identityClientId/identityClientSecret or pass directly an `identityClientAuthToken` and manage locally the authentication.
-# The authToken can be retrieve via the cli command "newrelic-auth-cli authenticate ...".
-# System identity creation is executed only once. Subsequent upgrades will not attempt to create the identity again.
+# -- Configuration for the system identity registration process.
+# These options provides the required data to run the pre-install job that creates the system identity that are used communicating via OpAMP.
+# System identity registration is executed only once. Subsequent upgrades will not attempt to create the identity again.
```
sigilioso: Nit: 
```suggestion
# These options provides the required data to run the pre-install job that creates the system identity that is used when communicating via OpAMP.
# System identity registration is executed only once. Subsequent upgrades will not attempt to create the identity again.
```

charts/agent-control-deployment/values.yaml
```diff
@@ -150,55 +150,44 @@ config:
...
 # @default -- See `values.yaml`
-systemIdentityRegistration:
-  # -- Volumes to mount in the containers
-  extraVolumes: []
-  # -- Defines where to mount volumes specified with `extraVolumes`
-  extraVolumeMounts: []
+systemIdentity:
+
+  # -- Set it to false to disable the registration of a new system identity.
+  # Leverage this option to configure a pre-existing system identity via secret.
```
sigilioso:
```suggestion
  # Set this to `false` to configure a pre-existing system identity via secret.
```

charts/agent-control-deployment/values.yaml
```diff
@@ -150,55 +150,44 @@ config:
...
+  # The secret should be already created in the namespace having as keys "CLIENT_ID" and "private_key" of the identity to leverage.
```
sigilioso, line 162: Is this documented somewhere else? If not we might want to add more details such as the secret type. It would be useful to also add an example (out of the scope of this PR)

charts/agent-control-deployment/values.yaml
```diff
@@ -150,55 +150,44 @@ config:
...
+  create: true
+
+  # -- Set it to customize the name of the secret that is created and mounted in the agentControl deployment pod
+  customSecretName:
```
sigilioso: do we need the `custom` prefix?

charts/agent-control-deployment/values.yaml
```diff
@@ -150,55 +150,44 @@ config:
...
+
+  # -- Configuration for the parent identity
+  # You can either authenticate via ClientId/ClientSecret or pass directly an `AuthToken` and manage locally the authentication.
+  # The authToken can be retrieve via the cli command "newrelic-auth-cli authenticate ...".
+  parentIdentity:
+    # -- Identity clientId to use.
+    clientId: ""
+    # -- Identity clientSecret to use.
+    clientSecret: ""
+    # -- Identity auth token. This option takes precedence over secret and skips authentication.
```
sigilioso, line 185: Do clientID/clientSecret also take precedence over the secret? Do they also take precedence over authToken? We should specify or just fail if they are mutually exclusive.

sigilioso (commented):
Thanks for sorting this out and simplifying! I've left some comments, merely _nits_

charts/agent-control-deployment/values.yaml
paologallinaharbur: I do not think so 😕 I think that before adding those lines the only way to know it was checking our chart source code

charts/agent-control-deployment/values.yaml
paologallinaharbur: I'll check again!

charts/agent-control-deployment/values.yaml
sigilioso: Should we document it here?

charts/agent-control-deployment/values.yaml
paologallinaharbur: I added tests and clear fail for that

charts/agent-control-deployment/values.yaml
paologallinaharbur: i'll remove it!

charts/agent-control-deployment/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: A Helm chart to install New Relic Agent Control on Kubernetes
 
 type: application
 
-version: 0.0.66
+version: 0.0.67
```
sigilioso, line 7: Maybe we can avoid bumping the version because we are going to merge to a feature branch, although it will make the CI unhappy 🙃

charts/agent-control-deployment/templates/_helpers.tpl
```diff
@@ -150,101 +150,62 @@ readOnlyRootFilesystem: true
...
-  {{- $clientId := include "newrelic-agent-control.auth.secret.clientId.data" . -}}
-
-  {{- if and $privateKey $clientId -}}
-    true
-  {{- else if or $privateKey $clientId -}}
-    {{- fail "If you provide your own system identity data you have to provide both private key and client id" -}}
-  {{- end -}}
-{{- end -}}
+{{- if and (include "newrelic-agent-control.auth.customIdentitySecretName" .) (include "newrelic-agent-control.auth.parentIdentity" .) -}}
+  {{- fail "You should not specify in .Values.systemIdentity both a customSecretName and identityClientId identityClientSecret/identityClientAuthToken" -}}
```
sigilioso: I think we also need to remove the `custom` prefix from messages (and maybe the helper names) `customSecretName` --> `secretName` as done in `Values`

sigilioso (commented):
I think there is just a minor thing regarding error messages / names left! Thanks!

charts/agent-control-deployment/templates/_helpers.tpl
paologallinaharbur: ups 100% right!

charts/agent-control-deployment/Chart.yaml
paologallinaharbur: I think I'll keep it for ci convenience!

PR #1864, opened 2025-08-12
[agent-control] feat: remove flux crs on ac uninstallation [WIP]

sigilioso:
#### Is this a new chart
No

Extend the agent-control uninstallation job to also remove the CRs used to handle Flux.

* On top of #1858
* Requires a new release of agent-control so the corresponding CLI includes the used commands.

#### Checklist
- [ ] Chart Version bumped


charts/agent-control/templates/install-flux.yaml
```diff
@@ -51,19 +51,21 @@ spec:
             - |
               HELM_RELEASE_EXISTS=$(kubectl get HelmRelease agent-control-cd -n {{ .Release.Namespace }})
               if [ -z "$HELM_RELEASE_EXISTS" ]; then
-                helm repo add newrelic-agent-control-cd {{ (index .Values "agent-control-cd").chartRepositoryUrl }} \
-{{- if .Values.installationJob.agentControlCd.repositoryCertificateSecretReferenceName }}
-                  # Use TLS options
+                echo "Adding agent-control-cd repository..."
+                helm repo add newrelic-agent-control-cd {{ (index .Values "installationJob").agentControlCd.chartRepositoryUrl }}
+{{- /* Use TLS Options */ -}}
+{{- if .Values.installationJob.agentControlCd.repositoryCertificateSecretReferenceName }} \
```
DavSanchez, line 54: Thankss!

charts/agent-control/templates/install-flux.yaml
```diff
@@ -28,11 +28,8 @@ spec:
               cpu: 50m
               memory: 64Mi
           command: ["/bin/sh", "-c"]
-          {{- if (or
-            .Values.installationJob.agentControlCd.repositorySecretReferenceName
-            (index .Values "agent-control-cd").installer.extraEnv
-          ) }}
           env:
```
DavSanchez, line 31: What happens if none of the conditions below hold?

It will be evaluated to

```yaml
env: # nothing else. Empty... what type?
```

But not

```yaml
env: [] # empty list
```

Is that the same?

charts/agent-control/templates/install-flux.yaml
```diff
@@ -43,27 +40,29 @@ spec:
...
+{{- /* Use basic auth options */ -}}
+{{- else if .Values.installationJob.agentControlCd.repositorySecretReferenceName }} \
                   --username $HELM_REPO_USERNAME \
                   --password $HELM_REPO_PASSWORD
 {{- end }}
-                helm upgrade --install agent-control-cd newrelic-agent-control-cd/{{ (index .Values "agent-control-cd").chartName }} \
-                  --version {{ (index .Values "agent-control-cd").chartVersion }} \
+                echo "Installing agent-control-cd..."
+                helm upgrade --install agent-control-cd newrelic-agent-control-cd/{{ (index .Values "installationJob").agentControlCd.chartName }} \
+                  --version {{ (index .Values "installationJob").agentControlCd.chartVersion }} \
```
DavSanchez, line 65:
```suggestion
                helm upgrade --install agent-control-cd newrelic-agent-control-cd/{{ .Values.installationJob.agentControlCd.chartName }} \
                  --version {{ .Values.installationJob.agentControlCd.chartVersion }} \
```

charts/agent-control/templates/install-flux.yaml
```diff
@@ -43,27 +40,29 @@ spec:
...
-          {{- end }}
           args:
             - |
               HELM_RELEASE_EXISTS=$(kubectl get HelmRelease agent-control-cd -n {{ .Release.Namespace }})
               if [ -z "$HELM_RELEASE_EXISTS" ]; then
-                helm repo add newrelic-agent-control-cd {{ (index .Values "agent-control-cd").chartRepositoryUrl }} \
-{{- if .Values.installationJob.agentControlCd.repositoryCertificateSecretReferenceName }}
-                  # Use TLS options
+                echo "Adding agent-control-cd repository..."
+                helm repo add newrelic-agent-control-cd {{ (index .Values "installationJob").agentControlCd.chartRepositoryUrl }}
```
DavSanchez, line 52:
```suggestion
                helm repo add newrelic-agent-control-cd {{ .Values.installationJob.agentControlCd.chartRepositoryUrl }}
```

These values are already in place so can be used no? Or are we waiting to your other task to fix them all?

charts/agent-control/templates/install-flux.yaml
sigilioso: I'm leaving the values (and its usage even if just improving syntax) for the follow-up PR if that's OK :)

charts/agent-control/templates/install-flux.yaml
sigilioso: I'm leaving the values (and its usage even if just improving syntax) for the follow-up PR if that's OK :)

DavSanchez: Absolutely! 😃

charts/agent-control/templates/install-flux.yaml
sigilioso: I've just tested it and it works the same `env:`, `env: []` and no env at all. A testing pod to reproduce it:

```yaml
apiVersion: v1
kind: Pod
metadata:
    name: test-pod
    labels:
        app: test
spec:
    containers:
    - name: test-container
      image: busybox
      command: ["sh", "-c", "echo Hello Kubernetes && sleep 3600"]
      env:
      resources:
        requests:
          memory: "64Mi"
          cpu: "250m"
        limits:
          memory: "128Mi"
          cpu: "500m"
```
The IDE complains because `env` doesn't have the expected type but it works.

Before we had a bug when `(index .Values "agent-control-cd").installer.extraEnv` had something but `.Values.installationJob.agentControlCd.repositorySecretReferenceName` didn't.

We could do better and add a new helper to build the environment variables for the job (assuring an empty if none is needed) but I  think we can leave it for a different PR. WDYT?

charts/agent-control/templates/install-flux.yaml
danielorihuela, line 52: Nit: Now that we don't have dashes in the value field, we can avoid the index notation.


```suggestion
                helm repo add newrelic-agent-control-cd {{ .Values.installationJob.agentControlCd.chartRepositoryUrl }}
```

danielorihuela: Oh no. We have one with dashes and one without dashes in different places.
This is a bit confusing. Should we take the opportunity to make it consistent and avoid dashes everywhere? In another PR.

sigilioso: I'm taking care of that here: https://github.com/newrelic/helm-charts/pull/1876

(of both, avoiding the index notation and making everything as consistent as possible)

PR #1876, opened 2025-08-14
[agent-control] clean and improve configuration

sigilioso:
#### Is this a new chart
No

Improves the way the agent control config is treated and it brings some coherence considering that the main purpose of `agent-control` chart is launching the jobs to install the underlying components.

* On  top of #1858 (feature branch)
* The jobs for installation and uninstallation are not unified yet (we can do it separately if we are happy with the configuration)
* Refactors unit-tests requiring base64 decoding
* The default version for the `agent-control-deployment` chart is set as Chart annotation.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent-control/values.yaml
+++ b/charts/agent-control/values.yaml
@@ -14,9 +14,10 @@ toolkitImage:
 
 
 installation:
-  # -- Log level for installation.
-  # @default -- debug
-  logLevel: "debug"
+  log:
+    # -- Log level for installation.
+    # @default -- debug
+    level: "debug"
 
   # -- Extra environment variables
   extraEnv: []
@@ -26,9 +27,10 @@ installation:
   extraVolumeMounts: []
 
 uninstallation:
-  # -- Log level for uninstallation.
-  # @default -- debug
-  logLevel: "debug"
+  log:
+    # -- Log level for installation.
+    # @default -- debug
+    level: "debug"
 
 
 agentControlDeployment:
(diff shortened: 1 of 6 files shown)


charts/agent-control/values.yaml
```diff
@@ -12,95 +12,80 @@ toolkitImage:
...
-
-# -- Values for the agent-control-deployment chart. Ref.: https://github.com/newrelic/helm-charts/blob/master/charts/agent-control-deployment/values.yaml
-# @default -- See `values.yaml`
-agent-control-deployment:
-  # -- Enable the installation of the Agent Control.
+
+installation:
+  # -- Log level for installation.
+  # @default -- debug
+  logLevel: "debug"
```
paologallinaharbur: nit: the agent control has:
```
log:
  level
```

charts/agent-control/values.yaml
```diff
@@ -12,95 +12,80 @@ toolkitImage:
...
+  # -- Values for the agent-control-deployment chart. Ref.: https://github.com/newrelic/helm-charts/blob/master/charts/agent-control-deployment/values.yaml
+  # @default -- See `values.yaml`
+  chartValues:
+    # -- Namespace where agents are deployed
+    # @default -- "newrelic"
+    subAgentsNamespace: "newrelic"
+
+  # -- The repository URL from where the `agent-control-deployment` chart will be installed.
+  # @default -- https://helm-charts.newrelic.com
+  chartRepositoryUrl: https://helm-charts.newrelic.com
```
paologallinaharbur: should we say in the comment where to add the url in the allowed list?

paologallinaharbur: (in the subchart I mean)

sigilioso: Even if they are different things (this url is not validated with the allowed list for the installation) it is related. I'm adding the comment here and in the `agentControlCd` parameter as well. Thanks for pointing out!

charts/agent-control/values.yaml
sigilioso: I've just updated it in the latest commit

charts/agent-control/values.yaml
sigilioso: Reference added in the latest commit

charts/agent-control/templates/ac-deployment-secret.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if (index .Values "agent-control-deployment").enabled -}}
+{{- if .Values.agentControlDeployment.enabled -}}
```
paologallinaharbur, line 1: here

PR #1883, opened 2025-08-20
[agent-control-deployment] hardcode cd release name

danielorihuela:
#### Is this a new chart

no

Hardcodes the "cd" release name. Otherwise, agent control will receive an empty string.


sigilioso (approved):
LGTM

PR #1884, opened 2025-08-20
fix: create flux helm release before installing agent control

danielorihuela:
#### Is this a new chart

no

Creates flux helm release before installing ac. Otherwise, AC will log an error when the cd remote update is enabled at installation time.


sigilioso (approved):
LGTM

sigilioso (approved):

PR #1885, opened 2025-08-20
[agent-control] unify installation and uninstallation jobs

sigilioso:
#### Is this a new chart
No

This PR unifies installation and installation jobs, reducing the number of jobs while keeping granularity regarding permissions

#### Checklist
- [x] Chart Version bumped


sigilioso:
Agent Control e2e check: https://github.com/newrelic/newrelic-agent-control/pull/1574

DavSanchez (approved):
Thanks! LGTM! 🚀

charts/agent-control/templates/installation-job.yaml
```diff
@@ -14,9 +16,30 @@ spec:
...
+              - key: tls.key
+                path: tls.key
+        {{- end }}
+        {{- with .Values.installation.extraVolumes }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      {{- end }}
       containers:
-        - name: install-flux
+        - name: install-agent-control-deployment
```
danielorihuela: Nit: Change the name to better express what it does.

The name is misleading. This container installs the "deployment" and "flux". Maybe `install-agent-control` is enough?

charts/agent-control/templates/installation-job.yaml
```diff
@@ -43,60 +65,77 @@ spec:
...
-                exit 0
               fi
-          volumeMounts:
-            - name: flux-values
-              mountPath: /tmp/flux-values
-              readOnly: true
+{{- end }}
+{{- if .Values.agentControlDeployment.enabled }}
+           {{- if .Values.agentControlCd.enabled }}
+              echo "Creating agent-control-cd resources..."
```
danielorihuela: Mmmm, shouldn't me move that block on top? Otherwise, if a client enables `cd` and disables `deployment`, resources won't be created. I know it doesn't make sense for a client to do that, but I think it makes more sense.

Not needed to do now. It can be a sustain task.

danielorihuela (approved):

charts/agent-control/templates/installation-job.yaml
sigilioso: You are 100% right

charts/agent-control/templates/installation-job.yaml
sigilioso: It is actually the expected behavior: the `agent-control-cd` resources (the CRs representing Flux's helm-release) are only useful if the Agent Control is installed, we don't need such resources if `agentControlDeployment` is disabled.

charts/agent-control/templates/installation-job.yaml
sigilioso: Updated here: [6aea5a7](https://github.com/newrelic/helm-charts/pull/1885/commits/6aea5a7ee0bc526d9b504a03d8e5eea14515cdf7)

charts/agent-control/templates/installation-job.yaml
```diff
@@ -1,5 +1,7 @@
-{{- if .Values.agentControlCd.enabled }}
-{{- $installJobName := include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "flux-install-job" ) -}}
+{{- if (or .Values.agentControlDeployment.enabled .Values.agentControlCd.enabled) -}}
+{{- $installJobName := include "newrelic.common.naming.truncateToDNSWithSuffix" ( dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "install-job" ) -}}
+{{- $secretName := include "agent-control.secret.name" . -}}
+{{- $fluxSecretName := printf "%s-flux" (include "agent-control.secret.name" .) -}}
```
paologallinaharbur: could we follow "just in case" the same naming conventions we have for subAgents localData/values?
This in case in the future we start receiving the whole config and we need to do the classical local vs remote thing

charts/agent-control/templates/installation-job.yaml
```diff
@@ -14,9 +16,30 @@ spec:
...
+            secretName: {{ .Values.agentControlCd.repositoryCertificateSecretReferenceName }}
+            items:
+              - key: ca.crt
+                path: ca.crt
+              - key: tls.crt
+                path: tls.crt
+              - key: tls.key
+                path: tls.key
+        {{- end }}
+        {{- with .Values.installation.extraVolumes }}
```
paologallinaharbur, line 40: in this case we have extra volumes just whenever flux is enabled, is it intended?

charts/agent-control/templates/installation-job.yaml
```diff
@@ -43,60 +65,77 @@ spec:
...
+            - name: flux-values
+              mountPath: /tmp/flux-values
+              readOnly: true
+            {{- if .Values.agentControlCd.repositoryCertificateSecretReferenceName }}
+            - name: flux-cert-secrets
+              mountPath: /etc/flux-cert
+              readOnly: true
+            {{- end }}
+            {{- end }}
+            {{- with .Values.installation.extraVolumeMounts }}
```
paologallinaharbur, line 85: in this case we are adding them also if flux is disabled, there is a discrepancy

charts/agent-control/templates/installation-job.yaml
```diff
@@ -43,60 +65,77 @@ spec:
...
                 echo "agent-control-cd HelmRelease already exists, skipping installation."
-                exit 0
               fi
-          volumeMounts:
-            - name: flux-values
-              mountPath: /tmp/flux-values
-              readOnly: true
+{{- end }}
+{{- if .Values.agentControlDeployment.enabled }}
+           {{- if .Values.agentControlCd.enabled }}
```
paologallinaharbur, line 119: maybe a comment here, it is a bit confusing

charts/agent-control/templates/installation-job.yaml
paologallinaharbur: echo "Creating agent-control-cd resources owned by agent contorl..."

charts/agent-control/templates/installation-job.yaml
sigilioso: Thanks for noticing! I'll rework that

charts/agent-control/templates/installation-job.yaml
sigilioso: Good catch! It matches the previous behavior (no volumes were allowed in the installer for agent-control-deployment) but I think its wrong. I'm fixing it and extending unit tests. Thanks!

charts/agent-control/templates/installation-job.yaml
```diff
@@ -43,60 +65,77 @@ spec:
...
-{{- /* Use basic auth options */ -}}
-{{- else if .Values.agentControlCd.repositorySecretReferenceName }} \
+              {{- /* Use basic auth options */ -}}
+              {{- else if .Values.agentControlCd.repositorySecretReferenceName }} \
                   --username $HELM_REPO_USERNAME \
                   --password $HELM_REPO_PASSWORD
-{{- end }}
+              {{- end }}
                 echo "Installing agent-control-cd..."
                 helm upgrade --install agent-control-cd newrelic-agent-control-cd/{{ .Values.agentControlCd.chartName }} \
```
paologallinaharbur, line 108: isn't it configurable after the change of @danielorihuela why is it hardcoded? https://github.com/newrelic/helm-charts/pull/1888/files

sigilioso: That's addressed in https://github.com/newrelic/helm-charts/pull/1886

sigilioso:
Re-triggering e2e here: https://github.com/newrelic/newrelic-agent-control/pull/1574

charts/agent-control/templates/installation-job.yaml
sigilioso: Updated here: [e84e508](https://github.com/newrelic/helm-charts/pull/1885/commits/e84e50841ab436250e6f73b021614b1c9082e290) LMKWYT

charts/agent-control/templates/installation-job.yaml
sigilioso: Addressed here: [422a966](https://github.com/newrelic/helm-charts/pull/1885/commits/422a9669b65865183ca1843eccd7b049479e6937)

charts/agent-control/templates/installation-job.yaml
sigilioso: Some comments here: [2867b99](https://github.com/newrelic/helm-charts/pull/1885/commits/2867b99e3b75f335053bea8eac0db9d604d8a0ff)

charts/agent-control/templates/installation-job.yaml
```diff
@@ -0,0 +1,151 @@
...
+            {{- /* installer extra volumeMounts */ -}}
+            {{- with .Values.installation.extraVolumeMounts }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
+          command: ["/bin/sh", "-c"]
+          args:
+            - |
+{{- /* Install agent-control-cd through helm unless the corresponding helm-release already exists */ -}}
+{{- if .Values.agentControlCd.enabled }}
+              HELM_RELEASE_EXISTS=$(kubectl get HelmRelease agent-control-cd -n {{ .Release.Namespace }})
```
gsanchezgavier, line 93: would make sense to make `agent-control-cd` a helper with some comment regarding where is defined? IIRC this is hardcoded in the cli right and cannot be changed?

charts/agent-control/templates/installation-job.yaml
```diff
@@ -0,0 +1,151 @@
...
+                  --ca-file /etc/flux-cert/ca.crt \
+                  --cert-file /etc/flux-cert/tls.crt \
+                  --key-file /etc/flux-cert/tls.key
+              {{- /* Use basic auth options */ -}}
+              {{- else if .Values.agentControlCd.repositorySecretReferenceName }} \
+                  --username $HELM_REPO_USERNAME \
+                  --password $HELM_REPO_PASSWORD
+              {{- end }}
+                echo "Installing agent-control-cd..."
+                helm upgrade --install agent-control-cd newrelic-agent-control-cd/{{ .Values.agentControlCd.chartName }} \
```
gsanchezgavier, line 108: is it possible to run the upgrade here? if not perhaps `helm install` will be more clear?

charts/agent-control/templates/installation-job.yaml
sigilioso: I think it will be configurable after https://github.com/newrelic/helm-charts/pull/1886, would it make sense to tackle that there (either use the values or a helper)

gsanchezgavier: oh is this pr actually 👍

charts/agent-control/templates/installation-job.yaml
sigilioso: `helm install` would be definitely more clear as upgrades are not expected here, but using `helm upgrade` also covers some "non-supported" use cases such as installing `agent-control-cd` separately before installing `agent-control`.

Do you think we should fail the installation in an `agent-control-cd` release already exists?

gsanchezgavier: is true, we will be blocking that possibility  instead of upgrading it, i didn't though about it, just was thinking about readability. 
If we don't allow it , we will understand if is a valid use case in the future and add it later WDYT? I don't have storng opinion though, if you think it can be useful let's keep it

sigilioso: I'm not completely sure if it will be useful but It might help with some sort of testing scenarios. Besides, currently the UX when the job fails is not the best, so this is one possible error lest. I'm keeping the upgrade for now and we can revisit in the future.

PR #1886, opened 2025-08-20
feat: make flux release name configurable

danielorihuela:
  - fixes #

#### Checklist
- [ ] Chart Version bumped

## 🚀 What's Changed


charts/agent-control/templates/install.yaml
```diff
@@ -61,5 +62,6 @@ spec:
             {{- if .Values.agentControlDeployment.repositoryCertificateSecretReferenceName }}
                 --repository-certificate-secret-reference-name={{ .Values.agentControlDeployment.repositoryCertificateSecretReferenceName }} \
             {{- end }}
-                --chart-name={{ .Values.agentControlDeployment.chartName }}
+                --chart-name={{ .Values.agentControlDeployment.chartName }} \
+                --release-name=agent-control-deployment
```
danielorihuela, line 66: Release name is mandatory in the agent control cli. Hence, we have to hardcode the deployment release name here until we work on https://new-relic.atlassian.net/browse/NR-441446.

Or we could leave it hardcoded here. No problem.

sigilioso: I'm not sure if there something else to be done in that task, but isn't it easier to simple include:

```yaml
agentControlDeployment:
  releaseName: "agent-control-deployment"
```

In values.yaml?

charts/agent-control/templates/_helpers.tpl
```diff
@@ -13,6 +13,9 @@ overrides:
 
   {{- if not .Values.agentControlCd.enabled -}}
     {{- $_ := set $config "cdRemoteUpdate" false -}}
+  {{- else -}}
+    {{- $newValues := merge $config.config (dict "cdReleaseName" .Values.agentControlCd.releaseName) -}}
+    {{- $_ := set $config "config" $newValues -}}
```
sigilioso: Could we extend the unit-tests to check this new behavior?

sigilioso: By the way, there is something odd. In #1888, I see `cdRelaseName` and `cdRemoteUpdate` at the same level but here I see something like:

```yaml
cdRemoteUpdate: false # at root level
config:
  cdReleaseName: "release-name" # at config level
```

Is it working as expected? 🤔

charts/agent-control/templates/install.yaml
danielorihuela: Then you can modify it from the values and it won't do anything. That's the reason I didn't want to do that. Cause it seemed pointless for now.

sigilioso: I don't understand. I meant adding it to values *also use it here** `{{ .Values.agentControlDeployment.releaseName }}` 🤔 

Maybe you don't want to do that because it isn't fully supported in Agent Control. Is that the case?

charts/agent-control/templates/_helpers.tpl
danielorihuela: > Is it working as expected? 

Last time I tested it locally, it worked. I could install and uninstall AC with a given name like "this-is-my-new-release-name". I want to try it again just in case. But I'm having problems with tilt now.

So I think it's working as expected. Haven't tested disabled cd remote update.

> Could we extend the unit-tests to check this new behavior?

Will do.

> By the way, there is something odd. In #1888, I see cdRelaseName and cdRemoteUpdate at the same level but here I see something like:

I think the `cdRemoteUpdate` might be wrong. However, I didn't test it, and I didn't check how it's done. I assume it works as expected if it's merged.

danielorihuela: > Is it working as expected?

The secret has the data at the correct level.

Image

Helm release has the correct name.

Image

No error is logged in AC, so the updated worked with the new release name.
Image

Uninstallation also worked.

sigilioso: I think `cdRemoteUpdate` is the one that is broken, following up our _offline_ conversation, we will address it separately.

charts/agent-control/templates/install.yaml
danielorihuela: That's what I wanted to say yes. It's not supported in AC, so I don't want to include it.
I think it makes sense to make it in the other task.

Will it be relevant, I don't think so, since we will attack it shortly after that (I believe). However, if for whatever reason we end up not making it configurable, why should we have the flag there?

sigilioso (approved):
LGTM 🚀

charts/agent-control/templates/_helpers.tpl
```diff
@@ -13,6 +13,10 @@ overrides:
 
   {{- if not .Values.agentControlCd.enabled -}}
     {{- $_ := set $config "cdRemoteUpdate" false -}}
+  {{- else -}}
```
paologallinaharbur, line 16: after rebasing we will have the agent-control.secret.name for flux as well in the helpers, can we make those those names dependent on the release name as well?

At the end that is going to be the "agentID"

paologallinaharbur: I mean these:
```
{{- define "agent-control-deployment.secret.name" -}}
  {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" .Release.Name "suffix" "deployment-local") }}
{{- end -}}

{{- define "agent-control-cd.secret.name" -}}
  {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" .Release.Name "suffix" "cd-local") }}
{{- end -}}
```

 believe we should not use `.Release.Name`, but the one chose by the customer for each one, and have only -local as suffix

(@sigilioso unless you are already hanging this elsewhere)

sigilioso: I'm not. I merely introduced one of the missing helpers. I totally agree on using the release name as base for those names

danielorihuela:
Closing in favour of https://github.com/newrelic/helm-charts/pull/1894

PR #1895, opened 2025-08-26
[agent-control] test: add AC e2e test workflow

gsanchezgavier:
Adds a workflow to run e2e test whenever any of the Agent Control charts are modified.

### DIFF
--- a/.github/workflows/agent-control-e2e.yml
+++ b/.github/workflows/agent-control-e2e.yml
@@ -1,0 +1,88 @@
+# These tests fetches the AC repo and executes the K8s e2e tests using the charts of the
+# current branch.
+name: Agent Control e2e test
+
+on:
+  pull_request:
+    paths:
+      - 'charts/agent-control/**'
+      - 'charts/agent-control-cd/**'
+      - 'charts/agent-control-deployment/**'
+      - '.github/workflows/agent-control-e2e.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  agent-control-e2e:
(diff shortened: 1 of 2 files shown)


.github/workflows/agent-control-e2e.yml
```diff
@@ -0,0 +1,88 @@
...
+jobs:
+  agent-control-e2e:
+    name: AC e2e scenario - ${{ matrix.group }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        group: ["infra", "apm", "dynamic", "custom-repo", "proxy"]
+    steps:
+      - uses: actions/checkout@v4
```
paologallinaharbur, line 25: should we support somehow getting a different branch from AC that is not main?

.github/workflows/agent-control-e2e.yml
```diff
@@ -0,0 +1,88 @@
+# These tests fetches the AC repo and executes the K8s e2e tests using the charts of the
```
paologallinaharbur, line 1: I really like the approach, it seems way easier than expected!

.github/workflows/agent-control-e2e.yml
gsanchezgavier: it is as simple as adding `ref: my/branch/tag` here. I was using it in order to do some tests also

sigilioso: Could we include a comment? We'd have an example if we needed to use it

Philip-R-Beckwith (approved):

PR #1923, opened 2025-09-12
[nr-k8s-otel-collector] Adding Metrics Generation to Deployment 

Philip-R-Beckwith:
#### Is this a new chart
NO. 

Synthesizing two new metrics [node.cpu.request.percentage, node.memory.request.percentage] using metrics generation processor. To synthesize metrics on NR's side we'd either need to create an ingest scraper or do the synthesizing via NRQL. The issue with synthesizing metrics in NRQL is that the queries needed to aggregate metrics in particular are fairly gratuitous. For example.... 
``` SQL
FROM(
  FROM(
    FROM Metric 
    SELECT latest(kube_pod_container_resource_requests) AS containerRequest
    WHERE resource = 'cpu' AND container_phase != 'terminated'
    FACET k8s.container.name, k8s.pod.name, k8s.node.name, k8s.namespace.name LIMIT MAX
  )
  SELECT sum(containerRequest) AS totalRequests FACET k8s.node.name LIMIT MAX
) 
JOIN (
  FROM Metric
  SELECT latest(kube_node_status_allocatable) AS totalAllocatable
  WHERE resource = 'cpu' 
  FACET k8s.node.name LIMIT MAX
) ON k8s.node.name
SELECT sum(totalRequests)/ sum(totalAllocatable) * 100
```
This query is simply getting the average CPU utilization as a percentage across all of a customer's nodes.
Yet it has to be so monstrous because we have to be very careful when doing math on metrics to insure we are dividing the cpu usage and cpu capacity from the same nodes. 

From our experimentation with the metrics generation processor we've been able to reduce this query down to this. 
``` SQL
FROM Metric SELECT average(node.cpu.usage.percentage) 
```

N/A

The metrics generation processor, much like most processors, is stateless. It cannot do work across batches. 
The issue here is that the routing connector breaks apart batches based on the context. Since most of our routing is based on metrics or datapoints, none of our batches on the other end of those routers will have all of the info we need to aggregate our metrics with the generation processor. 

This is why the metric generation pipeline was made. It allows us to create the new metrics before routing.
The other option was to do this work in the ingress pipeline, however customers can export the raw data from the ingress pipeline so it should not contain metrics that we are synthesizing. 

Updated the helm ignore file. 
Helm seems to be trying to use ever file and every file in ever subdirectory to create the build. 
While this doesn't seem to be causing any behavioral issues (yet), this _may_ be leading to overly inflated helm chart sizes. However it is defiantly leading to this weird log, which may be seen as a bug or at the very least isn't exactly a polished experience. 
Screenshot 2025-09-12 at 10 20 34 AM

After the helm ignore update:
Screenshot 2025-09-12 at 10 20 17 AM

#### Checklist
- [x] Chart Version bumped

## 🚀 What's Changed
* Synthesizing two new metrics [node.cpu.request.percentage, node.memory.request.percentage] using metrics generation processor.


charts/nr-k8s-otel-collector/Chart.yaml
dbudziwojskiNR, line 1: Please bump the `appVersion` to 1.5.0 to reflect the new image version

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -191,6 +191,73 @@ data:
...
+            action: update
+            new_name: kube_node_resource_requests
+            operations:
+              - action: aggregate_labels
+                label_set:
+                  - resource
+                  - node
+                aggregation_type: sum
+
+      metricstransform/extrac_metrics_by_label:
```
ramkrishankumarN, line 212: nit: /s/extrac/extract

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -191,6 +191,73 @@ data:
...
+        transforms:
+          - include: kube_node_resource_requests
+            action: insert
+            new_name: kube_node_cpu_requests
```
ramkrishankumarN, line 216: Are we missing a `match_type: regexp` here?

PR #1924, opened 2025-09-13
[nr-k8s-otel-collector] Adding Metrics Generation to Daemonset

Philip-R-Beckwith:
#### Is this a new chart
NO. 

Synthesizing two new metrics [node.cpu.request.percentage, node.memory.request.percentage] using metrics generation processor. To synthesize metrics on NR's side we'd either need to create an ingest scraper or do the synthesizing via NRQL. The issue with synthesizing metrics in NRQL is that the queries needed to aggregate metrics in particular are fairly gratuitous. For example.... 
``` SQL
FROM(
  FROM(
    FROM Metric 
    SELECT latest(kube_pod_container_resource_requests) AS containerRequest
    WHERE resource = 'cpu' AND container_phase != 'terminated'
    FACET k8s.container.name, k8s.pod.name, k8s.node.name, k8s.namespace.name LIMIT MAX
  )
  SELECT sum(containerRequest) AS totalRequests FACET k8s.node.name LIMIT MAX
) 
JOIN (
  FROM Metric
  SELECT latest(kube_node_status_allocatable) AS totalAllocatable
  WHERE resource = 'cpu' 
  FACET k8s.node.name LIMIT MAX
) ON k8s.node.name
SELECT sum(totalRequests)/ sum(totalAllocatable) * 100
```
This query is simply getting the average CPU utilization as a percentage across all of a customer's nodes.
Yet it has to be so monstrous because we have to be very careful when doing math on metrics to insure we are dividing the cpu usage and cpu capacity from the same nodes. 

From our experimentation with the metrics generation processor we've been able to reduce this query down to this. 
``` SQL
FROM Metric SELECT average(node.cpu.usage.percentage) 
```

Adds three new metrics to allow for easier/ more understandable and more accurate queries. 
* node.memory.usage.percentage
* k8s.pod.cpu_request_limit_ratio
* k8s.pod.memory_request_limit_ratio

The metrics generation processor, much like most processors, is stateless. It cannot do work across batches. 
The issue here is that the routing connector breaks apart batches based on the context. Since most of our routing is based on metrics or datapoints, none of our batches on the other end of those routers will have all of the info we need to aggregate our metrics with the generation processor. 

This is why metric generation is done at the begining of the pipeline. It allows us to create the new metrics before routing.

#### kubectl top node 
Screenshot 2025-10-30 at 9 59 07 AM

#### NRDB Query
Screenshot 2025-10-30 at 9 58 45 AM

#### Daemonset resource config
Screenshot 2025-10-30 at 10 00 38 AM

CPU request/limit = 50/75 = 0.667
Memory request/limit = 32/70 = 0.457

##### CPU
Screenshot 2025-10-30 at 9 59 47 AM

##### Memory
Screenshot 2025-10-30 at 10 00 10 AM

#### Checklist
- [x] Chart Version bumped

## 🚀 What's Changed
* Synthesizing three new metrics [k8s.pod.cpu_request_limit_ratio, k8s.pod.memory_request_limit_ratio, node.memory.usage.percentage] using metrics generation processor.


charts/nr-k8s-otel-collector/docs/metrics-full.md
```diff
@@ -1073,6 +1075,7 @@
 | KubeStateMetrics   | Prometheus Receiver   | scrape_samples_scraped                                                 | Gauge             | Number of samples the target exposed.                                                    |
 | KubeStateMetrics   | Prometheus Receiver   | scrape_series_added                                                    | Gauge             | Number of new series in the scrape.                                                      |
 | KubeStateMetrics   | Prometheus Receiver   | up                                                                     | Gauge             | 1 if the instance is healthy, 0 if the scrape failed.                                    |
+| Node               | HostMetric Receiver   | node.memory.usage.percentage                                           | Gauge             | Memory usage as a percentage of memory available.                                        |
```
ramkrishankumarN: Let's remove this metric altogether. It is not giving us the right node usage and could be misleading.

charts/nr-k8s-otel-collector/examples/k8s/rendered/daemonset-configmap.yaml
```diff
@@ -437,6 +446,69 @@ data:
...
+            operation: add
+
+      # The metricsgeneration processor does not support operating on a metric that it previously made in the same processor
+      metricsgeneration/calculate_percentage:
+        rules:
+          - name: node.memory.usage.percentage
+            type: calculate
+            metric1: k8s.node.memory.working_set_tmp
+            metric2: memory_capacity_tmp
+            operation: divide
```
ramkrishankumarN: Let's get rid of the `node.memory.usage.percentage`. This is giving us wrong data.

dbudziwojskiNR (approved):

PR #1937, opened 2025-09-22
[agent-control-deployment] make signing service url configurable

danielorihuela:
#### Is this a new chart

no

Makes signing service url configurable.

  - fixes #NR-462133

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent-control-deployment/Chart.yaml
+++ b/charts/agent-control-deployment/Chart.yaml
@@ -4,7 +4,7 @@ description: A Helm chart to install New Relic Agent Control on Kubernetes
 
 type: application
 
-version: 0.0.76
+version: 0.0.77
 appVersion: "0.49.0"
 
 dependencies:
--- a/charts/agent-control-deployment/Chart.yaml
+++ b/charts/agent-control-deployment/Chart.yaml
@@ -4,8 +4,8 @@ description: A Helm chart to install New Relic Agent Control on Kubernetes
 
 type: application
 
-version: 0.0.77
-appVersion: "0.49.0"
+version: 1.0.0
+appVersion: "1.0.0"
 
 dependencies:
   - name: common-library
--- a/charts/agent-control-deployment/tests/configmap_agentcontrol_config_test.yaml
+++ b/charts/agent-control-deployment/tests/configmap_agentcontrol_config_test.yaml
@@ -479,3 +479,69 @@ tests:
               enabled: true
               host: 0.0.0.0
               port: 51200
+
+  - it: signature validation can be disabled
+    set:
+      cluster: my-cluster
+      config:
+        fleet_control:
+          signature_validation:
+            enabled: false
+            public_key_server_url: https://my.public.key.server
+    asserts:
+      - equal:
+          path: data["local_config"]
+          value: |
+            agents: {}
+            fleet_control:
+              auth_config:
+                private_key_path: /etc/newrelic-agent-control/keys/from-secret.key
+                provider: local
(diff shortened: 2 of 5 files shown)


charts/agent-control-deployment/values.yaml
```diff
@@ -146,6 +146,12 @@ config:
     # -- Specify a fleet_id to automatically connect the Agent Control to an existing fleet.
     fleet_id: ""
 
+    signature_validation:
+      # -- Enables or disables signature validation for incoming requests.
+      enabled: true
+      # -- URL of the server from where to obtain the public key for signature validation.
+      public_key_server_url: ""
```
paologallinaharbur: why configurable and not computed automatically as in the case of opamp URL?

danielorihuela: You are right. In fact, this is what we did for the onhost part as @DavSanchez suggested.
I did the k8s part first, and this is what I understood we wanted. Let me rework it.

paologallinaharbur: They provided URLs for all environments (that are slightly different from the one I saw you used in the other PR)

charts/agent-control-deployment/tests/configmap_agentcontrol_config_test.yaml
```diff
@@ -75,6 +75,8 @@ tests:
                 token_url: https://system-identity-oauth.service.newrelic.com/oauth2/token
               endpoint: https://opamp.service.newrelic.com/v1/opamp
               fleet_id: abcefg
+              signature_validation:
```
paologallinaharbur, line 78: can we add a single test with the override?

paologallinaharbur, line 78: if not specified signature_validation is enabled=true by default, right?

charts/agent-control-deployment/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: A Helm chart to install New Relic Agent Control on Kubernetes
 
 type: application
 
-version: 0.0.76
+version: 0.0.77
```
paologallinaharbur: we should bump both versions to 1.0.0

paologallinaharbur: it is not related to this PR, but can you change the tag of the system identities from latest to [0.0.10](https://hub.docker.com/layers/newrelic/agent-control-system-identity-registration/0.0.10/images/sha256-b14de0e0da9c08560ca63f0fedd45700d2956a7c1bf357122f04021564480ed5)?

charts/agent-control-deployment/tests/configmap_agentcontrol_config_test.yaml
danielorihuela: I removed the possibility to make it configurable from the `values.yaml`. So the endpoint is always added.

charts/agent-control-deployment/Chart.yaml
danielorihuela: Discussed offline, there's nothing to change.

danielorihuela: Done

charts/agent-control-deployment/tests/configmap_agentcontrol_config_test.yaml
danielorihuela: We already have the test for US and EU. I removed the possibility to configure the url. Did we want it besides having the "default hardcoded" values?

PR #1949, opened 2025-09-30
fix: Enabled spans reporting to true in eBPF agent config

sumitsuthar:
#### Is this a new chart
NO

* Updated default config for eBPF agent.
* Enabled spans reporting to true.

### DIFF
--- a/charts/nr-ebpf-agent/values.yaml
+++ b/charts/nr-ebpf-agent/values.yaml
@@ -124,12 +124,12 @@ ebpfAgent:
   podSecurityContext: {}
   # -- Sets ebpfAgent pod containerSecurityContext. Overrides `containerSecurityContext` and `global.securityContext.container`
   containerSecurityContext: {}
-  # Sets the absolute path of the complete directory where the required linux headers are manually downloaded and placed for the eBPF agent to use. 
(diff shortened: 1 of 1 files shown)


charts/nri-bundle/Chart.yaml
```diff
@@ -68,7 +68,7 @@ dependencies:
 
   - name: nr-ebpf-agent
     version: 0.2.10
-    condition: newrelic-eapm-agent.enabled
+    condition: nr-ebpf-agent.enabled
```
dbudziwojskiNR: Hi! Changing this parameter name is considered a breaking change. As such, I'd recommend leaving the old config value (`newrelic-eapm-agent.enabled`) in the chart until the next major release. You can just add the new config as an OR clause: i.e. `logging.enabled,newrelic-logging.enabled`.

dbudziwojskiNR: You can also leave the docs updated with the new chart name/config.

burhan-nr: Since we have a freeze, we have decided not to go with these changes in this release. We'll revert it for now and make these changes in the next one.

PR #1992, opened 2025-11-05
[nr-k8s-otel-collector] Fetching Node Allocatable (CPU/Memory) from downward API

Philip-R-Beckwith:
Creating a couple extra metrics so that we can avoid the 5k problem with NRQL Query Math.

N/A

This PR grabs the daemonset collector's memory/ cpu limits. When these limit are not set, the downward API returns the node's allocatable values instead. We are "abusing" this fact to grab the node allocatable cpu/Memory values. We then use these values and the metrics generation processor to create metrics that show node usage vs node allocatable for CPU/Memory. Since these values in the downward api are only available on K8s 1.33.0 and up, this feature is put behind a config flag. This config flag also crashes the application when a user tries to set limits on the daemonset and enable these new metrics. Unit tests added. 

##### Kubectl top  node
######Note that values are rounded to nearest percentage
Screenshot 2025-11-04 at 4 18 21 PM

###### Values are rounding to two decimal places
Screenshot 2025-11-04 at 4 18 12 PM

##### Error message when setting limits with new feature enabled. 
Screenshot 2025-11-04 at 4 22 53 PM

#### Checklist
- [x] Chart Version bumped

## 🚀 What's Changed
* Adding generated node utilization metrics via the metricsgeneration processor by scraping node allocatable from resourceFieldRef.

### DIFF
--- a/charts/nr-k8s-otel-collector/templates/daemonset.yaml
+++ b/charts/nr-k8s-otel-collector/templates/daemonset.yaml
@@ -55,7 +55,7 @@ spec:
           env:
             {{- if .Values.enableNodeUtilizationMetrics }}
             # Resource Limits default to node allocatable when not set.
-            # We are abusing this fact to get node allocatable values into env vars.
+            # We are using this fact to get node allocatable values into env vars. when the resource limits are not set
             # See: https://kubernetes.io/docs/concepts/workloads/pods/downward-api/#downwardapi-resourceFieldRef
             - name: NODE_CPU_ALLOCATABLE
               valueFrom:


charts/nr-k8s-otel-collector/templates/daemonset.yaml
```diff
@@ -51,6 +53,22 @@ spec:
             {{- . | toYaml | nindent 12 }}
           {{- end }}
           env:
+            {{- if .Values.enableNodeUtilizationMetrics }}
+            # Resource Limits default to node allocatable when not set.
+            # We are abusing this fact to get node allocatable values into env vars.
```
ramkrishankumarN: Could we rephrase this to 
`We are using this fact to get node allocatable values into env vars.` when the resource limit defaults are not set

Philip-R-Beckwith: Will do.

PR #2020, opened 2025-11-26
feat: delete secret volumne for private key and add secret name

vjripoll:

Delete de volumne for the private key for opamp and and adding the name to be requested with the k8s api

  - fixes #

#### Checklist
- [ ] Chart Version bumped

## 🚀 What's Changed

### DIFF
--- a/charts/agent-control-deployment/Chart.yaml
+++ b/charts/agent-control-deployment/Chart.yaml
@@ -4,7 +4,7 @@ description: A Helm chart to install New Relic Agent Control on Kubernetes
 
 type: application
 
-version: 1.1.0
+version: 1.1.1
 appVersion: "1.6.1"
 
 dependencies:
--- a/charts/agent-control-deployment/README.md
+++ b/charts/agent-control-deployment/README.md
@@ -96,6 +96,13 @@ agents:
 			<td>"true"</td>
 			<td>enables or disables remote update from Fleet Control for the agent-control-cd chart</td>
 		</tr>
+        <tr>
+			<td>config.secretPrivateKeyName</td>
+			<td>string</td>
+			<td>`""`</td>
+			<td>Provide the secret name from where the private key should be loaded</td>
+		</tr>
+		<tr>
 		<tr>
 			<td>config.fleet_control.enabled</td>
 			<td>bool</td>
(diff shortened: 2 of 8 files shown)


charts/agent-control-bootstrap/templates/_helpers.tpl
```diff
@@ -21,5 +21,13 @@ overrides:
     {{- $config = mustMergeOverwrite $config (dict "config" (dict "cdRemoteUpdate" false "cdReleaseName" "")) -}}
   {{- end -}}
 
+  {{- $authSecret := (default dict .Values.config).authSecret | default dict -}}
+  {{- $sName := $authSecret.secretName | default "agent-control-auth" -}}
+  {{- $sKey  := $authSecret.secretKeyName | default "private_key" -}}
```
danielorihuela: Do we need the `| default ...`? Aren't these set in the chart values?

charts/agent-control-bootstrap/tests/ac_deployment_secret_test.yaml
```diff
@@ -4,18 +4,26 @@ chart:
 
 tests:
   - it: should render expected defaults
+    set:
+      config.authSecret.secretName: agent-control-auth
+      config.authSecret.secretKeyName: private_key
```
danielorihuela: If these are defaults, we shouldn't put them here, they should already be set from somewhere else. I think we can remove them from all tests except the new one that you created.

charts/agent-control-bootstrap/templates/_helpers.tpl
```diff
@@ -21,5 +21,13 @@ overrides:
...
+
+  {{- $secretObj := dict "secret_name" $sName "secret_key_name" $sKey -}}
+
+  {{- $config = mustMergeOverwrite $config (dict "config" (dict "auth_secret" $secretObj)) -}}
```
danielorihuela: Why do we need that here? The deployment chart already has the default values and the client can modify them with the `values.yaml` file, right? We are not changing the behaviour here in any way.

charts/agent-control-bootstrap/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: agent-control-bootstrap
 description: Bootstraps New Relic' Agent Control
 
 type: application
-version: 1.1.0
+version: 1.1.1
 # agent-control-deployment chart default version.
 appVersion: 1.1.0
```
danielorihuela: We should update the `appVersion`

charts/agent-control-bootstrap/tests/ac_deployment_secret_test.yaml
danielorihuela: Discussed offline

charts/agent-control-bootstrap/templates/_helpers.tpl
danielorihuela: Discussed offline

charts/agent-control-bootstrap/Chart.yaml
danielorihuela: Discussed offline

PR #2081, opened 2026-01-22
feat: enable nr-k8s-otel-collector self-observability via oltp export pattern

jamescripter:
https://new-relic.atlassian.net/browse/NR-510604

testing:
I did some e2e tests just installing nr-k8s-otel-collector in a local minikube cluster and seeing if it would report its own metrics via the new method. I was able to see [these metrics](https://opentelemetry.io/docs/collector/internal-telemetry/#lists-of-internal-metrics) after deploying these changes in a local cluster with `collectorObservability.enabled=true`

### DIFF
--- a/charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
+++ b/charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
@@ -771,6 +771,13 @@ data:
         {{- end }}
         metrics:
           readers:
+            {{- if (((.Values.receivers).collectorMetrics).enabled) }}
+            - pull:
+                exporter:
+                  prometheus:
+                    host: '0.0.0.0'
+                    port: 8888
+            {{- end }}
             {{- if .Values.collectorObservability.enabled }}
             - periodic:
                 interval: {{ .Values.collectorObservability.scrapeIntervalSeconds }}
--- a/charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
+++ b/charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
@@ -795,6 +795,13 @@ data:
         {{- end }}
         metrics:
           readers:
+            {{- if (((.Values.receivers).collectorMetrics).enabled) }}
+            - pull:
+                exporter:
+                  prometheus:
+                    host: '0.0.0.0'
+                    port: 8888
+            {{- end }}
             {{- if .Values.collectorObservability.enabled }}
             - periodic:
                 interval: {{ .Values.collectorObservability.scrapeIntervalSeconds }}


charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -788,10 +788,11 @@ data:
             pipelines: [metrics/nr_controlplane]
 
     service:
-      {{- if include "newrelic.common.verboseLog" . }}
       telemetry:
+        {{- if include "newrelic.common.verboseLog" . }}
```
jamescripter, line 792: had to narrow the scope of this condition to allow `collectorObservability` to work below

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -764,10 +764,11 @@ data:
             pipelines: [metrics/nr]
 
     service:
-      {{- if include "newrelic.common.verboseLog" . }}
       telemetry:
+        {{- if include "newrelic.common.verboseLog" . }}
```
jamescripter, line 768: had to narrow the scope of this condition to allow `collectorObservability` to work below

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -323,6 +325,14 @@ receivers:
     # @default -- `1m`
     scrapeInterval: 1m
 
+collectorObservability:
```
jamescripter, line 329: based on the wording of the story I assumed that I should initialize new config values for this feature and deprecate the old ones. let me know if I should reuse the existing `receivers.collectorMetrics...` keys instead though

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -801,7 +802,17 @@ data:
                     host: '0.0.0.0'
                     port: 8888
         {{- end }}
-      {{- end }}
+        {{- if .Values.collectorObservability.enabled }}
```
Philip-R-Beckwith: What happens if someone enables both `.Values.collectorObservability`
collectorObservability as well as `.Values.receivers.collectorMetrics` ?
Looks like that would lead to a malformed chart due to this object having to `metrics` fields.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -777,7 +778,17 @@ data:
                     host: '0.0.0.0'
                     port: 8888
         {{- end }}
-      {{- end }}
+        {{- if .Values.collectorObservability.enabled }}
```
Philip-R-Beckwith: What happens if someone enables both .Values.collectorObservability
collectorObservability as well as .Values.receivers.collectorMetrics ?
Looks like that would lead to a malformed chart due to this object having to metrics fields.

Philip-R-Beckwith (approved):

PR #2085, opened 2026-01-26
[newrelic-logging] Allow templatizing extra environment variables

jsubirat:

The `newrelic-logging` Helm chart allows passing in extra environment variables for the Fluent Bit containers via `fluentBit.extraEnv` or `fluentBit.additonalEnvVariables`. However, these variables can't currently depend on other values passed via `values.yaml`. With this PR, we can now make these environment variable values depend on other values, like this:

```
      someConfigurationKey: "someConfigurationValue"
      fluentBit:
        extraEnv:
        - name: EXTRA_ENV_VAR_WITH_TEMPLATE
          value: "Reads from someConfigurationKey: {{ .Values.someConfigurationKey }}"
```

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/.github/workflows/lint_test_charts.yaml
+++ b/.github/workflows/lint_test_charts.yaml
@@ -28,7 +28,7 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: actions/setup-python@v5
+      - uses: actions/setup-python@v6
       - uses: helm/chart-testing-action@v2.8.0
       - uses: azure/setup-helm@v4
         with:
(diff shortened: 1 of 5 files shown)


charts/newrelic-logging/templates/_helpers.tpl
```diff
@@ -277,9 +277,9 @@ If additionalEnvVariables is set, renames to extraEnv. Returns extraEnv.
 {{- define "newrelic-logging.extraEnv" -}}
 {{- if .Values.fluentBit }}
   {{- if .Values.fluentBit.additionalEnvVariables }}
-    {{- toYaml .Values.fluentBit.additionalEnvVariables -}}
+    {{- tpl (toYaml .Values.fluentBit.additionalEnvVariables) . -}}
```
maya-jha, line 280: what does `- tpl` do?

jsubirat: It templates (Helm resolves variable values ) the resulting string returned by `toYaml`

PR #2086, opened 2026-01-27
Gaurab/atp final helm chart

gmanandhar-nr:
# Add Adaptive Telemetry Processor (ATP) Support to nr-k8s-otel-collector

## Overview
This PR introduces support for the Adaptive Telemetry Processor (ATP) in the `nr-k8s-otel-collector` Helm chart. ATP provides intelligent filtering of process-level metrics based on dynamic thresholds, anomaly detection, and multi-metric composite evaluation, reducing data volume and cost while ensuring critical process metrics are captured.

## Key Features
- **Opt-in ATP Support**: Disabled by default to maintain backward compatibility
- **Intelligent Process Metrics Filtering**: Dynamic thresholds that adapt to workload patterns
- **Dual Example Manifests**: Separate rendered examples for standard k8s and k8s-with-atp deployments

### 1. ATP Configuration Flag
- Added `atp.enabled` flag to `values.yaml` (default: `false`)
- Comprehensive documentation explaining ATP functionality and requirements
- Backward compatible - no impact on existing deployments

#### `templates/daemonset-configmap.yaml`
- **Process Metrics Collection**: Full process metrics (CPU, memory, disk, network, threads, file descriptors, etc.) enabled conditionally when ATP is active
- **ATP Processors**: 
  - `attributes/atp_identifier`: Tags metrics for ATP processing
  - `adaptivetelemetryprocessor`: Applies intelligent filtering with configurable thresholds
- **Routing Rules**: Process/processes metrics routed through ATP pipeline when enabled
- **Backward Compatibility**: Commented process metrics block preserved when ATP is disabled
- **Conditional Metric Transforms**: Process metric tagging only when ATP is enabled

#### `templates/daemonset.yaml`
- **Persistent Storage**: Added `nrdot-data-storage` hostPath volume conditionally
  - Mounted at `/var/lib/nrdot-collector`
  - Type: `DirectoryOrCreate`
  - Persists ATP state (thresholds, history) across pod restarts

#### Code Documentation
- Added comprehensive inline comments at every ATP conditional
- Explains why each conditional exists and what happens in each branch
- Helps future maintainers understand ATP integration points

#### New Directory Structure
```
examples/
├── k8s/                      # Standard deployment (ATP disabled)
│   ├── values.yaml -> ../../values.yaml (symlink)
│   └── rendered/             # Standard manifests
└── k8s-with-atp/            # ATP-enabled deployment
    ├── values.yaml           # Minimal override (atp.enabled: true)
    └── rendered/             # ATP-enabled manifests
```

#### Automatic Generation
- Existing `make generate-examples` command automatically processes both directories
- No Makefile changes required - generic discovery pattern
- Helm's built-in value merging handles override application

### DIFF
--- a/charts/pipeline-control-gateway/Chart.yaml
+++ b/charts/pipeline-control-gateway/Chart.yaml
@@ -1,12 +1,12 @@
 apiVersion: v2
 name: pipeline-control-gateway
 type: application
-version: 2.0.0
+version: 2.0.1
 dependencies:
   - name: common-library
     version: 1.4.0
     repository: https://helm-charts.newrelic.com
-appVersion: "2.0.0"
+appVersion: "2.0.1"
 maintainers:
   - name: pipeline-control
     url: https://github.com/orgs/newrelic/teams/pcon/members
--- a/charts/pipeline-control-gateway/values.yaml
+++ b/charts/pipeline-control-gateway/values.yaml
@@ -22,7 +22,7 @@ image:
   # -- The pull policy is defaulted to IfNotPresent, which skips pulling an image if it already exists. If pullPolicy is defined without a specific value, it is also set to Always.
   pullPolicy: IfNotPresent
   # --  Overrides the image tag whose default is the chart appVersion.
-  tag: "2.0.0"
+  tag: "2.0.1"
 
 # -- Name of the Kubernetes cluster monitored. Mandatory. Can be configured also with `global.cluster`
 cluster: ""


charts/nr-k8s-otel-collector/README.md
```diff
@@ -123,6 +123,8 @@ to export data to this connector which can then be connected to the New Relic ma
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | affinity | object | `{}` | Sets all pods' affinities. Can be configured also with `global.affinity` |
+| atp | object | `{"enabled":false,"image":{"pullPolicy":"IfNotPresent","registry":"","repository":"newrelic/nrdot-collector","tag":"1.8.0"}}` | Enable Adaptive Telemetry Processor (ATP) for intelligent process metrics filtering. When disabled (default), ATP processors are not included in the pipeline. When enabled, activates ATP with opinionated process metrics collection. Note: When enabled, automatically uses ATP-capable collector image. |
+| atp.image | object | `{"pullPolicy":"IfNotPresent","registry":"","repository":"newrelic/nrdot-collector","tag":"1.8.0"}` | Image configuration for ATP-enabled collector (only used when atp.enabled=true) TODO: Once a version of nrdot-collector with ATP is published, update the tag |
```
Philip-R-Beckwith: If this isn't ready for production, we should not merge it. 
Merging any PR in this repo is releasing it to production.

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -2,6 +2,20 @@
...
+# Note: When enabled, automatically uses ATP-capable collector image.
+atp:
+  enabled: false
+  # -- Image configuration for ATP-enabled collector (only used when atp.enabled=true)
+  # TODO: Once a version of nrdot-collector with ATP is published, update the tag
+  image:
+    registry: ""
+    repository: newrelic/nrdot-collector
+    tag: "1.8.0"
+    pullPolicy: IfNotPresent
```
Philip-R-Beckwith: If this is going to be the same image as default, we don't need this block.

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -2,6 +2,20 @@
...
 # Declare variables to be passed into your templates.
 
+# -- Enable Adaptive Telemetry Processor (ATP) for intelligent process metrics filtering.
+# When disabled (default), ATP processors are not included in the pipeline.
+# When enabled, activates ATP with opinionated process metrics collection.
+# Note: When enabled, automatically uses ATP-capable collector image.
+atp:
+  enabled: false
+  # -- Image configuration for ATP-enabled collector (only used when atp.enabled=true)
+  # TODO: Once a version of nrdot-collector with ATP is published, update the tag
```
Philip-R-Beckwith: Put this below required configs. 

"cluster" and "licensekey"

charts/nr-k8s-otel-collector/README.md
```diff
@@ -123,6 +123,8 @@ to export data to this connector which can then be connected to the New Relic ma
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | affinity | object | `{}` | Sets all pods' affinities. Can be configured also with `global.affinity` |
+| atp | object | `{"enabled":false,"image":{"pullPolicy":"IfNotPresent","registry":"","repository":"newrelic/nrdot-collector","tag":"1.8.0"}}` | Enable Adaptive Telemetry Processor (ATP) for intelligent process metrics filtering. When disabled (default), ATP processors are not included in the pipeline. When enabled, activates ATP with opinionated process metrics collection. Note: When enabled, automatically uses ATP-capable collector image. |
```
Philip-R-Beckwith: Since the image will not be different when enabling ATP, Fix Readme to reflect the change in direction.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -756,6 +813,15 @@ data:
           - context: metric
             condition: instrumentation_scope.name == "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/hostmetricsreceiver/internal/scraper/pagingscraper"
             pipelines: [metrics/nr]
+          {{- if .Values.atp.enabled }}
```
Philip-R-Beckwith: Doesn't seem like we need an if statement here. 
We may be adding unneeded cognitive load.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -426,6 +482,7 @@ data:
               value_actions:
               - value: 'false'
                 new_value: 'true'
+          {{- end }}
```
Philip-R-Beckwith, line 479: Looks like this if statement is swallowing some metrics that should exist even when it's not enabled.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -207,6 +229,17 @@ data:
           - node
           - namespace
 
+      {{- if .Values.atp.enabled }}
```
Philip-R-Beckwith: Do we need this if statement?

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -54,7 +54,28 @@ data:
...
+                enabled: true
+              process.memory.usage:
+                enabled: true
+              process.memory.virtual:
+                enabled: true
+            mute_process_name_error: true
+            mute_process_exe_error: true
+            mute_process_io_error: true
+            mute_process_user_error: true
+          {{- else }}
```
Philip-R-Beckwith: Let's not have an else statement that just wraps commented out code. 
Just remove these. 
If a customer wants support for enabling these, we can add a config flag. 
We don't expect customers to be looking though the chart to uncomment blocks of code.

charts/nr-k8s-otel-collector/templates/_images.tpl
```diff
@@ -1,20 +1,26 @@
 {{- /*
 Temporary image helper to transition from old image config structure to new structure.
+Supports ATP-specific image when atp.enabled=true.
 */ -}}
 {{- define "nrKubernetesOtel.images.collector.image" }}
 {{- if or (.Values.image).repository (.Values.image).tag }}
    {{- (.Values.image).repository | default .Values.images.collector.repository }}:{{- (.Values.image).tag | default .Chart.AppVersion }}
+{{- else if .Values.atp.enabled }}
```
Philip-R-Beckwith: I don't think we need changes in this file.

charts/nr-k8s-otel-collector/README.md
gmanandhar-nr: decided that we will approve this **provisionally**, and not merge it yet. Once we have a final image, we will update the image, and then merge the PR.

charts/nr-k8s-otel-collector/values.yaml
gmanandhar-nr: we would need a separate image. explained in the slack thread [here](https://newrelic.slack.com/archives/C04S2FT1LU9/p1770026915958569?thread_ts=1769547078.216119&cid=C04S2FT1LU9)

charts/nr-k8s-otel-collector/values.yaml
gmanandhar-nr: done

charts/nr-k8s-otel-collector/README.md
gmanandhar-nr: As discussed [here](https://newrelic.slack.com/archives/C04S2FT1LU9/p1770026915958569?thread_ts=1769547078.216119&cid=C04S2FT1LU9), the direction is that we will have a separate image.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
gmanandhar-nr: agreed. made these changes in the latest commit. will push soon

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
gmanandhar-nr: although, we could technically define it, but not use it, we think it might be cleaner to not "define" it in the "processors:" block altogether, if `atp.enabled=false`. Would help make it less confusing and and makes it clear what's active vs inactive.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
gmanandhar-nr: Removed the conditional. This isn't atp specific. Since the process scrapers are only enabled when atp.enabled=true, these routing rules will match nothing when ATP is disabled. The conditional control at the scraper and processor levels is sufficient.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
gmanandhar-nr: The [previous code](https://github.com/newrelic/helm-charts/blob/master/charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml#L420-L428) was such that if low data mode was ON, it would mark only few process metrics such as process.cpu.utilization, process.disk.io, process.memory.usage and process.memory.virtual as `low.data.mode` and only send these to NR, and drop rest of the process metrics. However, with ATP, we want to send all the `process` and `processes` level metrics to ATP processor, as ATP should be filtering them out and decide whether or not to send them to NR.
If the atp is disabled, the existing code would still function the same way! We can verify this in the [rendered daemonset-configmap](https://github.com/newrelic/helm-charts/pull/2086/changes#diff-c2d9770e6db76467be9f222f36127f9e63532f4119f17f5de2270834eadaeb01) in this PR, when the atp is disabled. There are no changes around this block.

charts/nr-k8s-otel-collector/templates/_images.tpl
gmanandhar-nr: As discussed [here](https://newrelic.slack.com/archives/C04S2FT1LU9/p1770026915958569?thread_ts=1769547078.216119&cid=C04S2FT1LU9), the direction is that we will have a separate image, and therefore, this change would be required.

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
Philip-R-Beckwith: Why is this block of code inside an if statement when it was not inside an if statement before? 

          - include: process\.(cpu\.utilization|disk\.io|memory\.(usage|virtual))
            action: update
            match_type: regexp
              value_actions:
              - value: 'false'
                new_value: 'true'

Philip-R-Beckwith: This comment is still outstanding.

gmanandhar-nr: Hi @Philip-R-Beckwith , have posted the clarification for this yesterday above:

> The above block of code is inside the 'else' statement (not 'if' statement). So that, if atp is disabled ('else' condition kicks in), the above block of code would be included in the final config, the same way as it is today.
> 
> But, if atp is enabled ('if' condition kicks in), instead of simply process\.(cpu\.utilization|disk\.io|memory\.(usage|virtual)), we want to include ALL process\..* and processes\..* to the ATP block, instead of simply dropping it
> 
> To see what would be the final config when the 'if' condition is true and false, we can see the rendered manifests.
> When ATP is disabled, [this](https://github.com/newrelic/helm-charts/pull/2086/changes#diff-c2d9770e6db76467be9f222f36127f9e63532f4119f17f5de2270834eadaeb01R393-R401) is the relevant block in the rendered config (there is no change in the above block!)
> When ATP is enabled, [this](https://github.com/newrelic/helm-charts/pull/2086/changes#diff-3296fa5963a8ae87692ca71c2ea8e276d350d19b60037479b275f495bd23014dR417-R435) is the relevant block in the rendered config

So, I am not sure what else needs to be done. Could you please check if the above comments address your doubts or pls let me know how exactly we can resolve this? thank you!

Philip-R-Beckwith (approved):

PR #2035, opened 2025-12-11
[nr-ebpf-agent]: cos and bottle rocket support

kkhandelwal-nr:
#### Is this a new chart
No
#### What this PR does / why we need it:
- Container-Optimized OS (COS) Support: Added full support for Google's Container-Optimized OS with automatic kernel header installation
- Bottlerocket OS Support: Added support for AWS Bottlerocket with automated kernel header retrieval
- Enhanced Data Filtering: Introduced granular filtering capabilities for APM and network metrics data
  - `apmDataFilters`: Filter APM data by pod labels, entity names, or APM agent presence
  - `networkMetricsDataFilter`: Filter network metrics by pod labels or entity names
  - `allDataFilters`: Global filters affecting all data types
- Refactored Init Container: Moved kernel header installation logic from inline bash to ConfigMap-based script for better maintainability
- Configuration Reorganization: Restructured values.yaml with clearer naming and logical grouping
  - `tcpStatsReporting` → networkMetricsReporting
  - `dropDataNewRelic` → allDataFilters.dropNewRelicBundle
  - `dropDataForNamespaces` → allDataFilters.dropNamespaces
  - `dropDataServiceNameRegex` → allDataFilters.dropServiceNameRegex
  - `allowServiceNameRegex` → allDataFilters.keepServiceNameRegex
- Added installer-script ConfigMap for init container logic
#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version bumped
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[mychartname]`)

PR #2091, opened 2026-01-28
[nr-ebpf-agent] Fix precedence model: local > global > default for all values

dpacheconr:
## Overview

Add support for cascading registry, pullSecrets, and pullPolicy from global settings to nr-ebpf-agent.
This PR also adds comprehensive test coverage validating global value inheritance for all applicable values from the nri-bundle global values contract. The chart uses the common-library pattern; this work adds explicit validation tests and completes all applicable global value support.

### Phase 1: Global.Images.Registry Support ✅
- Added a separate `registry` field to both image configs (`ebpfAgent.image.registry` and `ebpfAgent.kernelHeaderInstaller.image.registry`)
- Stripped registry prefix from `repository` defaults — now path-only (e.g., `newrelic/newrelic-ebpf-agent`)
- Added image helpers to `_helpers.tpl` for both images following the same pattern as `newrelic-pixie` and `common-library`
- Updated daemonset template to use new image helpers

### Phase 2: Global.Images.PullSecrets Support ✅
- Added chart-level `imagePullSecrets` configuration to values.yaml
- Updated daemonset (`nr-ebpf-agent-daemonset.yaml`)
- Precedence: `imagePullSecrets` > `global.images.pullSecrets` > default (nothing)
- Renders `imagePullSecrets` only when configured

### Phase 3: Global.Images.PullPolicy Support ✅
- Added chart-level `pullPolicy` configuration to values.yaml for both container images
- Added pullPolicy helpers to `_helpers.tpl` for both container images
- Updated daemonset to use pullPolicy helpers
- Precedence: chart-level `pullPolicy` > `global.images.pullPolicy` > default (`IfNotPresent`)

## Configuration Hierarchy (Using nri-bundle Global Contract)

**For Image Registry** — consistent with `newrelic-pixie` and `common-library` pattern:

| Priority | Source | Example |
|---|---|---|
| 1 (highest) | `ebpfAgent.image.registry` (local) | `my.registry.io` |
| 2 | `global.images.registry` | `my.registry.io` |
| 3 (default) | `docker.io` | — |

The `registry` and `repository` fields are separate. `repository` holds only the image path (e.g., `newrelic/newrelic-ebpf-agent`). This allows `global.images.registry` to work regardless of whether a custom repository is set.

**For ImagePullSecrets:**

| Priority | Source |
|---|---|
| 1 (highest) | `imagePullSecrets` (chart-level) |
| 2 | `global.images.pullSecrets` |
| 3 | None |

**For ImagePullPolicy:**

| Priority | Source |
|---|---|
| 1 (highest) | `ebpfAgent.image.pullPolicy` / `ebpfAgent.kernelHeaderInstaller.image.pullPolicy` |
| 2 | `global.images.pullPolicy` |
| 3 | `IfNotPresent` |


dpacheconr:
Rebased onto master, updated hardcoded image tags (1.2.0 → 1.3.0), added YAML anchors/aliases to all new tests, and strengthened custom secret assertions to verify both secretKeyRef.name and secretKeyRef.key.

charts/nr-ebpf-agent/values.yaml
```diff
@@ -174,10 +174,19 @@ ebpfAgent:
...
+    pullPolicy: ""
+  # -- Configuration for the kernel header installer init container
+  kernelHeaderInstaller:
+    image:
+      # -- Kernel header installer image repository. Can be overridden with `global.images.registry`
+      repository: docker.io/newrelic/newrelic-ebpf-agent
+      # -- The tag of the kernel header installer image. Defaults to agent-base-image-latest
+      tag: agent-base-image-latest
+      # -- Image pull policy for kernel header installer. Overrides `global.images.pullPolicy`
+      pullPolicy: ""
```
kkhandelwal-nr, line 193: Any specific reason to introduce this config - `kernelHeaderInstaller` ?

dpacheconr: That’s the initcontainer, it’s named after the pre existing container name

charts/nr-ebpf-agent/templates/nr-ebpf-agent-daemonset.yaml
```diff
@@ -69,7 +69,7 @@ spec:
           - name: KUBERNETES_CLUSTER_DOMAIN
             value: "cluster.local"
           - name: NEW_RELIC_LOG_LEVEL
-            value: "{{ .Values.logLevel }}"
+            value: "{{ if and .Values.global .Values.global.verboseLog }}DEBUG{{ else }}{{ .Values.logLevel }}{{ end }}"
```
burhan-nr: @dpacheconr Why does verboseLog have a precendance of global > local?

charts/nr-ebpf-agent/values.yaml
dpacheconr: we can rename it to initcontainer if you prefer

charts/nr-ebpf-agent/templates/nr-ebpf-agent-daemonset.yaml
dpacheconr: good catch, fixed with latest commit

charts/nr-ebpf-agent/values.yaml
kkhandelwal-nr: @dpacheconr The issue is not with the naming here. We don't intend to make this user configurable. 
Here is an older discussion thread regarding the same - https://github.com/newrelic/helm-charts/pull/2035#pullrequestreview-3569854588

charts/nr-ebpf-agent/templates/nr-ebpf-agent-daemonset.yaml
```diff
@@ -42,8 +42,8 @@ spec:
       {{- end }}
       initContainers:
       - name: kernel-header-installer
-        image: {{ include "nr-ebpf-agent.initContainerImage" . }}
-        imagePullPolicy: {{ .Values.ebpfAgent.image.pullPolicy }}
+        image: {{ include "nr-ebpf-agent.kernelHeaderInstaller.image" . }}
```
burhan-nr: This change will overwrite the check that we have in place for handling RHCOS support. Please add the call to the `kernelHeaderInstaller.image` method as the fallback to this check [here](https://github.com/dpacheconr/helm-charts/blob/1eae0370914ef6d669c72c5aae1f8f9645e2d4f7/charts/nr-ebpf-agent/templates/_helpers.tpl#L143).

burhan-nr:
@dpacheconr I've refactored the implementation for kernel header installer image. Replaced the hardcoded fallback in `initContainerImage` with a call to `kernelHeaderInstaller.image`.  The daemonset now uses `initContainerImage`

dpacheconr:
@burhan-nr some of your changes break functionality we aiming to implement to support global values 
I pushed some changes to restore it but keep in line with your changes

burhan-nr:
> @burhan-nr some of your changes break functionality we aiming to implement to support global values I pushed some changes to restore it but keep in line with your changes

Ah yes. Only meant to update the initcontainer image call from the daemonset.

burhan-nr (approved):
LGTM

kkhandelwal-nr (approved):
LGTM

PR #2133, opened 2026-02-26
[nr-k8s-otel-collector] make collector observability resource attributes configurable

achang209:
Exposes `service::telemetry::resource` as a configurable Helm value under `collectorObservability.resource`, allowing users to enrich internal `otelcol_*` metrics with custom resource attributes.

Without this, internal collector metrics arrive in New Relic with no cluster or environment context, making it impossible to distinguish which cluster or environment a collector's internal metrics originate from in multi-cluster deployments. The `resource/newrelic` processor correctly enriches pipeline metrics with `k8s.cluster.name`, but `service::telemetry` exports directly via OTLP and bypasses all pipeline processors, so it never receives that enrichment.

This PR adds a `resource` block under `service::telemetry` in both the daemonset and deployment configmaps. `k8s.cluster.name` is automatically populated from the existing `newrelic.common.cluster` helper, following the same pattern already used in the `resource/newrelic` processor. Additional arbitrary attributes can be passed via `collectorObservability.resource` for things like `env`.

This also fixes a minor inconsistency where `values.yaml` documented the scrape interval field as `scrapeIntervalMs` but the configmap template referenced it as `scrapeIntervalSeconds`.
  - fixes #

- The `resource` block is placed at `service::telemetry::resource` per the [OTel collector internal telemetry docs](https://opentelemetry.io/docs/collector/internal-telemetry/#configure-resource-attributes), which applies resource attributes across all internal telemetry signals (metrics, logs, and traces)
- The `resource` block is only rendered when `collectorObservability.enabled` is `true`, so there is no impact to users who have not enabled collector observability
- `k8s.cluster.name` is always set automatically and does not need to be specified by the user
- The change is identical in both `daemonset-configmap.yaml` and `deployment-configmap.yaml`
- The `scrapeIntervalMs` → `scrapeIntervalSeconds` rename in `values.yaml` is a documentation fix only — the configmap template already used `scrapeIntervalSeconds` so behavior is unchanged

#### Checklist
- [x] Chart Version bumped

## 🚀 What's Changed
* Added `collectorObservability.resource` to allow users to enrich internal collector metrics with custom resource attributes. `k8s.cluster.name` is automatically populated from the chart's cluster value, enabling multi-cluster observability out of the box. Additional attributes such as `env` can be passed via `collectorObservability.resource`.
* Fixed a documentation inconsistency where the scrape interval field was referenced as `scrapeIntervalMs` in `values.yaml` but `scrapeIntervalSeconds` in the configmap template.

### DIFF
--- a/charts/nr-k8s-otel-collector/Chart.yaml
+++ b/charts/nr-k8s-otel-collector/Chart.yaml
@@ -17,7 +17,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.10.8
+version: 0.10.9
 
 
 dependencies:
--- a/charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
+++ b/charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
@@ -827,6 +827,11 @@ data:
                     endpoint: {{ include "nrKubernetesOtel.endpoint" . }}
                     headers:
                       api-key: ${env:NR_LICENSE_KEY}
+        resource:
+          k8s.cluster.name: {{ include "newrelic.common.cluster" . }}
+          {{- range $k, $v := .Values.collectorObservability.resource }}
+          {{ $k }}: {{ $v }}
+          {{- end }}
             {{- end }}
       pipelines:
         {{- include "nrKubernetesOtel.daemonset.configMap.extraConfig.pipelines" . | nindent 8 }}
--- a/charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
+++ b/charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
@@ -804,6 +804,11 @@ data:
                     endpoint: {{ include "nrKubernetesOtel.endpoint" . }}
                     headers:
                       api-key: ${env:NR_LICENSE_KEY}
+        resource:
+          k8s.cluster.name: {{ include "newrelic.common.cluster" . }}
+          {{- range $k, $v := .Values.collectorObservability.resource }}
+          {{ $k }}: {{ $v }}
+          {{- end }}
             {{- end }}
       pipelines:
         {{- include "nrKubernetesOtel.deployment.configMap.extraConfig.pipelines" . | nindent 8 }}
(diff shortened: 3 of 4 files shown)


achang209:
@dbudziwojskiNR @Philip-R-Beckwith would love feedback on this at your earliest convenience. Thanks!

charts/nr-k8s-otel-collector/templates/daemonset-configmap.yaml
```diff
@@ -827,6 +827,11 @@ data:
                     endpoint: {{ include "nrKubernetesOtel.endpoint" . }}
                     headers:
                       api-key: ${env:NR_LICENSE_KEY}
+        resource:
```
dbudziwojskiNR, line 830: Could we stamp the cluster name through a traditional processor in a pipeline instead? This leaves the door open to stamping other attributes from processors like `resource_detection`

charts/nr-k8s-otel-collector/templates/deployment-configmap.yaml
```diff
@@ -804,6 +804,11 @@ data:
                     endpoint: {{ include "nrKubernetesOtel.endpoint" . }}
                     headers:
                       api-key: ${env:NR_LICENSE_KEY}
+        resource:
```
dbudziwojskiNR, line 807: Same thing here.

charts/nr-k8s-otel-collector/Chart.yaml
```diff
@@ -17,7 +17,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.10.8
+version: 0.10.9
```
dbudziwojskiNR, line 20: You don't need to bump this version anymore, our CICD process will do so on release.

charts/nr-k8s-otel-collector/values.yaml
```diff
@@ -332,9 +332,18 @@ collectorObservability:
   # -- (bool) Specifies whether the collector reports its own metrics
   # @default -- `false`
   enabled: false
-  # -- Specifies the interval at which the collector reports its metrics (in milliseconds)
-  # @default -- `60000`
-  scrapeIntervalMs: 60000
+  # -- Specifies the interval at which the collector reports its metrics (in seconds)
+  # @default -- `60`
+  scrapeIntervalSeconds: 60
```
dbudziwojskiNR, line 337: This value should probably remain as `scrapeIntervalMs`, and instead we should fix this on the deployment and daemonset since `periodic.interval` accepts ms, not seconds. Alternatively, we could try and normalize this to seconds. Open to ideas.

dbudziwojskiNR (changes requested):

PR #2156, opened 2026-03-11
feat: add new role with less permisions for AC

vjripoll:
## Description

  This PR adds support for minimal RBAC permissions when installing Agent Control with Flux, as an alternative to the current cluster-admin approach.

  ## Problem

  Currently, the Agent Control Bootstrap chart requires `cluster-admin` permissions during installation to deploy Flux CD and Agent Control. While this works, it grants excessive privileges that may not align with security best practices in
  production environments.

#### Checklist
- [ ] Chart Version bumped

## 🚀 What's Changed

### DIFF
--- a/charts/nri-bundle/Chart.lock
+++ b/charts/nri-bundle/Chart.lock
@@ -31,12 +31,12 @@ dependencies:
   version: 1.3.1
 - name: k8s-agents-operator
   repository: https://newrelic.github.io/k8s-agents-operator
-  version: 0.38.2
+  version: 0.38.3
 - name: pixie-operator-chart
   repository: https://pixie-operator-charts.storage.googleapis.com
   version: 0.1.7
 - name: newrelic-infra-operator
   repository: https://newrelic.github.io/newrelic-infra-operator
-  version: 2.18.7
-digest: sha256:976ce3ab217a93bc8dc7b605d8f1f81a20b48610e674e0ca9578179106489b32
-generated: "2026-03-17T11:40:38.731342905Z"
+  version: 2.19.0
+digest: sha256:9d416061ae09f36267a2854f078c322e3e4fc577bb07bd859db2793c42fd4e5d
+generated: "2026-03-24T08:27:10.860553449Z"
--- a/charts/nri-bundle/Chart.yaml
+++ b/charts/nri-bundle/Chart.yaml
@@ -18,7 +18,7 @@ sources:
   - https://github.com/newrelic/newrelic-infra-operator/tree/master/charts/newrelic-infra-operator
   - https://github.com/newrelic/k8s-agents-operator/tree/master/charts/k8s-agents-operator
 
-version: 6.0.42
+version: 6.0.43
 
 dependencies:
   - name: newrelic-infrastructure
@@ -74,7 +74,7 @@ dependencies:
   - name: k8s-agents-operator
     repository: https://newrelic.github.io/k8s-agents-operator
     condition: k8s-agents-operator.enabled
-    version: 0.38.2
+    version: 0.38.3
 
   # Keep the version of pixie-operator-chart in sync with the CRD versions for
   # olm_crd.yaml and px.dev_viziers.yaml in
@@ -88,7 +88,7 @@ dependencies:
   - name: newrelic-infra-operator
     repository: https://newrelic.github.io/newrelic-infra-operator
     condition: newrelic-infra-operator.enabled
-    version: 2.18.7
+    version: 2.19.0
 
 maintainers:
   - name: Philip-R-Beckwith


charts/agent-control-bootstrap/templates/rbac-cd-minimal.yaml
```diff
@@ -0,0 +1,226 @@
...
+
+  # Services - Required for Flux notification webhook
+  - apiGroups: [""]
+    resources: [services]
+    verbs: [get, list, create, update, patch, delete]
+
+  # Network Policies - Uncomment if your Flux deployment requires NetworkPolicies
+  # - apiGroups: [networking.k8s.io]
+  #   resources: [networkpolicies]
+  #   verbs: [get, list, create, update, patch, delete]
```
paologallinaharbur: the user cannot uncomment this

vjripoll: 👌

charts/agent-control-bootstrap/templates/rbac-cd-minimal.yaml
```diff
@@ -0,0 +1,221 @@
...
+      - helmreleases/status
+      - helmcharts/status
+      - gitrepositories/status
+      - buckets/status
+      - ocirepositories/status
+      - kustomizations/status
+      - alerts/status
+      - providers/status
+      - receivers/status
+    verbs: [get, update, patch]
```
sigilioso: Do we need all of these? alerts, receivers, ...? Maybe we are adding all flux-related in order to avoid potential changes in the future.

vjripoll: yep currently are not needed but is our flux at least is not cluster-admin :D

charts/agent-control-bootstrap/templates/rbac-cd-minimal.yaml
```diff
@@ -0,0 +1,221 @@
+{{- if and .Values.agentControlCd.enabled .Values.rbac.useMinimalPermissions }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  annotations:
+    helm.sh/hook: post-install,post-upgrade
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+    helm.sh/hook-weight: "10"
```
paologallinaharbur: why we have hook annotations?

charts/agent-control-bootstrap/templates/rbac-cd-minimal.yaml
```diff
@@ -0,0 +1,221 @@
...
+  labels:
+    {{- include "newrelic.common.labels" . | nindent 4 }}
+    agent-control.newrelic.com/minimal-permissions: "true"
+  name: newrelic-ac-minimal
```
paologallinaharbur: I would avoid having "minimal" in the name 😄

charts/agent-control-bootstrap/values.yaml
```diff
@@ -104,3 +104,13 @@ agentControlCd:
...
+
+# -- RBAC minimal permissions configuration
+# When enabled, uses minimal required permissions instead of cluster-admin
+# @default -- See `values.yaml`
+rbac:
+  # -- Use minimal permissions instead of cluster-admin
+  # When true, installation jobs will use a restricted ClusterRole with only
+  # the permissions required for Flux and Agent Control to function
+  # @default -- false
+  useMinimalPermissions: false
```
paologallinaharbur: Isn't the agentControlCd chart still leveraging cluster admin?

paologallinaharbur: I wonder how it is working right now, since the flux chart is still being installed with admin, no?

charts/agent-control-bootstrap/templates/rbac-cd-minimal.yaml
```diff
@@ -0,0 +1,221 @@
...
+    verbs: [get, list, create, update, patch, delete]
+
+  # RBAC - Required to create roles for Flux controllers
+  - apiGroups: [rbac.authorization.k8s.io]
+    resources:
+      - clusterroles
+      - clusterrolebindings
+      - roles
+      - rolebindings
+    verbs: [get, list, create, update, patch, delete, bind, escalate]
```
paologallinaharbur: The fact that you have `escalate` isn't the very same of having cluster_admin? I think it is working becasue of that

charts/agent-control-bootstrap/templates/rbac-cd.yaml
```diff
@@ -9,18 +9,150 @@ metadata:
...
+    resources: [events]
+    verbs: [get, list, watch, create, patch]
+
+  # New Relic Custom Resources - Required for Agent Control
+  # Restricted to specific resources instead of wildcard for better security
+  - apiGroups: [newrelic.com]
+    resources:
+      - instrumentations
+      - instrumentations/status
+    verbs: [get, list, watch, create, update, patch, delete]
```
paologallinaharbur: I believe we should be able to remove these thanks to the escalation

charts/agent-control-bootstrap/templates/rbac-cd.yaml
```diff
@@ -9,18 +9,150 @@ metadata:
   labels:
     {{- include "newrelic.common.labels" . | nindent 4 }}
   name: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "flux-install-job") }}
+rules:
```
paologallinaharbur, line 12: We could remove move many of these permissions to a role and grant those only in the newrelic namespace.

(all non namespaces resources could stay here, for example namespaces, customresourcedefinitions, etc)

charts/agent-control-bootstrap/templates/rbac-cd.yaml
```diff
@@ -9,18 +9,150 @@ metadata:
...
+      - kustomizations/status
+      - alerts/status
+      - providers/status
+      - receivers/status
+    verbs: [get, update, patch]
+
+  # Secrets - Required for Flux and Agent Control to store credentials
+  - apiGroups: [""]
+    resources: [secrets]
+    verbs: [get, list, watch, create, update, patch, delete]
```
paologallinaharbur: could we check if we can remove the: get, list and watch?

I am not sure if they are needed to install/upgrade the flux chart

charts/agent-control-bootstrap/templates/rbac-cd.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if .Values.agentControlCd.enabled }}
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
+kind: ClusterRole
 metadata:
   annotations:
     helm.sh/hook: post-install,post-upgrade
```
paologallinaharbur, line 6: have you tested manually that it works both when installing and also when upgrading?

charts/agent-control-bootstrap/templates/rbac-cd.yaml
```diff
@@ -9,18 +9,174 @@ metadata:
...
+  # Reference: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#privilege-escalation-prevention-and-bootstrapping
+  # Note: 'delete' removed as installation job only creates/updates RBAC, never deletes.
+  # Uninstall job has separate permissions for deletion.
+  - apiGroups: [ rbac.authorization.k8s.io ]
+    resources:
+      - clusterroles
+      - clusterrolebindings
+    verbs: [ get, list, create, update, patch, bind, escalate ]
+
+---
```
paologallinaharbur, line 55: thanks for splitting them!

charts/agent-control-bootstrap/templates/rbac-cd.yaml
```diff
@@ -9,18 +9,174 @@ metadata:
...
+  - kind: ServiceAccount
+    name: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" .Release.Name "suffix" "install-job") }}
+    namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: Role
+  name: {{ include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" (include "newrelic.common.naming.fullname" .) "suffix" "flux-install-job") }}
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
paologallinaharbur, line 179: why is this a cluster role?

As far as I understand we could break this down as well and move it to be a role

paologallinaharbur: we need just to delete the newrelic-agent-control content and related resources outside, we should not need more than the installation job

PR #2170, opened 2026-03-20
[agent-control] Improve release automation [NR-475694]

gsanchezgavier:
This PR goal is to reduce friction on agent control char releases and allow both charts to be released at the same time.

- Moves the AC tag version from appVersion to the values so it get's automatically bumped by renovate
- ~Disable the ct test for the bootstrap, to remove the chart dependency lock. Testing is already performed with the e2e running in the same pr exclusive for agent-control. Current test fails on the upgrade scenario is expected~
- ~Once #2173  a single PR will be open by renovate ( we will still need to bump the bootstrap chart versions) but the rest of versions should be bumped automatically~
- ~A new version of both charts are being released witht this PR.~

### DIFF
--- a/charts/agent-control-deployment/values.yaml
+++ b/charts/agent-control-deployment/values.yaml
@@ -19,8 +19,7 @@ subAgentsNamespace: "newrelic"
 image:
   registry:
   repository: newrelic/newrelic-agent-control
-  # @default It defaults to `appVersion` in `Chart.yaml`.
-  tag: ""
+  tag: "1.11.0"
   pullPolicy: IfNotPresent
   # -- The secrets that are needed to pull images from a custom registry.
   pullSecrets: []
--- a/charts/agent-control-deployment/Chart.yaml
+++ b/charts/agent-control-deployment/Chart.yaml
@@ -5,7 +5,6 @@ description: A Helm chart to install New Relic Agent Control on Kubernetes
 type: application
 
 version: 1.6.1
-appVersion: "1.11.0"
 
 dependencies:
   - name: common-library


charts/agent-control-deployment/values.yaml
```diff
@@ -19,8 +19,7 @@ subAgentsNamespace: "newrelic"
 image:
   registry:
   repository: newrelic/newrelic-agent-control
-  # @default It defaults to `appVersion` in `Chart.yaml`.
-  tag: ""
+  tag: "1.11.0"
```
gsanchezgavier, line 22: having the tag defined here will make renovate automatically bump it

DavSanchez: I get it cannot do it from the `Chart.yaml`?

charts/agent-control-bootstrap/values.yaml
```diff
@@ -7,7 +7,7 @@ fullnameOverride: ""
 toolkitImage:
   registry:
   repository: newrelic/newrelic-agent-control-cli
-  tag: "1.10.0"
+  tag: "1.11.0"
```
danielorihuela: There was another task that proposed getting the toolkitImage tag from an annotation. If we are going to bump them from here, that's not needed anymore. Right?

I'm not against it, but a bit confused. I discussed the task yesterday with @sigilioso and the goal of doing an annotation was to have all version in the Chart.yaml to make them explicit and making the automatic bump easier.

By this PR I assume that:
* It's equally hard
* We don't care about having version in the `Chart.yaml`

charts/agent-control-deployment/values.yaml
gsanchezgavier: it would required different regex manager

PR #2184, opened 2026-03-25
eBPF Log Reporting Support Config

IshikaDawda:
- `APPLICATION_LOG_LEVEL_THRESHOLD` env
> Minimum log level to report (e.g., TRACE, DEBUG, INFO, WARN, ERROR). Default: `INFO`

- `REPORT_LOGS` env
> "true" — Always report application logs regardless of other settings.
"false" (Default) — Never report application logs.
"auto" — Reports application logs by default, but automatically suppresses them if an APM agent is detected with the entity.
**Precedence Rules -** If `REPORT_APM_DATA` is set to `false` or `auto`, it takes priority over the `REPORT_LOGS` configuration.

#### Deprecated following
- `log_reporting`


manoj-k2 (approved):

charts/nr-ebpf-agent/values.yaml
```diff
@@ -93,9 +92,11 @@ logDataFilters:
   # -- Application/entity log reporting filters
   applicationLogReporting:
     # -- Enable log reporting for entity monitored by eBPF Agent
-    enabled: false
+    enabled: "true"
```
sumitsuthar, line 95: if the accepted values are true and false then make it boolean

IshikaDawda: In recent changes, the ebpf agent is accepting the true/false values wrapped in string.

sumitsuthar: it should be boolean

IshikaDawda: As per discussion provided changes in [commit](https://github.com/newrelic/helm-charts/pull/2184/commits/4ab14e3f14b60c386e53071071250b99e55ef2b7)

kkhandelwal-nr (approved):
LGTM

PR #2225, opened 2026-04-28
[Common library] Region updates, added resolve function, region aware endpoint lookup. 

Philip-R-Beckwith:
Adds support for Japan and GOV regions in region TPL. 
Talking with DB we decided to make region names more uniform so shortened staging and local to "dev"
Which is why we are bumping the chart version to 2.0.0

Digging through this chart it looks like the ONLY thing that regions are used for is to set ULS, which makes sense. There really shouldn't be any changes to our charts other than collection endpoints.

So I scraped all of the endpoints in this repo, and created an "endpoint" template that uses the region template to automatically give the right endpoint. I flushed out the JP endpoints using Discovery and to decide the name of each endpoint. 

Added some helper functions for users who really want to know what region they are in for reasons other than endpoints. 

Also added some functions to make checking for value overrides in the root and global fields a lot easier. 
This is probably the number one duplicated code block that we have. So having a function we can call instead of re-writing this logic over and over is a huge boon to our helm charts. 

#### Checklist
- [x] Chart Version bumped

## 🚀 What's Changed

### DIFF
--- a/library/CHART-TEMPLATE/templates/example-cm-resolve.yaml
+++ b/library/CHART-TEMPLATE/templates/example-cm-resolve.yaml
@@ -7,7 +7,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
 data:
   {{/* Using lookup With Default  */}}
-  custom: {{ include "newrelic.common.resolve_or" (dict "ctx" . "key" "custom"  "default" "myDefaultData") -}}
+  custom: {{ include "newrelic.common.resolve" (dict "ctx" . "key" "custom"  "default" "myDefaultData") -}}
 
   {{/* Look up only will return nothing if value not found */}}
   proxy: {{ include "newrelic.common.resolve" (dict "ctx" . "key" "proxy") -}}
@@ -22,7 +22,7 @@ data:
   nested_three_deep: {{ include "newrelic.common.resolve" (dict "ctx" . "key" "nested.secondLevel.value") -}}
 
   {{/* Example of using lookup with default with a nested path */}}
-  nested_two_deep_with_default: {{ include "newrelic.common.resolve_or" (dict "ctx" . "key" "nested.value" "default" "defaultNestedValue") -}}
+  nested_two_deep_with_default: {{ include "newrelic.common.resolve" (dict "ctx" . "key" "nested.value" "default" "defaultNestedValue") -}}
 
   {{/* Example of looking up a nested path that doesn't exist... (Test that this doesn't cause errors) */}}
   nested_value_DNE: {{ include "newrelic.common.resolve" (dict "ctx" . "key" "nested.blah.value.DNE") -}}
--- a/library/CHART-TEMPLATE/templates/example-cm-resolve.yaml
+++ b/library/CHART-TEMPLATE/templates/example-cm-resolve.yaml
@@ -1,0 +1,28 @@
+# This is a dummy CM to test the resolve template
+# and give examples on how to use it
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "newrelic.common.naming.fullname" . }}-examples-cm-resolve
+  namespace: {{ .Release.Namespace }}
+data:
+  {{/* Using lookup With Default  */}}
+  custom: {{ include "newrelic.common.resolve_or" (dict "ctx" . "key" "custom"  "default" "myDefaultData") -}}
+
+  {{/* Look up only will return nothing if value not found */}}
+  proxy: {{ include "newrelic.common.resolve" (dict "ctx" . "key" "proxy") -}}
+
+  {{/* Example of using lookup with custom logic to set default */}}
+  endpoint: {{ include "newrelic.common.resolve" (dict "ctx" . "key" "endpoint") | default "inline-default" -}}
+
+  {{/* Example of using lookup test with a nested two-level path */}}
+  nested_two_deep: {{ include "newrelic.common.resolve" (dict "ctx" . "key" "nested.value") -}}
+
+  {{/* Example of using lookup test with a nested three-level path */}}
+  nested_three_deep: {{ include "newrelic.common.resolve" (dict "ctx" . "key" "nested.secondLevel.value") -}}
+
+  {{/* Example of using lookup with default with a nested path */}}
+  nested_two_deep_with_default: {{ include "newrelic.common.resolve_or" (dict "ctx" . "key" "nested.value" "default" "defaultNestedValue") -}}
+
+  {{/* Example of looking up a nested path that doesn't exist... (Test that this doesn't cause errors) */}}
+  nested_value_DNE: {{ include "newrelic.common.resolve" (dict "ctx" . "key" "nested.blah.value.DNE") -}}
--- a/library/CHART-TEMPLATE/tests/cl-images_registry_test.yaml
+++ b/library/CHART-TEMPLATE/tests/cl-images_registry_test.yaml
@@ -10,7 +10,7 @@ tests:
     asserts:
       - equal:
           path: spec.template.spec.containers[0].image
-          value: nginx:1.16.0
+          value: docker.io/nginx:1.16.0
 
   - it: sets registry if specified in values
     set:
@@ -50,7 +50,7 @@ tests:
     asserts:
       - equal:
           path: spec.template.spec.containers[0].image
-          value: nginx:1.16.0
+          value: docker.io/nginx:1.16.0
 
   - it: does not break if global.image=null
     set:
@@ -59,4 +59,4 @@ tests:
     asserts:
       - equal:
           path: spec.template.spec.containers[0].image
-          value: nginx:1.16.0
+          value: docker.io/nginx:1.16.0
--- a/library/CHART-TEMPLATE/tests/cl-images_registry_test.yaml
+++ b/library/CHART-TEMPLATE/tests/cl-images_registry_test.yaml
@@ -6,7 +6,7 @@ release:
   name: my-release
   namespace: my-namespace
 tests:
-  - it: does not set registry if not specified
+  - it: Sets docker.io as default registry if no registry set.
     asserts:
       - equal:
           path: spec.template.spec.containers[0].image
--- a/library/CHART-TEMPLATE/tests/cl-region.yaml
+++ b/library/CHART-TEMPLATE/tests/cl-region.yaml
@@ -125,23 +125,15 @@ tests:
         region: US
     asserts:
       - equal:
-          path: data.is_us
-          value: Yes
-      - equal:
-          path: data.is_eu
-          value: NO
-      - equal:
-          path: data.is_jp
-          value: NO
-      - equal:
-          path: data.is_gov
-          value: NO
-      - equal:
-          path: data.is_stg
-          value: NO
-      - equal:
-          path: data.is_dev
-          value: NO
+          path: data
+          value:
+            region: US
+            is_us: Yes
+            is_eu: NO
+            is_jp: NO
+            is_gov: NO
+            is_stg: NO
+            is_dev: NO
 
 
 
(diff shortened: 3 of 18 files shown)


NRhzhao (dismissed):

library/common-library/templates/_resolve.tpl
```diff
@@ -0,0 +1,43 @@
...
+  {{- else if (kindIs "bool" $val) -}}
+    {{/* If we get here, we know $val is a falsy bool, so let's return a falsy statement */}}
+  {{- else if $root.global -}}
+    {{- include "newrelic.common.resolve" (dict "ctx" $root.global "key" .key) -}}
+  {{- end -}}
+{{- end -}}
+
+
+{{- define "newrelic.common.resolve_or" -}}
+  {{- include "newrelic.common.resolve" (dict "ctx" .ctx "key" .key) | default .default -}}
```
jamescripter: should this be one of the cases of `newrelic.common.resolve` instead of a separate helper?

library/CHART-TEMPLATE/templates/example-cm-resolve.yaml
```diff
@@ -0,0 +1,28 @@
...
+  namespace: {{ .Release.Namespace }}
+data:
+  {{/* Using lookup With Default  */}}
+  custom: {{ include "newrelic.common.resolve_or" (dict "ctx" . "key" "custom"  "default" "myDefaultData") -}}
+
+  {{/* Look up only will return nothing if value not found */}}
+  proxy: {{ include "newrelic.common.resolve" (dict "ctx" . "key" "proxy") -}}
+
+  {{/* Example of using lookup with custom logic to set default */}}
+  endpoint: {{ include "newrelic.common.resolve" (dict "ctx" . "key" "endpoint") | default "inline-default" -}}
```
jamescripter, line 16: does this test anything more than the functionality of `default`? seems like this test may be redundant considering the test above it.

library/CHART-TEMPLATE/tests/cl-images_registry_test.yaml
jamescripter: maybe we should update the name of this test since it is now expecting registry to be set?

library/CHART-TEMPLATE/tests/cl-region.yaml
```diff
@@ -47,20 +57,209 @@ tests:
...
+          value: JP
+
+  - it: Set Japan region (locally)
+    set:
+      region: JP
+    asserts:
+      - equal:
+          path: data.region
+          value: JP
+
```
jamescripter, line 76: do we need to have these tests per region? I think the ones above these are sufficient since the priority chain works the same regardless of region

library/CHART-TEMPLATE/tests/cl-region.yaml
```diff
@@ -47,20 +57,209 @@ tests:
...
+    set:
+      global:
+        fedramp:
+          enabled: true
+    asserts:
+      - equal:
+          path: data.region
+          value: GOV
+
+  - it: Convenience Functions work for US
```
jamescripter, line 122: could probably shorten to one of these:

```
  asserts:
    - equal:
        path: [data.is_us, data.is_eu, data.is_gov, data.is_stg, data.is_dev]
        value: NO
    - equal:
        path: data.is_jp
        value: Yes
```

or 

```
  asserts:
    - match:
        data:
          is_us: NO
          is_eu: NO
          is_jp: Yes
          is_gov: NO
          is_stg: NO
          is_dev: NO
```

library/common-library/templates/_fedramp.tpl
```diff
@@ -1,25 +1,10 @@
...
-            {{- end -}}
-        {{- end -}}
-    {{- end -}}
+  {{- include "newrelic.common.resolve" (dict "ctx" . "key" "fedramp.enabled") -}}
 {{- end -}}
 
 
-
 {{- /* Return FedRAMP value directly ready to be templated */ -}}
 {{- define "newrelic.common.fedramp.enabled.value" -}}
```
jamescripter, line 8: should we remove this def? seems like it just supports checking for equality instead of a boolean check. I think users would prefer boolean check?

library/common-library/templates/_region.tpl
```diff
@@ -32,30 +36,27 @@ Usage: `include "newrelic.common.region.validate" "us"`
...
-  Staging
-{{- else if eq $region "local" -}}
-  Local
+{{- else if eq $region "jp" -}}
+  JP
+{{- else if eq $region "stg" -}}
+  STG
+{{- else if eq $region "gov" -}}
+  GOV
+{{- else if eq $region "dev" -}}
```
jamescripter, line 45: doesn't look like this case is possible

library/common-library/templates/_resolve.tpl
```diff
@@ -0,0 +1,43 @@
...
+  {{- $val := $root -}}
+  {{- range $keys -}}
+    {{- if or ($val) (kindIs "bool" $val) -}}
+      {{- $val = index $val . -}}
+    {{- else -}}
+      {{- $val = "" -}}
+    {{- end -}}
+  {{- end -}}
+
+  {{- if or ($val) -}}
```
jamescripter: probably don't need `or` here

library/common-library/templates/_resolve.tpl
```diff
@@ -0,0 +1,43 @@
...
+    {{- $val -}}
+  {{- else if (kindIs "bool" $val) -}}
+    {{/* If we get here, we know $val is a falsy bool, so let's return a falsy statement */}}
+  {{- else if $root.global -}}
+    {{- include "newrelic.common.resolve" (dict "ctx" $root.global "key" .key) -}}
```
jamescripter, line 38: nice recursion! I didn't know this was possible in templates

library/common-library/templates/_resolve.tpl
```diff
@@ -0,0 +1,43 @@
...
+
+{{- define "newrelic.common.resolve" -}}
+  {{- $ctx := .ctx -}}
+  {{- $keys := splitList "." .key -}}
+  {{- $root := $ctx.Values | default $ctx -}}
+
+  {{- $val := $root -}}
+  {{- range $keys -}}
+    {{- if or ($val) (kindIs "bool" $val) -}}
+      {{- $val = index $val . -}}
```
jamescripter: won't `index` fail on any non-indexable `$val`?

library/common-library/templates/_fedramp.tpl
jamescripter: nvm I guess it would ensure backwards compatibility

library/common-library/templates/_resolve.tpl
Philip-R-Beckwith: Good catch.

library/common-library/templates/_fedramp.tpl
Philip-R-Beckwith: Yup, I don't love this function, and I am half tempted to just rip it out. 
Haven't made up my mind.... We'll see which way it goes....

library/CHART-TEMPLATE/tests/cl-region.yaml
Philip-R-Beckwith: Looks like you may have hade some AI hallucinations on these suggestions as the proposed syntax isn't supported. 

However I genuinely do like this Idea, and was able to get it to work with.... 
```
      - equal:
          path: data
          value:
            region: STG
            is_us: NO
            is_eu: NO
            is_jp: NO
            is_gov: NO
            is_stg: Yes
            is_dev: NO
```

library/common-library/templates/_resolve.tpl
Philip-R-Beckwith: Same actually lol. 
Learned it specifically for this function.

library/common-library/templates/_resolve.tpl
Philip-R-Beckwith: Sometimes we may want the chart to fail. 
Would be a clear indication that something is misconfigured. 

If the user supplies a config that doesn't have the correct path this would alert them. 

Or if the dev writes a path that won't work, this would resolve quietly instead of letting them know there's an issue. 

So we want the chart to fail if the path is wrong.

library/common-library/templates/_region.tpl
Philip-R-Beckwith: We do have unit tests that look for this case. 

```
  - it: Convenience Functions work for Dev
    set:
      region: dev
      ...
```

While the other regions have flags that set them such as the license key fedramp and nrStaging, dev.... isn't a real region.... There's no endpoints in dev... So we'd expect the dev working on it to be setting their own endpoints and what not. 

For this reason we have a less support for this region.... 
It's a very special use case. To set a chart to the "dev" region, the user has to set the `region` or `global.region` to `dev`. 

TLDR: 
dev can come from `newrelic.common.region._fromValues`

```
{{- include "newrelic.common.region.validate" (include "newrelic.common.region._fromValues" . | default $region ) -}}
```

library/CHART-TEMPLATE/tests/cl-region.yaml
Philip-R-Beckwith: 🤔  You right.
No point in ripping them out though. 
These tests take maybe 5 extra MS per run, and do offer a bit of extra protection. 

But if I was writing it again, I probably wouldn't dupe all of the tests again.

library/CHART-TEMPLATE/tests/cl-images_registry_test.yaml
Philip-R-Beckwith: Good catch

library/CHART-TEMPLATE/templates/example-cm-resolve.yaml
Philip-R-Beckwith: Not really, it's just here as an example of ways that the function can be used. 
Its not a test so much... It's more like documentation for people to get ideas from.

library/common-library/templates/_resolve.tpl
Philip-R-Beckwith: 🤔  
I don't hate that idea... Does make the function itself a bit longer... Buuuuut... It's not a bad idea... The "give me a default if it fails" is kinda inline documented by the fact that we're passing "default" as a key.... 🤔 ...

library/common-library/templates/_resolve.tpl
jamescripter: makes sense

library/common-library/templates/_resolve.tpl
jamescripter: I guess the main reason I suggested this is that the name is not completely intuitive. but resolve_with_default is too long. that's why I thought to consolidate everything under `resolve` but I'm not sure now

jamescripter (approved):

library/common-library/templates/_resolve.tpl
jamescripter: "The "give me a default if it fails" is kinda inline documented by the fact that we're passing "default" as a key"

wouldn't this be the case either way? since you would pass in default to either resolve or resolve_or

Philip-R-Beckwith: > I guess the main reason I suggested this is that the name is not completely intuitive. but resolve_with_default is too long. that's why I thought to consolidate everything under resolve but I'm not sure now 

My thoughts exactly. The initial implementation of this had the name "override_lookup" and "override_lookup_with_default" which, felt too long. Although I didn't love the ambiguity of the hanging "or." Maybe following Java's promise naming would have worked resolve_or_else.... Kinda sounds like a threat lol 😅 ...

> wouldn't this be the case either way? since you would pass in default to either resolve or resolve_or

Yes. Either way the fact that we pass a "default" key in the map is basically inline documentation that we'll get a default value if one isn't found. So that's ultimately why I switched to just having the default case in the resolve function.

PR #2237, opened 2026-05-04
[pipeline-control-gateway] NR-551079: add DaemonSet support                                                                                                                                          

kush-newrelic:
Is this a new chart:
  NO

  What this PR does / why we need it:
  Adds DaemonSet support to the pipeline-control-gateway chart. When daemonset.enabled is set to true, the chart deploys PCG as a DaemonSet (one pod per node) instead of a Deployment with
  HPA. This enables node-local telemetry collection patterns.

  Key changes:
  - New daemonset.yaml and daemonset-configmap.yaml templates
  - Helper templates extended for daemonset-specific nodeSelector, tolerations, affinity, and securityContext
  - Deployment, deployment-configmap, and HPA are conditionally disabled when daemonset mode is enabled
  - Service selector switches between deployment/daemonset component labels
  - DaemonSet uses same resource defaults as Deployment (2048Mi/1000m)
  - DaemonSet configmap uses same Values.generated pattern as deployment configmap
  - FLEET_ID and NEW_RELIC_HOST env vars included in daemonset template

  Special notes for your reviewer:
  - The daemonset and deployment modes are mutually exclusive, controlled by daemonset.enabled (default: false)
  - The daemonset configmap uses the same Values.generated pipeline configuration as the deployment configmap
  - All existing deployment functionality remains unchanged when daemonset.enabled is false
  - DaemonSet includes updateStrategy (RollingUpdate, maxUnavailable: 1)
  - Rebased on latest master (includes multi-fleet/fleet_id changes)

  Checklist:
  - Variables are documented in the README.md
  - Title of the PR starts with chart name

### DIFF
--- a/charts/pipeline-control-gateway/templates/daemonset-configmap.yaml
+++ b/charts/pipeline-control-gateway/templates/daemonset-configmap.yaml
@@ -1,8 +1,8 @@
-{{- if not .Values.daemonset.enabled }}
+{{- if .Values.daemonset.enabled }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ include "nrKubernetesOtel.deployment.configMap.fullname" . }}
+  name: {{ include "nrKubernetesOtel.daemonset.configMap.fullname" . }}
   namespace: {{ .Release.Namespace }}
 data:
   gateway-config.yaml: |
@@ -10,11 +10,11 @@ data:
       Metric:
         - interval.ms
         - metricName
-  deployment-config.yaml: |
-    {{- with include "nrKubernetesOtel.deployment.configMap.config" . }}
+  daemonset-config.yaml: |
+    {{- with include "nrKubernetesOtel.daemonset.configMap.config" . }}
     {{- . | nindent 4 }}
     {{- end }}
-    {{- if not (include "nrKubernetesOtel.deployment.configMap.config" .) }}
+    {{- if not (include "nrKubernetesOtel.daemonset.configMap.config" .) }}
     {{- if .Values.generated }}
     {{- tpl (.Values.generated | toYaml) . | nindent 4 }}
     {{- else }}
--- a/charts/pipeline-control-gateway/templates/daemonset.yaml
+++ b/charts/pipeline-control-gateway/templates/daemonset.yaml
@@ -1,0 +1,146 @@
+{{- if .Values.daemonset.enabled }}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ include "nrKubernetesOtel.daemonset.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "newrelic.common.labels" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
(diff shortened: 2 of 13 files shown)


charts/pipeline-control-gateway/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,193 @@
...
+        client:
+          compression: gzip
+          timeout: {{ .Values.nrcollectorexporter.timeout }}
+      otlp:
+        protocols:
+          http:
+            endpoint: ${env:MY_POD_IP}:{{ .Values.ports.otlpHttp }}
+          grpc:
+            endpoint: ${env:MY_POD_IP}:{{ .Values.ports.otlpGrpc }}
+      prometheus/usage:
```
prafulla-kiran: This is our usage exporter right - we should remove this from here. I understand this is the default - but we shouldn't have it here

charts/pipeline-control-gateway/templates/daemonset-configmap.yaml
```diff
@@ -0,0 +1,193 @@
...
+                  regex: 'service_version|service_name|service_instance_id'
+                - source_labels: [ __name__ ]
+                  regex: 'godebug_.*'
+                  action: drop
+
+    processors:
+      memory_limiter:
+        check_interval: {{ .Values.processors.memory_limiter.check_interval }}
+        limit_mib:  {{ .Values.processors.memory_limiter.limit_mib }}
+      nrprocessor:
```
prafulla-kiran: We don't have nrprocessor anymore right - can we update this with the latest config

charts/pipeline-control-gateway/templates/daemonset.yaml
```diff
@@ -0,0 +1,142 @@
...
+                  fieldPath: metadata.name
+            - name: MY_POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: GOMEMLIMIT
+              valueFrom:
+                resourceFieldRef:
+                  resource: limits.memory
+            - name: NEW_RELIC_LICENSE_KEY
```
prafulla-kiran, line 85: Prasanthi has some changes related to fleet_id - we should probably include that here as well - for the multi-fleet stuff

prafulla-kiran (changes requested):

charts/pipeline-control-gateway/templates/daemonset.yaml
prafulla-kiran: @kush-newrelic You might want to rebase this after recent changes for mult-fleet

charts/pipeline-control-gateway/templates/daemonset-configmap.yaml
kush-newrelic: yes. It was not the latest. I did the downmerge from master Its fine now.

charts/pipeline-control-gateway/templates/daemonset.yaml
kush-newrelic: yes. done.

prafulla-kiran (approved):

PR #2238, opened 2026-05-04
[agent-control-bootstrap] feat(watch): watching resources

paologallinaharbur:
This fix is to mitigate an issue that we are facing from time to time with the ci/cd pipeline https://github.com/newrelic/newrelic-agent-control/actions/runs/25091889289/job/73519561304

The root cause seems to be a race condition in the Helm binary itself, a watch is usually not needed since all resources are created and ready without the need to wait on them with a watch.


sigilioso (approved):
Thanks! LGTM

PR #2244, opened 2026-05-05
[Common Libary] endpoint funcions should not fail by default

Philip-R-Beckwith:
Some charts do not throw when customers use custom license keys without specifying region, others do. 
Odds are that customers are using the US env, and for charts to start throwing on installation just because a customer isn't including the license key in the chart would be a big breaking change. 

This update keeps the functionality of region the same, however provides functions that will fall back to US when region isn't explicitly set and the user is using custom secrets. 

Also adds some helper functions for creating a notes.txt. This will allow us to put warnings front and center for our customers for all sorts of things. First up is the warning about not setting region. But as we develop more use-cases for this chart we'll continue to add more notes to help our customer's instalation process go smoother. 

  - fixes #

#### Checklist
- [x] Chart Version bumped

## 🚀 What's Changed


library/CHART-TEMPLATE/templates/example-cm-region-with-default.yaml
```diff
@@ -0,0 +1,15 @@
+# This is a dummy CM to test that region is templated and how to use it
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "newrelic.common.naming.fullname" . }}-examples-region
+  namespace: {{ .Release.Namespace }}
+data:
+  region: "{{ include "newrelic.common.region.or_us" . }}"
+
+  is_us: "{{ if include "newrelic.common.region.is_us" . }}Yes{{ else }}NO{{ end }}"
```
dbudziwojskiNR: nit: Should `Yes` and `NO` have the same case?

library/CHART-TEMPLATE/templates/example-cm-region.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
 data:
   region: "{{ include "newrelic.common.region" . }}"
 
+  {{/*This is to show an example of using region.fail_if_unresolvable */}}
+  region_with_interposed_check: "{{ include "newrelic.common.region.fail_if_unresolvable" . }}{{ include "newrelic.common.region" . }}"
```
dbudziwojskiNR: Thoughts on this being one function instead of two. Otherwise it might be easy to forget `{{ include "newrelic.common.region.fail_if_unresolvable" . }}`

library/common-library/templates/_notes.tpl
```diff
@@ -0,0 +1,21 @@
+{{/*Warning notes*/}}
+{{ define "newrelic.common.notes.region" }}
+{{ if and (include "newrelic.common.license._usesCustomSecret" .) (not (include "newrelic.common.region._fromValues" .)) }}
+Warning: This chart was installed using a custom secret for the license key, but no region was specified. 
+         You may need to set the `region` or `global.region` explicitly if your license key is for the EU or Japan regions.
```
dbudziwojskiNR, line 5:
```suggestion
         You may need to set the `region` or `global.region` explicitly if your license key is for the EU or Japan regions. An explicitly configured `region` or `global.region` will be required in a future release.

```

library/CHART-TEMPLATE/templates/example-cm-region.yaml
Philip-R-Beckwith: newrelic.common.region
Already calls fail_if_unresolvable. Users have to call newrelic.common.region.or_us to not fail. 

newrelic.common.region's implementation is just 
{
newrelic.common.region.fail_if_unresolvable
newrelic.common.region.or_us
}

library/CHART-TEMPLATE/templates/example-cm-region-with-default.yaml
Philip-R-Beckwith: Sure. I'll update it. 👍

library/common-library/templates/_notes.tpl
Philip-R-Beckwith: 👍

dbudziwojskiNR (approved):

PR #2257, opened 2026-05-12
[agent-control-deployment] bump common-library dependency [NR-554515]

sigilioso:

This PR:
- Bumps `common-library` from `1.4.0` → `2.X`.
- Adds support for JP region
- Replaces the hand-rolled region → endpoint switch in `_helpers.tpl` (~55 lines) with thin
  wrappers around `common-library`'s endpoint helpers (`opamp_endpoint`,
  `public_keys_endpoint`, `system_identity_oauth_endpoint`).
- Removes the undocumented `.Values.development.backend.*` dev escape hatch. Dev overrides
  now go through the standard `.Values.opampEndpoint` / `.Values.publicKeysEndpoint` /
  `.Values.systemIdentityOauthEndpoint` knobs (also honored under `.Values.global.*`).
- Adds unit tests locking in the STG rendering, the override path, and JP region.

#### ⚠️  Internal breaking changes

**Region identifier renames** (introduced when bumping common-library to 2.X):

`common-library` 2.x renamed the values returned by `newrelic.common.region`. Since AC's
old endpoint switch compared against the pre-2.x strings, those paths were silently dead
after the bump. Callers who set `.Values.region` explicitly must migrate:

| Before (`common-library` 1.x) | After (`common-library` 2.x) |
| ----------------------------- | ---------------------------- |
| `Staging`                     | `stg`                        |
| `Local`                       | `dev`                        |
| `US` / `EU`                   | unchanged (`us`/`eu` also accepted — validation is case-insensitive) |

Users relying on `nrStaging: true` (not `region: Staging`) are unaffected — the staging
routing still happens automatically.

`.Values.development.backend.*` **removed**:

The undocumented dev-only escape hatch under `.Values.development.backend.{fleet_control,tokenRenewal}`
is no longer read. It's replaced by the standard override mechanism shared with every other
New Relic chart.

**Before**
```yaml
region: Local
development:
  backend:
    fleet_control: http://localhost:8080/v1/opamp
    tokenRenewal: http://localhost:8080/oauth2/token
```

**After**
```yaml
# No region needed; these overrides apply in every region.
opampEndpoint: http://localhost:8080
systemIdentityOauthEndpoint: http://localhost:8080
# .Values.publicKeysEndpoint also available
# .Values.global.* variants work too
```

Note the values are now **base URLs** — the path suffixes (`/v1/opamp`, `/oauth2/token`,
`/r/blob-management/.../jwks.json`) are appended by the chart.

#### Unsupported regions (GOV and others) fail fast with an actionable error

`common-library`'s endpoint helpers don't define opamp/public-keys entries for GOV. Previously these regions were silently routed to US endpoints. Now they fail rendering with a clear message pointing at the exact override value:

```
This chart does not have support for the region provided 'GOV'. Please either supply an
override URL via .Values.opampEndpoint or .Values.global.opampEndpoint, or set a different
region.
```

#### Checklist
- [x] Chart Version bumped

## 🚀 What's Changed

### DIFF
--- a/charts/agent-control-deployment/templates/_helpers.tpl
+++ b/charts/agent-control-deployment/templates/_helpers.tpl
@@ -1,59 +1,28 @@
 {{- /*
-Return to which endpoint should the agent control connect to get fleet_control data
+Return the OpAMP endpoint the agent control connects to for fleet_control data.
+The base URL is resolved per-region by the common-library helper, which also
+honors overrides via `.Values.opampEndpoint` / `.Values.global.opampEndpoint`.
 */ -}}
 {{- define "newrelic-agent-control.config.endpoints.fleet_control" -}}
-{{- $region := include "newrelic.common.region" . -}}
-
-{{- if eq $region "Staging" -}}
-  https://opamp.staging-service.newrelic.com/v1/opamp
-{{- else if eq $region "EU" -}}
-  https://opamp.service.eu.newrelic.com/v1/opamp
-{{- else if eq $region "US" -}}
-  https://opamp.service.newrelic.com/v1/opamp
-{{- else if eq $region "Local" -}}
-  {{- /* Accessing the value directly without protection. A developer should now how to read the error. */ -}}
-  {{ .Values.development.backend.fleet_control }}
-{{- else -}}
-  {{- fail "Unknown/unsupported region set for this chart" -}}
-{{- end -}}
+{{ include "newrelic.common.opamp_endpoint" . }}/v1/opamp
 {{- end -}}
 
 {{- /*
-Return to which endpoint should the agent control connect to get the public key for signature validation
+Return the endpoint from which the agent control fetches the public key used
+for signature validation. Override via `.Values.publicKeysEndpoint` /
+`.Values.global.publicKeysEndpoint`.
 */ -}}
 {{- define "newrelic-agent-control.config.endpoints.public_key_server_url" -}}
-{{- $region := include "newrelic.common.region" . -}}
-
-{{- if eq $region "Staging" -}}
-  https://staging-publickeys.newrelic.com/r/blob-management/global/agentconfiguration/jwks.json
-{{- else if eq $region "EU" -}}
-  https://publickeys.eu.newrelic.com/r/blob-management/global/agentconfiguration/jwks.json
-{{- else if eq $region "US" -}}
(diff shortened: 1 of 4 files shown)


charts/agent-control-deployment/templates/_helpers.tpl
```diff
@@ -1,59 +1,28 @@
...
 {{- define "newrelic-agent-control.config.endpoints.tokenRenewal" -}}
-{{- $region := include "newrelic.common.region" . -}}
-
-{{- if eq $region "Staging" -}}
-  https://system-identity-oauth.staging-service.newrelic.com/oauth2/token
-{{- else if eq $region "EU" -}}
-  https://system-identity-oauth.service.newrelic.com/oauth2/token
-{{- else if eq $region "US" -}}
-  https://system-identity-oauth.service.newrelic.com/oauth2/token
-{{- else if eq $region "Local" -}}
```
gsanchezgavier, line 51: are we removing this right? was it used ?

charts/agent-control-deployment/tests/configmap_agentcontrol_config_test.yaml
```diff
@@ -310,6 +310,112 @@ tests:
...
               port: 51200
 
+  - it: agent control's config honors the JP endpoints
+    # NOTE: requires common-library >= 2.2.0 (JP entries for opamp/public-keys endpoints).
+    # On common-library 2.1.1 this fails with "This chart does not have support for the region provided 'JP'".
+    set:
+      cluster: my-cluster
+      region: jp
+    asserts:
+      - equal:
```
gsanchezgavier: nit: have you check if there is no way to assert on contains or something more foucused ?

gsanchezgavier (dismissed):
thanks

charts/agent-control-deployment/templates/_helpers.tpl
gsanchezgavier: oh just read the pr description 👍

charts/agent-control-deployment/tests/configmap_agentcontrol_config_test.yaml
sigilioso: I've just simplified it with matchRegex, LTMWYT

gsanchezgavier (approved):
thanks!

PR #2275, opened 2026-05-26
[pipeline-control-gateway]  feat: add customConfigMap and config-watcher sidecar for DaemonSet mode

kush-newrelic:
---                                                                                                                                                                                                    
  What this PR does / why we need it:                                                                                                                                                                      
                                                                                                                                                                                                         
  Adds customConfigMap and configWatcher sidecar support for the DaemonSet mode in pipeline-control-gateway, matching the existing Deployment mode behavior.
                                                                                                                                                                                                           
  Changes:
  - values.yaml — Added daemonset.customConfigMap and daemonset.configWatcher settings                                                                                                                     
  - daemonset.yaml — Added config-watcher sidecar container and conditional volume source (custom vs chart-generated ConfigMap)                                                                            
  - daemonset-configmap.yaml — Skip chart-generated ConfigMap when external customConfigMap is provided                        
  - Chart.yaml — Version bump 2.3.1 → 2.4.0                                                                                                                                                                
                                                                                                                                                                                                           
  This allows the control plane (Agent Control) to push config updates to an external ConfigMap for DaemonSet mode, with the config-watcher sidecar automatically restarting pods when changes are detected
   — same pattern used in Deployment mode.                                                                                                                                                                 
                                                                                                                                                                                                           
  Which issue this PR fixes                                                                                                                                                                                
                                                                                                                                                                                                         
  - fixes NR-569112

  Special notes for your reviewer:                                                                                                                                                                         
   
  - DaemonSet customConfigMap uses the same ConfigMap name as Deployment mode (as its required for the flux less flow and confirmed by the AC team)                                                                                          
  - Config-watcher sidecar is identical to the Deployment version — watches ConfigMap resourceVersion and deletes the pod on change                                                                      
  - Tested with helm template for all three scenarios: daemonset+customConfigMap, daemonset without customConfigMap, and deployment mode (all pass)

### DIFF
--- a/charts/pipeline-control-gateway/templates/daemonset.yaml
+++ b/charts/pipeline-control-gateway/templates/daemonset.yaml
@@ -127,59 +127,8 @@ spec:
           volumeMounts:
             - name: daemonset-config
               mountPath: /config
-        {{- if and .Values.daemonset.customConfigMap .Values.daemonset.configWatcher.enabled }}
-        - name: config-watcher
-          image: {{ .Values.daemonset.configWatcher.image }}
-          imagePullPolicy: {{ .Values.daemonset.configWatcher.imagePullPolicy }}
-          command:
-            - /bin/bash
-            - -c
-            - |
-              set -e
-
-              CONFIGMAP_NAME="{{ .Values.daemonset.customConfigMap }}"
-              NAMESPACE="{{ .Release.Namespace }}"
-              POD_NAME="${MY_POD_NAME}"
-
-              echo "Config watcher started. Monitoring ConfigMap: ${CONFIGMAP_NAME}"
-              echo "Will restart pod: ${POD_NAME} in namespace: ${NAMESPACE} when a configuration update is detected"
-
-              LAST_VERSION=$(kubectl get configmap "${CONFIGMAP_NAME}" -n "${NAMESPACE}" -o jsonpath='{.metadata.resourceVersion}')
-              echo "Initial ConfigMap resourceVersion: ${LAST_VERSION}"
-
-              while true; do
-                echo "Opening watch on ConfigMap ${CONFIGMAP_NAME}..."
-                while read -r CURRENT_VERSION; do
-                  if [ -z "$CURRENT_VERSION" ]; then
-                    sleep 1
-                    continue
-                  fi
-
-                  echo "Event received. ResourceVersion: ${CURRENT_VERSION}"
-
-                  if [ "$CURRENT_VERSION" != "$LAST_VERSION" ]; then
-                    echo "ConfigMap change detected! ResourceVersion changed from ${LAST_VERSION} to ${CURRENT_VERSION}"
-                    echo "Deleting pod ${POD_NAME} to trigger restart..."
-
-                    kubectl delete pod "${POD_NAME}" -n "${NAMESPACE}" --grace-period={{ .Values.daemonset.configWatcher.gracePeriod }}
-
-                    echo "Pod deletion initiated. Kubernetes will recreate the pod with new config."
-                    exit 0
-                  fi
-
-                  LAST_VERSION="${CURRENT_VERSION}"
-                done < <(kubectl get configmap "${CONFIGMAP_NAME}" -n "${NAMESPACE}" --watch -o jsonpath='{.metadata.resourceVersion}{"\n"}' || true)
-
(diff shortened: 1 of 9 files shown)


vkumarsabbu (dismissed):

charts/pipeline-control-gateway/templates/daemonset.yaml
```diff
@@ -127,10 +127,69 @@ spec:
...
+                    echo "Pod deletion initiated. Kubernetes will recreate the pod with new config."
+                    exit 0
+                  fi
+
+                  LAST_VERSION="${CURRENT_VERSION}"
+                done < <(kubectl get configmap "${CONFIGMAP_NAME}" -n "${NAMESPACE}" --watch -o jsonpath='{.metadata.resourceVersion}{"\n"}' || true)
+
+                echo "Watch stream ended (server timeout or transient error). Reconnecting in 2s..."
+                sleep 2
+              done
```
paologallinaharbur: it would be better to reuse this code from a tpl function instead of copy/pasting it

charts/pipeline-control-gateway/values.yaml
```diff
@@ -41,6 +41,28 @@ daemonset:
...
+    # -- Grace period in seconds for pod deletion when ConfigMap changes
+    gracePeriod: 30
+    # -- Resource limits and requests for the config-watcher sidecar
+    resources:
+      limits:
+        cpu: 100m
+        memory: 128Mi
+      requests:
+        cpu: 10m
+        memory: 64Mi
```
paologallinaharbur: you could also decide to extract this out of deployment and daemonset to avoid duplicating the values

charts/pipeline-control-gateway/templates/role.yaml
```diff
@@ -1,16 +1,24 @@
-{{- if and (include "newrelic.common.serviceAccount.create" .) .Values.deployment.customConfigMap .Values.deployment.configWatcher.enabled }}
+{{- $deploymentWatcher := and (not .Values.daemonset.enabled) .Values.deployment.customConfigMap .Values.deployment.configWatcher.enabled }}
+{{- $daemonsetWatcher := and .Values.daemonset.enabled .Values.daemonset.customConfigMap .Values.daemonset.configWatcher.enabled }}
```
paologallinaharbur: for example this condition can be simplified if we simply have `.Values.configWatcher.enabled`

charts/pipeline-control-gateway/templates/rolebinding.yaml
```diff
@@ -1,15 +1,17 @@
-{{- if and (include "newrelic.common.serviceAccount.create" .) .Values.deployment.customConfigMap .Values.deployment.configWatcher.enabled }}
+{{- $deploymentWatcher := and (not .Values.daemonset.enabled) .Values.deployment.customConfigMap .Values.deployment.configWatcher.enabled }}
+{{- $daemonsetWatcher := and .Values.daemonset.enabled .Values.daemonset.customConfigMap .Values.daemonset.configWatcher.enabled }}
+{{- if and (include "newrelic.common.serviceAccount.create" .) (or $deploymentWatcher $daemonsetWatcher) }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  name: {{ include "nrKubernetesOtel.deployment.fullname" . }}-config-watcher
+  name: {{ include "newrelic.common.naming.fullname" . }}-config-watcher
```
paologallinaharbur, line 6: this change seems unrelated

paologallinaharbur (commented):
You can decide to remove some code duplication and avoid duplicating the values, but the implementation is ok

charts/pipeline-control-gateway/templates/daemonset.yaml
kush-newrelic: charts/pipeline-control-gateway/templates/_config_watcher.tpl done

charts/pipeline-control-gateway/templates/rolebinding.yaml
kush-newrelic: reverted

charts/pipeline-control-gateway/templates/role.yaml
kush-newrelic: done

vkumarsabbu (approved):

PR #2279, opened 2026-05-27
[pipeline-control-gateway] fix: DaemonSet config path when customConfigMap is set

kush-newrelic:

Fixes the DaemonSet config file path when `customConfigMap` is set. Agent Control writes config under the key `deployment-config.yaml` in the external ConfigMap regardless of mode (Deployment or DaemonSet). The DaemonSet template now conditionally uses `/config/deployment-config.yaml` when `customConfigMap` is set, and `/config/daemonset-config.yaml` when using the chart-generated ConfigMap.

Without this fix, DaemonSet pods crash with `CrashLoopBackOff` because they look for `/config/daemonset-config.yaml` which doesn't exist in the external ConfigMap.

- fixes NR-569112

- Tested with `helm template` for both cases (with/without customConfigMap)
- Verified on EKS cluster: pods were in CrashLoopBackOff before fix, running correctly after

#### Checklist
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[pipeline-control-gateway]`)

### DIFF
--- a/charts/pipeline-control-gateway/templates/daemonset.yaml
+++ b/charts/pipeline-control-gateway/templates/daemonset.yaml
@@ -48,7 +48,11 @@ spec:
           {{- end }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if .Values.daemonset.customConfigMap }}
+          args: ["./pipeline-control-gateway", "--config", "/config/deployment-config.yaml"]
+          {{- else }}
           args: ["./pipeline-control-gateway", "--config", "/config/daemonset-config.yaml"]
+          {{- end }}
           resources:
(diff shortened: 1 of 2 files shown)


charts/pipeline-control-gateway/templates/daemonset.yaml
```diff
@@ -48,7 +48,11 @@ spec:
           {{- end }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if .Values.daemonset.customConfigMap }}
+          args: ["./pipeline-control-gateway", "--config", "/config/deployment-config.yaml"]
+          {{- else }}
```
vkumarsabbu, line 53: I don't see a change in the if & else block. does it really require?

vkumarsabbu (approved):

PR #2327, opened 2026-07-06
synthetics-job-manager: add global.customCertificates for K8s custom CA delivery

kanaksharma58:
## What this PR does 

Adds a new optional `global.customCertificates` block to the `synthetics-job-manager` chart. When enabled, Helm mounts a customer-provided volume (typically a Kubernetes Secret holding CA PEM file(s)) **read-only** into the runtime pods and sets the `RUNTIME_CERT_PATH` env var so the runtime cert-loading logic knows where to read them. This lets Synthetics runtimes trust internal/private Certificate Authorities without rebuilding images.

Follows the existing `customNodeModules` volume-mount pattern (`_helpers.tpl` helpers + gated template blocks).

```yaml
global:
  customCertificates:
    enabled: false
    volumeMountPath: /var/lib/newrelic/synthetics/certs
    volume: {}
    # volume:
    #   secret:
    #     secretName: my-internal-ca-bundle
```

## Changes
- `values.yaml` — new `global.customCertificates` block.
- `templates/_helpers.tpl` — `customCertificatesVolumeName`, `toMountCustomCertificates`, `customCertificatesMount` (`readOnly: true`), `customCertificatesVolume`, `customCertificatesEnv`.
- `charts/node-api-runtime/templates/job.yaml` & `charts/node-browser-runtime/templates/job.yaml` — gated cert env + volumeMount + volume (consolidated into single `volumeMounts`/`volumes` keys to avoid duplicate-key YAML when combined with `customNodeModules`).
- `ci/custom-certs-values.yaml` — chart-testing values exercising the feature.
- `README.md` — documents the three new parameters.
- Version bumps: `node-api-runtime` 1.0.57→1.0.58, `node-browser-runtime` 2.0.22→2.0.23, and parent `synthetics-job-manager` 3.0.69→3.0.70 (with matching dependency pins).

## Scope notes
- **node-api + node-browser runtimes only.** Ping-runtime is intentionally **excluded** (CDD: "Ping-runtime yet to decide to add the cert support") — this differs from the ticket's acceptance criteria, which listed ping.
- Mount path default is `/var/lib/newrelic/synthetics/certs` (CDD / Docker-mode path), not the ticket's `/etc/nr-custom-certs`.
- Scope is `global.customCertificates` only; `global.clientCertificate` (mTLS, Requirement 2) is a separate follow-up.
- Runtime-side cert-loading scripts land separately in the node-api / node-browser runtime repos (Swetha's tickets).

## Validation
- `helm lint` clean.
- Default render (feature disabled) is byte-identical to before aside from the version-label bump.
- Enabled render (`-f ci/custom-certs-values.yaml`): `RUNTIME_CERT_PATH` env + `readOnly: true` mount + secret-backed volume present in node-api and node-browser CronJobs only; **zero** cert artifacts in ping-runtime or the SJM deployment.
- Overriding `volumeMountPath` propagates to both the mount path and the env value.
- `customNodeModules` + `customCertificates` together render into single `volumeMounts`/`volumes` keys (no duplicate keys).


charts/synthetics-job-manager/values.yaml
```diff
@@ -58,6 +58,23 @@ global:
     ## and include the package.json file before installing Helm Chart.
     customNodeModulesPath: ""
 
+  # Mounts a Kubernetes Secret containing custom CA certificate PEM file(s) into the runtime pods so they trust
+  # internal/private Certificate Authorities. When enabled, Helm mounts the volume read-only into the runtime pods
+  # and sets the RUNTIME_CERT_PATH environment variable so the runtime cert-loading logic knows where to look.
+  customCertificates:
+    ## Set to true to mount the custom certificate volume into the runtime pods.
+    enabled: false
```
syeddula-279: **Suggestion:** drop enabled and volumeMountPath — just activate on volume presence
The runtime already decides everything from RUNTIME_CERT_PATH: run.sh/cert-inject.sh only act when it's set, and legacy behavior stays on when it isn't. So enabled is a second switch we don't need.

**Proposed:** if volume is empty → mount nothing, don't set RUNTIME_CERT_PATH; if it's set → mount read-only and set RUNTIME_CERT_PATH. This matches the existing customNodeModules/userDefinedVariables helpers (both activate on presence, no enabled flag) and avoids confusing states like enabled: true + volume: {} (invalid volume) or enabled: false + a real Secret (silently no validation).

kanaksharma58: sure, this is cleaner and matches the existing customNodeModules/userDefinedVariables convention  I'll drop both enabled and volumeMountPath: gate on volume presence (empty → mount nothing + don't set RUNTIME_CERT_PATH, that is legacy behavior; set → mount read-only + set RUNTIME_CERT_PATH), and make the mount path a fixed internal constant like customNodeModulesPath since the runtime only keys off RUNTIME_CERT_PATH. Will mirror the same for clientCertificate.

syeddula-279 (dismissed):

abhishek-kana (approved):

syeddula-279 (approved):
LGTM

Issue #2220, opened 2026-04-20
[newrelic-logging] Using newrelic-fluent-bit-output with fluentbit v5 fails liveness probe


JoeNorth:
**Bug description**

When using a fluent-bit v5 image with [newrelic-fluent-bit-output](https://github.com/newrelic/newrelic-fluent-bit-output) the liveness probes for the daemonset fail due to a 404 response from `localhost:2020/api/v1/health`.

**Version of Helm and Kubernetes**

All Helm and Kubernetes versions

**Which chart?**

All versions of `newrelic-logging`

**What happened?**

When attempting to use a fluent-bit:v5 image with the newrelic output plugin the daemonset's [liveness probes](https://github.com/newrelic/helm-charts/blob/master/charts/newrelic-logging/templates/daemonset.yaml#L179) fail as it returns a 404 error. 

**What you expected to happen?**

The default configuration passed through to fluent-bit should set `Health_Check On` to enable the health check endpoints.

**How to reproduce it?**

Create `fluent-bit.conf`:

```
      [SERVICE]
          Flush         1
          Log_Level     ${LOG_LEVEL}
          Daemon        off
          HTTP_Server   On
          HTTP_Listen   0.0.0.0
          HTTP_Port     2020
```

Run both fluent-bit v4 and v5:
```
$ docker run -p 2020:2020 -v $PWD/fluent-bit.conf:/fluent-bit/etc/fluent-bit.conf -d --rm -it fluent/fluent-bit:5.0.3 -c /fluent-bit/etc/fluent-bit.conf
$ curl localhost:2020/api/v1/health -vvvv
* Host localhost:2020 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:2020...
* Connected to localhost (::1) port 2020
> GET /api/v1/health HTTP/1.1
> Host: localhost:2020
> User-Agent: curl/8.7.1
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 404
< server: Fluent Bit
< x-http-engine: Monkey heritage
< content-length: 0
<
* Connection #0 to host localhost left intact

$ docker run -p 2020:2020 -v $PWD/fluent-bit.conf:/fluent-bit/etc/fluent-bit.conf -d --rm -it fluent/fluent-bit:4.2.3 -c /fluent-bit/etc/fluent-bit.conf

$ curl localhost:2020/api/v1/health
[...]

PR #2345, opened 2026-07-15
[newrelic-logging] Add Helm chart E2E test for default install (NR-587117)

satya-nr:

1. Adds the Helm chart E2E scenario for `newrelic-logging`, closing the gap identified, the chart previously had only static template checks (`helm-unittest`) and an install smoke test (`ct install`), with nothing verifying that logs actually reach New Relic.

2. Also adds a pod-health check: after the NRQL check, it asserts the pod is Ready with zero
restarts. Log delivery alone can pass even while the pod is crash-looping (see [#2220](https://github.com/newrelic/helm-charts/issues/2220)) — this
check closes that gap.

This adds a PR-gated workflow (`newrelic-logging-e2e.yml`) that installs the chart on Minikube, deploys a log-producer emitting uniquely-marked lines, and asserts via NRQL that they arrive in `Log` data.

  - fixes #

- Two workflows: newrelic-logging-e2e.yml (internal PRs) and newrelic-logging-e2e-external.yml (fork PRs — gated behind the E2E environment, requiring approval from the K8S-AGENTS team before secrets are exposed).
- Scope: Linux only. Both scenarios (default install, lowDataMode) run against Minikube on ubuntu-latest. Windows nodes and other K8s distributions (EKS/GKE/AKS) are not covered by CI yet.
- Uses the existing K8S_AGENTS_E2E_ACCOUNT_ID / _API_KEY / _LICENSE_KEY secrets already configured in this repo.
- No chart version bump — this PR only adds test/CI scaffolding, no template or values changes.

Next steps:
- [x] Add Windows e2e coverage — tracked in https://new-relic.atlassian.net/browse/NR-601614, needs its own test harness (no Minikube equivalent for Windows nodes).

#### Checklist
- [ ] Chart Version bumped

## 🚀 What's Changed

### DIFF
--- a/.github/workflows/newrelic-logging-e2e-external.yml
+++ b/.github/workflows/newrelic-logging-e2e-external.yml
@@ -1,7 +1,7 @@
-name: Newrelic Logging chart E2E
+name: Newrelic Logging chart E2E (External Contribution)
 
 on:
-  pull_request:
+  pull_request_target:
     paths:
       - 'charts/newrelic-logging/**'
       - '.github/workflows/newrelic-logging-e2e.yml'
@@ -16,14 +16,17 @@ env:
   MINIKUBE_VERSION: 'v1.38.1'
 
 jobs:
-  e2e-test:
+  e2e-test-external:
     name: Run Tests
-    # Do not run e2e tests if PR has skip-e2e label or if it's from a forked repo (external contribution)
-    if: ${{ !contains(github.event.pull_request.labels.*.name, 'ci/skip-e2e') && github.event.pull_request.head.repo.full_name == 'newrelic/helm-charts' }}
+    # Do not run e2e tests if PR has skip-e2e label or if it's a non-forked repo (internal contribution)
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'ci/skip-e2e') && github.event.pull_request.head.repo.full_name != 'newrelic/helm-charts' }}
     runs-on: ubuntu-latest
+    environment: E2E # Required so that approval is required to run this job since it is untrusted (forked).
     timeout-minutes: 20
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
       - name: Setup Minikube
         uses: manusa/actions-setup-minikube@v2.18.0
         with:
--- a/charts/newrelic-logging/e2e/test-specs.yml
+++ b/charts/newrelic-logging/e2e/test-specs.yml
@@ -1,8 +1,7 @@
 description: New Relic newrelic-logging chart E2E
 
-# The e2e-action appends `WHERE <custom_test_key> = '<SCENARIO_TAG>'` to every NRQL query,
-# scoping each run to its own data. The chart records `cluster_name` from `global.cluster`,
-# which we set to ${SCENARIO_TAG} on install below — so every query is scoped to this run.
+# Every NRQL query below is auto-scoped to this run via cluster_name, which
+# we set to ${SCENARIO_TAG} when installing the chart.
 custom_test_key: cluster_name
 
 scenarios:
@@ -20,15 +19,13 @@ scenarios:
       - kubectl delete namespace nr-${SCENARIO_TAG} || true
     tests:
       nrqls:
-        # Assert the marker lines emitted by the log-producer (e2e-resources.yml) reached NRDB.
-        # No WHERE/SINCE here: the action appends the cluster_name scope clause automatically.
+        # Checks the marker log actually arrived.
         - query: "SELECT filter(count(*), WHERE message LIKE '%nr-logging-e2e-marker%') AS markerCount FROM Log"
           expected_results:
             - key: "markerCount"
               lowerBoundedValue: 1
-      # Unlike `after:` above, a failing command here fails the scenario.
-      # Logs can arrive fine even if the pod is crash-looping, so we
-      # check pod health too, not just that logs showed up.
+      # A failing script here fails the scenario (unlike `after:`, which is cleanup-only).
+      # We also check pod health, since logs can arrive fine even if the pod is crash-looping.
       scripts:
         - kubectl logs -n nr-${SCENARIO_TAG} -l app.kubernetes.io/name=newrelic-logging --all-containers --prefix=true --tail=200 || true  # for debugging, doesn't affect pass/fail
         - kubectl get pods -n nr-${SCENARIO_TAG} -o wide  # for debugging, doesn't affect pass/fail
--- a/charts/newrelic-logging/e2e/test-specs.yml
+++ b/charts/newrelic-logging/e2e/test-specs.yml
@@ -34,3 +34,35 @@ scenarios:
         - kubectl get pods -n nr-${SCENARIO_TAG} -o wide  # for debugging, doesn't affect pass/fail
         - kubectl wait --for=condition=Ready pod -n nr-${SCENARIO_TAG} -l app.kubernetes.io/name=newrelic-logging --timeout=60s  # fails if the pod never became healthy
         - kubectl get pods -n nr-${SCENARIO_TAG} -l app.kubernetes.io/name=newrelic-logging -o json | jq -e '[.items[].status.containerStatuses[].restartCount] | all(. == 0)' > /dev/null  # fails if the pod restarted
+
+  - description: lowDataMode strips extra metadata but keeps the fields the chart says it will
+    before:
+      - helm repo add newrelic https://helm-charts.newrelic.com
+      - helm dependency update ../
+      - helm upgrade --install ${SCENARIO_TAG} --namespace nr-${SCENARIO_TAG} --create-namespace ../ --values e2e-values.yml --set global.licenseKey=${LICENSE_KEY} --set global.cluster=${SCENARIO_TAG} --set lowDataMode=true
+      - kubectl apply -f e2e-resources.yml
+    after:
+      - kubectl logs -l app.kubernetes.io/name=newrelic-logging --all-containers --prefix=true || true
+      - kubectl get pods -A -o wide || true
+      - kubectl delete -f e2e-resources.yml || true
+      - helm uninstall ${SCENARIO_TAG} --namespace nr-${SCENARIO_TAG} || true
+      - kubectl delete namespace nr-${SCENARIO_TAG} || true
+    tests:
+      nrqls:
+        # lowDataMode strips most metadata, but a few basic fields should still survive.
+        # This checks those fields didn't get dropped by mistake.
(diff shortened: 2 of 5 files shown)


skothuri-nr:
@satya-nr any reason for it not doing on forked repo PRs? is this how it is in newrelic?

satya-nr:
> @satya-nr any reason for it not doing on forked repo PRs? is this how it is in newrelic?

GitHub doesn't give secrets to fork PRs, so this test can't run there by default. I've now added a second workflow (same pattern [nr-k8s-otel-collector](https://github.com/newrelic/helm-charts/blob/master/.github/workflows/nr-k8s-otel-e2e-external.yml) uses) to cover fork PRs too.

satya-nr:
@skothuri-nr this PR contains integration tests for Linux only. Windows will need extra effort — Linux uses Minikube, and we'll need to figure out an equivalent setup for Windows. It's worth adding, so I'll do the Windows work in a new PR next sprint. Please review this one for Linux for now.

skothuri-nr (dismissed):

pbhadra0112 (commented):
LGTM

.github/workflows/newrelic-logging-e2e-external.yml
```diff
@@ -0,0 +1,48 @@
...
+
+env:
+  # NOTICE that apart from this, the versions in the chart linter matrix needs to be bumped too.
+  LATEST_K8S_VERSION: 'v1.35.0'
+  MINIKUBE_VERSION: 'v1.38.1'
+
+jobs:
+  e2e-test-external:
+    name: Run Tests
+    # Do not run e2e tests if PR has skip-e2e label or if it's a non-forked repo (internal contribution)
```
maya-jha: description is confusing. Its still running e2e tests. Do you really need two different yml files for this?

.github/workflows/newrelic-logging-e2e-external.yml
```diff
@@ -0,0 +1,48 @@
...
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'ci/skip-e2e') && github.event.pull_request.head.repo.full_name != 'newrelic/helm-charts' }}
+    runs-on: ubuntu-latest
+    environment: E2E # Required so that approval is required to run this job since it is untrusted (forked).
```
maya-jha, line 25: How are we making sure that approval is received for running it?

charts/newrelic-logging/e2e/test-specs.yml
```diff
@@ -0,0 +1,60 @@
+description: New Relic newrelic-logging chart E2E
+
+# Every NRQL query below is auto-scoped to this run via cluster_name, which
+# we set to ${SCENARIO_TAG} when installing the chart.
+custom_test_key: cluster_name
+
+scenarios:
```
maya-jha, line 7: How are we testing for different distros and windows?

maya-jha (changes requested):

maya-jha:
Update PR description if it only support windows? Also list next action items.

maya-jha:
Please review comments from Claude as well:

Major

- [charts/newrelic-logging/e2e/test-specs.yml:50-55] The labelLeakCount == 0 assertion passes on an empty result set, so it cannot fail (rule #5). The run log for this PR proves it: at 08:03:25 the sibling retainedCount query failed with got 0.000000 — no marker logs had reached NRDB yet — while labelLeakCount produced no error at that same iteration=0, meaning it was already "satisfied" by zero rows. A count == 0 check against a not-yet-populated dataset would pass even if fluent-bit were dead or lowDataMode were leaking labels under a different attribute name. Fix: fold both aggregates into one query so the retry loop can only settle once markers have actually arrived:

- query: "SELECT filter(count(*), WHERE message LIKE '%nr-logging-e2e-marker%' AND pod_name IS NOT NULL AND container_name IS NOT NULL AND namespace_name IS NOT NULL) AS retainedCount, filter(count(*), WHERE message LIKE '%nr-logging-e2e-marker%' AND labels.app IS NOT NULL) AS labelLeakCount FROM Log"
  expected_results:
    - key: "retainedCount"
      lowerBoundedValue: 1
    - key: "labelLeakCount"
      value: 0

- Worth confirming separately that labels.app is the attribute name the default (non-lowDataMode) path actually produces — values.yaml:129-144 has no nest/lift filter, so default-mode metadata stays nested under kubernetes, whereas the lifted top-level labels.app shape only exists on the lowDataMode path (values.yaml:165-170). If the name is wrong, the negative check stays green forever even after the fix above.
- [charts/newrelic-logging/e2e/ (whole directory)] The new e2e/ directory is packaged into the published chart (rule #6 — test scaffolding crossing into the distributed artifact). Verified locally: helm package charts/newrelic-logging then tar -tzf lists newrelic-logging/e2e/e2e-resources.yml and newrelic-logging/e2e/test-specs.yml. charts/newrelic-logging has no .helmignore, while the sibling charts/nr-k8s-otel-collector/.helmignore excludes exactly ci/, e2e/, tests/. Customers pulling the chart get a busybox Deployment manifest and CI specs. Fix: add charts/newrelic-logging/.helmignore mirroring the sibling chart. (ci/ and tests/ already leak today — pre-existing, and your call whether to sweep them in the same file or leave them.)

Minor

- [charts/newrelic-logging/e2e/test-specs.yml:15] The first after: command omits -n nr-${SCENARIO_TAG}, so it queries the default namespace and always returns nothing. The chart installs into nr-${SCENARIO_TAG}, so this diagnostic is dead weight — and it duplicates line 30, which does pass the namespace. Fix: add -n nr-${SCENARIO_TAG}, or delete the line since scripts: already captures those logs.
- [.github/workflows/newrelic-logging-e2e.yml:24,39-40] Worst-case NRQL polling exceeds the job timeout. retry_attempts: 10 × retry_seconds: 90 = up to 15 min per scenario × 2 scenarios = 30 min against timeout-minutes: 20. A genuinely broken pipeline burns the whole timeout and reports "job cancelled" rather than the assertion error, which is the more useful signal. Observed on this PR: each scenario needed exactly one retry (~90s). Fix: drop retry_attempts to ~6, keeping ample headroom over observed latency while leaving both scenarios inside 20 min. Same applies to newrelic-logging-e2e-external.yml:33-34,25.
- [charts/newrelic-logging/e2e/ — missing README.md] No e2e/README.md, where the sibling chart documents scenarios, CI entry point, and how to run the suite manually against EKS/GKE/AKS. Without it, the only way to learn how to run these locally is to reverse-engineer test-specs.yml and the action's env contract. Fix: short README following charts/nr-k8s-otel-collector/e2e/README.md.

.github/workflows/newrelic-logging-e2e-external.yml
satya-nr: Two files are needed because fork PRs don't get repo secrets under a normal pull_request trigger — that's a GitHub security restriction. To test fork PRs at all, we need a separate pull_request_target workflow gated by manual approval. This mirrors the same pattern nr-k8s-otel-collector already uses (nr-k8s-otel-e2e.yml / nr-[k8s-otel-e2e-external.yml](https://github.com/newrelic/helm-charts/blob/master/.github/workflows/nr-k8s-otel-e2e-external.yml)).

You're right that the comment on that line was confusingly worded — it mentioned "internal contribution" without saying where those PRs actually get tested. I've updated properly.

.github/workflows/newrelic-logging-e2e-external.yml
satya-nr: This job needs the E2E GitHub environment, which requires approval from the K8S-AGENTS team before it runs. So it pauses until someone on that team approves.

charts/newrelic-logging/e2e/test-specs.yml
satya-nr: Right now CI only runs against Minikube on Linux (ubuntu-latest), covering the default install and lowDataMode scenarios — it doesn't test other managed Linux distros (EKS/GKE/AKS) automatically, though charts/newrelic-logging/e2e/README.md now has instructions for running manually against those.

For Windows: that's tracked separately in [ticket](https://new-relic.atlassian.net/browse/NR-601614) —already  working on it.

satya-nr:
> Update PR description if it only support windows? Also list next action items.

Updated the PR description — added a "Next steps" section calling out Windows e2e coverage, and clarified in "Special notes" that this PR's e2e scope is Linux only via Minikube.

satya-nr:
Addressed all the Claude review findings except one — the labelLeakCount merge broke CI (the test tool doesn't support combining two checks into one query), so I reverted that one back to the original two-query setup, which already works fine. The other fixes are in. CI is green.

.github/workflows/newrelic-logging-e2e-external.yml
maya-jha: We need to run it when our team also approves it. Will it work for that?

satya-nr: Yes — added @newrelic/logint as an additional required reviewer on the E2E environment. Either team's approval now unblocks the job.

charts/newrelic-logging/e2e/test-specs.yml
```diff
@@ -0,0 +1,60 @@
...
+    before:
+      - helm dependency update ../
+      - helm upgrade --install ${SCENARIO_TAG} --namespace nr-${SCENARIO_TAG} --create-namespace ../ --set global.licenseKey=${LICENSE_KEY} --set global.cluster=${SCENARIO_TAG} --set lowDataMode=true
+      - kubectl rollout status daemonset/${SCENARIO_TAG}-newrelic-logging -n nr-${SCENARIO_TAG} --timeout=60s
+      - kubectl apply -f e2e-resources.yml
+    after: *cleanup
+    tests:
+      nrqls:
+        # Basic fields should survive lowDataMode. Checks they weren't dropped by mistake.
+        - query: "SELECT filter(count(*), WHERE message LIKE '%nr-logging-e2e-marker%' AND pod_name IS NOT NULL AND container_name IS NOT NULL AND namespace_name IS NOT NULL) AS retainedCount FROM Log"
```
maya-jha, line 45: should we use different marker here : `nr-logging-e2e-marker-lowdatamode`? Can both tests conflict?

maya-jha:
How can we test for pod crash scenario?

maya-jha:
> How can we test for pod crash scenario?

nevermind, I see tests for that `We also check pod health, since logs can arrive fine even if the pod is crash-looping.`

charts/newrelic-logging/e2e/test-specs.yml
satya-nr: No conflict — isolation comes from cluster_name (set per scenario via global.cluster=${SCENARIO_TAG}), not the marker text, and scenarios run sequentially with full cleanup in between. Same marker is fine as-is.

maya-jha (approved):

Philip-R-Beckwith (approved):

PR #2362, opened 2026-07-28
[agent-control-deployment] feat: cleanup resources on uninstall [NR-572127]

sigilioso:

This PR extends the uninstallation-job for `agent-control-deployment` in order to cleanup the Agent Control resources on uninstallation.

Before the change only ConfigMaps where removed when the chart was used standalone.

Uses the k8s CLI's updates on
- https://github.com/newrelic/newrelic-agent-control/pull/2742
- https://github.com/newrelic/newrelic-agent-control/pull/2751

#### Checklist
- [x] Chart Version bumped

## 🚀 What's Changed

### DIFF
--- a/charts/agent-control-deployment/templates/service-account-uninstallation.yaml
+++ b/charts/agent-control-deployment/templates/service-account-uninstallation.yaml
@@ -1,5 +1,5 @@
 
-{{- if and (not .Values.cdEnabled) .Values.rbac.create -}}
+{{- if .Values.rbac.create -}}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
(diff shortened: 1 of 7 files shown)


charts/agent-control-deployment/templates/service-account-uninstallation.yaml
```diff
@@ -1,5 +1,5 @@
 
-{{- if and (not .Values.cdEnabled) .Values.rbac.create -}}
+{{- if .Values.rbac.create -}}
```
paologallinaharbur, line 2: do you know why we were executing the cleaning just if cdEnabled was false?

was is like this since it was a wordaround in place just for pipeline control, right?

paologallinaharbur (approved):

charts/agent-control-deployment/templates/service-account-uninstallation.yaml
sigilioso: We were **trying** to exclude it when `cdEnabled` was false because it was a workaround, but we were failing! `.Values.cdEnabled` was not defined and never provided, it should have been `.Values.config.cdEnabled`. As a result `(not .Values.cdEnabled)` was always `true` and the check was useless 😅

PR #2385, opened 2026-08-06
[agent-control-bootstrap] drop permissions drop permissions bootstrap

paologallinaharbur:
We are dropping the permissions not needed anymore.

We are mounting such ClusterRoles with RoleBindings therefore the cluster-wide resources are dropped anyway


sigilioso:
It also bumps agent-control-bootstrap dependencies, includes #2389

sigilioso (approved):

PR #2397, opened 2026-08-19
[newrelic-logging] make persistence init container image configurable

satya-nr:
The fluent-bit persistence cleanup init container hardcoded `busybox:1.36`, unlike every other image in this chart which comes from `values.yaml`. This blocked customers whose security policy requires all images to come from an internal registry (e.g. Harbor/Artifactory).

Makes it configurable via `fluentBit.persistence.initContainer.image` (repository/tag/registry/pullPolicy), using the same image helper as the main container. Default behavior unchanged.

  - No GitHub issue — reported via Slack (Case ID 00352319, banking prospect POV blocker).

Nested under `fluentBit.persistence` (not top-level) to avoid colliding with the existing `initContainers`/`windows.initContainers` list keys.

#### Testing
- **Unit tests**: 3 new cases in `tests/images_test.yaml` — default image, global registry override, local override wins over global. All 67 tests pass.
- **Static checks**: `helm lint` clean, `helm template` verified for all 4 combinations.
- **Live cluster test**: deployed to a real cluster with a staging license key. Default image (`docker.io/busybox:1.36`) and an overridden registry (`public.ecr.aws`) both pulled and completed successfully; logs confirmed delivered to New Relic staging in both cases.

#### Checklist
- [x] Chart Version bumped

## 🚀 What's Changed


skothuri-nr (dismissed):

pbhadra0112 (commented):
Approved

charts/newrelic-logging/templates/daemonset.yaml
```diff
@@ -53,7 +53,8 @@ spec:
       initContainers:
       {{- if and (.Values.fluentBit.persistence) (eq .Values.fluentBit.persistence.mode "persistentVolume") }}
         - name: init
-          image: busybox:1.36
+          image: {{ include "newrelic.common.images.image" ( dict "imageRoot" .Values.fluentBit.persistence.initContainer.image "context" .) }}
```
maya-jha, line 56: Will it handle the case when someone is upgrading but still uses the old values.yaml file? should we add  nil-safe chain with fallbacks, e.g. build the image root as (.Values.fluentBit.persistence.initContainer).image | default (dict "repository" "busybox" "tag" "1.36") and render imagePullPolicy via (…).pullPolicy | default "IfNotPresent".

charts/newrelic-logging/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart to deploy New Relic Kubernetes Logging as a DaemonSet, supporting both Linux and Windows nodes and containers
 name: newrelic-logging
-version: 1.40.0
+version: 1.40.1
```
maya-jha: bump minor version 1.41.0

charts/newrelic-logging/templates/daemonset.yaml
satya-nr: Verified this — Helm deep-merges the chart's default values.yaml with a customer's override file, so an old file that never mentions `initContainer` still gets the chart's default (`busybox:1.36`) filled in. Tested it directly: rendered with a values file missing that key entirely and it worked fine.

charts/newrelic-logging/Chart.yaml
satya-nr: updated chart version to 1.41.0

maya-jha (approved):

maya-jha:
Get approval from @nr-rkallempudi , lets merge on Monday if not critical.

PR #2401, opened 2026-08-20
[nr-k8s-otel-collector] feat: GKE Autopilot allowlist host-fs and host-namespace flags

TmNguyen12:
Adds three opt-in values to the nr-k8s-otel-collector chart for the GKE Autopilot partner allowlist work (NR-593526 / NR-606183). All default to false, so existing installs are unchanged.

## What's added

- `gkeAutopilotAllowList` (default false): when true and provider=GKE_AUTOPILOT, keeps the host root filesystem mount (host-fs to /hostfs) and the hostmetrics root_path that Autopilot normally strips. Lets a partner WorkloadAllowlist admit the mount so the filesystem scraper can read node filesystems (system.filesystem.*).
- `daemonset.hostPID` (default false): shares the host PID namespace so the hostmetrics process scraper can see node processes.
- `daemonset.hostNetwork` (default false): uses the host network namespace, which re-scopes the network scraper from the pod's own veth to the node's interfaces.

## Why

On GKE Autopilot, Warden blocks host mounts and host namespaces by default. These flags produce the exact pod shape a partner WorkloadAllowlist can admit, which is what unlocks node-level metrics. The namespace flags have no provider gating, so on GKE Standard or other clusters they apply directly with no allowlist involved.

## What hostNetwork changes

No new metric names appear. What changes is the scope of system.network.*:

| | Without hostNetwork | With hostNetwork |
|---|---|---|
| Interfaces reported | 1 (the collector pod's own eth0 veth) | 28 (node interfaces) |
| Example throughput | rx ~823 KB, tx ~66 KB | physical eth0 rx ~7.29 MB, tx ~6.6 MB |
| Coverage | pod-local traffic only | node eth0, cilium_host/net, docker0, all pod lxc* veths |
| Scope | pod | node |

This is the one capability that genuinely needs the host namespace grant. kubelet only emits per-pod k8s.pod.network.io, so node-level network on Autopilot has a single path: hostNetwork plus the allowlist exemption.

hostPID adds one metric, process.cpu.time (verified 372 datapoints with it, 0 without over 3 days). Coverage is partial, around 7 of 31 processes, not node-wide.

## Notes

- hostNetwork does not crash the collector. The :8888 self-telemetry bind conflict only occurs if self-telemetry is enabled, which is off by default.
- No entity-definitions changes required. These are existing metric names, and node/process metrics land as raw datapoints on Autopilot either way.
- `hostPID`/`hostNetwork` are exposed as daemonset-scoped values rather than wired to the common-library `newrelic.common.hostNetwork` helper, because host namespaces only apply to the node-level daemonset (never the cluster-level deployment) and `hostPID` has no common-library equivalent, so keeping both as explicit `daemonset.*` values scopes them correctly and keeps the two flags on one mechanism.

## Testing

Live-tested on a GKE Autopilot cluster against a partner WorkloadAllowlist. Unit coverage in charts/nr-k8s-otel-collector/tests/gke_auto_pilot_test.yaml. Full capability breakdown and NRDB verification captured under NR-606183.

#### Checklist

_Chart version is handled by the automated release process for `nr-k8s-otel-collector`, so no manual version bump is included. Three new values were added (`gkeAutopilotAllowList`, `daemonset.hostPID`, `daemonset.hostNetwork`), so the README was regenerated via `make generate-nr-k8s-chart-docs`._

## 🚀 What's Changed
* Add opt-in `gkeAutopilotAllowList`, `daemonset.hostPID`, and `daemonset.hostNetwork` flags to enable node-level filesystem, process, and network metrics on GKE Autopilot (via a partner WorkloadAllowlist) and other clusters; all default false

### DIFF
--- a/charts/nr-k8s-otel-collector/.helmignore
+++ b/charts/nr-k8s-otel-collector/.helmignore
@@ -30,3 +30,8 @@ ci/
 e2e/
 examples/
 tests/
+
+# Local scratch / generated output (never part of the chart package)
+_scratch/
+otel_output/
+_claude/
--- a/charts/nr-k8s-otel-collector/.helmignore
+++ b/charts/nr-k8s-otel-collector/.helmignore
@@ -30,8 +30,3 @@ ci/
 e2e/
 examples/
 tests/
-
-# Local scratch / generated output (never part of the chart package)
-_scratch/
-otel_output/
-_claude/
--- a/charts/nr-k8s-otel-collector/README.md
+++ b/charts/nr-k8s-otel-collector/README.md
@@ -102,6 +102,8 @@ If using GKE Autopilot, please set the following configuration in your values.ya
 provider: "GKE_AUTOPILOT"
 ```
 
+Setting `gkeAutopilotAllowList: true` keeps the `host-fs` mount and the hostmetrics `root_path` that GKE Autopilot normally strips, so node filesystem metrics (`system.filesystem.*`) can be collected. GKE Autopilot only admits that mount if a `WorkloadAllowlist` permitting the `/` hostPath is installed in the cluster. That allowlist is installed by an `AllowlistSynchronizer` custom resource (`auto.gke.io/v1`) that references the allowlist path. Without it, this flag has no effect.
+
 ## OpenShift
 
 If using OpenShift, please set the following configuration in your values.yaml file in order for the agent to work with OpenShift.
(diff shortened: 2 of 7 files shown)


TmNguyen12:
Closing in favor of two focused PRs, split by concern so the tested work is not held up by the untested work:

- #2402: `gkeAutopilotAllowList` flag for GKE Autopilot node filesystem metrics. Gated to `provider: GKE_AUTOPILOT`, live-tested on Autopilot (NR-606183).
- #2403: `daemonset.hostPID` and `daemonset.hostNetwork` flags. Generic pod-spec toggles, verified on Autopilot. Validation on non-Autopilot providers is tracked in NR-609210.

The allowlist change is already tested and does not depend on the host-namespace validation, so splitting lets #2402 move independently. All content here is preserved across the two PRs.

PR #2402, opened 2026-08-21
[nr-k8s-otel-collector] feat: gkeAutopilotAllowList flag for GKE Autopilot node filesystem metrics

TmNguyen12:
Adds one opt-in value, `gkeAutopilotAllowList` (default false), for the GKE Autopilot partner allowlist work ([NR-607198](https://new-relic.atlassian.net/browse/NR-607198), parent [NR-593526](https://new-relic.atlassian.net/browse/NR-593526)). Default false, so existing installs are unchanged.

## What it does
When `provider: GKE_AUTOPILOT` and `gkeAutopilotAllowList: true`, the chart keeps the host root filesystem mount (`host-fs` -> `/hostfs`) and the hostmetrics `root_path` that Autopilot's Warden normally strips. That is the exact pod shape a partner `WorkloadAllowlist` can admit, which lets the filesystem scraper read the node's filesystems (`system.filesystem.*`). The allowlist is installed cluster-side by an `AllowlistSynchronizer` (`auto.gke.io/v1`). Without it, the flag has no effect.

## Provider gating
The flag is a no-op on any provider other than GKE Autopilot. It is ANDed with the `gkeAutopilot` provider check, so on OpenShift or a standard cluster setting it true changes nothing (OpenShift still strips `root_path`).

## Testing
Live-tested on a GKE Autopilot cluster against a partner `WorkloadAllowlist`. NRDB verification captured under [NR-606183](https://new-relic.atlassian.net/browse/NR-606183). Unit tests cover three cases: Autopilot + flag keeps the mount and `root_path`, Autopilot without the flag strips them, and OpenShift + flag does not add `root_path`.

## Notes
README regenerated via `make generate-nr-k8s-chart-docs`. The regen also corrects pre-existing drift in the committed README (collector image tag `1.14.0` -> `1.19.0`, `kubeletstats` -> `kubelet_stats` wording). Those are not functional changes.

## Update — global inheritance for the allowlist flag, plus a stale pin I ran into

- Set global.* inheritance for `gkeAutopilotAllowlist`, same pattern `provider` already uses via `global.provider`.
- New helper: `newrelic.common.gkeAutopilotAllowlist` in `common-library` (`2.3.3` -> `2.3.4`), copied from the `lowDataMode` helper.
- Local value wins if set, otherwise falls back to `global.gkeAutopilotAllowlist`.
- Swapped the 3 places in `daemonset.yaml`/`daemonset-configmap.yaml` that read `.Values.gkeAutopilotAllowlist` directly to go through the helper instead. Provider gating around them is untouched.
- Fixed a stale dependency pin along the way (see below).

### Bundled cleanup: `CHART-TEMPLATE`'s dependency pin was stale

- Not related to this feature, but I hit it while adding tests. `library/CHART-TEMPLATE` (the harness chart the library uses to test its own helpers in isolation) had `common-library` pinned at `2.1.0`, two minor versions behind the real `2.3.3`.
- Because of that, `helm dependency update` was failing outright for that chart, which meant all 38 of its existing test suites were erroring out with "no template associated" - not because anything was actually broken, just because the dependency never resolved.
- Bumped the pin to track current. Split this into its own commit ahead of the feature commit so it doesn't get buried in an unrelated diff, and so it's independently obvious in the history that it was a pin-only fix.

### Testing (update)

- Added `cl-gkeautopilotallowlist_test.yaml` in `CHART-TEMPLATE` to cover the helper itself, same 6-case matrix `lowDataMode` uses: default, both-null, global-true, local-true, local-overrides-global-false, local-overrides-global-true.
- Added 2 cases to this chart's own `gke_auto_pilot_test.yaml` to cover actual usage: `global.gkeAutopilotAllowlist: true` keeps the host-fs mount and `root_path`, and a local `false` correctly overrides a global `true`.
- Ran `helm unittest` on both charts after every change: 39/39 suites on `CHART-TEMPLATE` (all passing now, not just mine), 21/21 on `nr-k8s-otel-collector`.
- Also manually diffed `helm template` output with `--set global.provider=... --set global.gkeAutopilotAllowlist=true` against the equivalent top-level `--set` flags to confirm they render byte-identical.

#### Checklist

_Chart version is handled by the automated release process for `nr-k8s-otel-collector`, so no manual version bump is included._

## 🚀 What's Changed
* Add opt-in `gkeAutopilotAllowlist` flag (default false): on GKE Autopilot, keeps the host-fs mount and hostmetrics `root_path` so a partner WorkloadAllowlist can admit them and the filesystem scraper can read node filesystems (`system.filesystem.*`). No-op off Autopilot. Now inheritable via `global.gkeAutopilotAllowlist`.

[NR-607198]: https://new-relic.atlassian.net/browse/NR-607198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NR-593526]: https://new-relic.atlassian.net/browse/NR-593526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NR-606183]: https://new-relic.atlassian.net/browse/NR-606183?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TmNguyen12:
Splitting this into two PRs so the common-library helper can release independently of the CI failure it was causing here (pinning common-library@2.3.4, which didn't exist upstream yet):
- #2425 — common-library: adds the gkeAutopilotAllowlist helper
- #2426 — nr-k8s-otel-collector: consumes it (depends on #2425 merging/releasing first)

PR #2418, opened 2026-08-31
[newrelic-logging] fix liveness probe failing on IPv6/dual-stack clusters

skothuri-nr:
 fix liveness probe failing on IPv6/dual-stack clusters

  - fixes #

#### Checklist
- [x] Chart Version bumped

## 🚀 What's Changed

#### Testing

Verified locally against both an IPv4-only and an IPv6-only kind cluster: no restarts/Unhealthy events on either after the change, and no regression for existing IPv4-only clusters.

### DIFF
--- a/charts/newrelic-logging/Chart.yaml
+++ b/charts/newrelic-logging/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart to deploy New Relic Kubernetes Logging as a DaemonSet, supporting both Linux and Windows nodes and containers
 name: newrelic-logging
-version: 1.41.0
+version: 1.41.1
 appVersion: 3.8.0
 home: https://github.com/newrelic/kubernetes-logging
 icon: https://newrelic.com/assets/newrelic/source/NewRelic-logo-square.svg
--- a/charts/newrelic-logging/values.yaml
+++ b/charts/newrelic-logging/values.yaml
@@ -114,7 +114,9 @@ fluentBit:
           Daemon        off
           Parsers_File  parsers.conf
           HTTP_Server   On
-          HTTP_Listen   0.0.0.0
+          # Listening on :: binds a dual-stack socket, so the health check
+          # (and liveness probe) works on IPv4-only, IPv6-only, and dual-stack clusters.
+          HTTP_Listen   ::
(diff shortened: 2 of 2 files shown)


charts/newrelic-logging/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart to deploy New Relic Kubernetes Logging as a DaemonSet, supporting both Linux and Windows nodes and containers
 name: newrelic-logging
-version: 1.41.0
+version: 1.41.1
 appVersion: 3.8.0
```
pbhadra0112, line 5: How is this version update will help the image to support IPv6 / dual-stack listening?Please mention a bit detail in the description or add a comment .

charts/newrelic-logging/values.yaml
```diff
@@ -114,7 +114,9 @@ fluentBit:
           Daemon        off
           Parsers_File  parsers.conf
           HTTP_Server   On
-          HTTP_Listen   0.0.0.0
+          # Listening on :: binds a dual-stack socket, so the health check
+          # (and liveness probe) works on IPv4-only, IPv6-only, and dual-stack clusters.
+          HTTP_Listen   ::
```
nikhil31sai, line 119: How did you validate the happy path for ipv6?

nikhil31sai (commented):
Change looks small, only query is if we validated it.

charts/newrelic-logging/values.yaml
skothuri-nr: I have created local kind kubernetes cluster with ipv6 pods and validated it. I reproduced the issue first and then applied this new change to confirm whether its working or not

charts/newrelic-logging/Chart.yaml
skothuri-nr: its because of this change `HTTP_Listen   ::` made in values.yaml which is a dual stack wildcard which listens for both ipv6 and ipv4 instead of `0.0.0.0 ` which only listens foripv4

nikhil31sai (approved):
Change looks clean. 
LGTM.

pbhadra0112 (approved):

maya-jha (approved):

